PR #31, opened 2019-08-23
Added docker support

JStumpp:
Currently the build has to be done manually. If you could grant me access to the label-studio repository I can enable automatic builds for each push/release. 

Thanks,

Jürgen


package.json
```diff
@@ -66,7 +66,7 @@
     "prettier-styles": "prettier --write src/**/*.{css,scss}",
     "lint": "eslint --debug src/",
     "lint:write": "eslint --debug src/ --fix",
-    "publish": "npm run copy-examples && GENERATE_SOURCEMAP=false npm run build",
+    "publish": "GENERATE_SOURCEMAP=false npm run build",
```
shevchenkonik:
```javascript
npm run copy-examples
```
@JStumpp This command is responsible for the correct bundle and future testing of LS.

PS. I will be back on Friday checking this PR.

JStumpp: @shevchenkonik yes, I know, but it is already in npm build and I did not see a reason to execute it twice.

PR #34, opened 2019-08-28
Predictions (with new task API)

makseq:


backend/server.py
```diff
@@ -163,7 +163,7 @@ def api_completions(task_id):
         completion.pop('state', None)  # remove editor state
         db.save_completion(task_id, completion)
         log.info(msg='Completion saved', extra={'task_id': task_id, 'output': request.json})
-        return answer(201, 'ok')
+        return answer(201, json.dumps({'id': task_id}))
```
shevchenkonik: Incorrect condition because of this case could be duplicate between completion ID and task ID

PR #51, opened 2019-09-23
Support multi labeling of studio

shevchenkonik:
- Change the title of the button. (Honeypot -> Ground Truth)
- Fix bug with an incorrect id of completion.
- Correct logic of handling boolean values in MST.
- Fix bug with incorrect variable of env.
- The first iteration of HOC in modules (Audio).
- Support multi labeling in Tags (Image).


deppp:
@shevchenkonik Regarding the

> Correct logic of handling boolean values in MST.

I've tried yesterday to configure the Audio tag and pass zoom="false" and it didn't work. Does this PR fix that?

shevchenkonik:
> @shevchenkonik Regarding the
> 
> > Correct logic of handling boolean values in MST.
> 
> I've tried yesterday to configure the Audio tag and pass zoom="false" and it didn't work. Does this PR fix that?

@deppp Yes. In part with boolean tag values.

PR #73, opened 2019-10-01
fixing lunch->launch typos

kmader:
it says lunch a few places on the site and I think it should be launch.


deppp:
Thanks for noticing that @kmader merging it in

kmader:
@deppp I hadn't been sure if it was intentional or not, but it's a great project!

deppp:
It wasn't, but makes me think about the lunch though haha
Thanks for the feedback!

PR #76, opened 2019-10-04
duration it takes to create a new completion

deppp:


deppp:
@shevchenkonik When you have a minute can you please check that I'm not missing anything here? especially that loadDate is correctly initialized

PR #97, opened 2019-10-18
Return app outside of new LabelStudio

makseq:


shevchenkonik:
Correct way: https://github.com/heartexlabs/label-studio/pull/98

PR #108, opened 2019-10-24
Adjust docs for embedding the frontend

tammoippen:
Hi,

I am experimenting with embedding the label studio with [labelbox](https://labelbox.com/) as a [custom editor](https://labelbox.com/docs/custom-interface-sdk/custom-interfaces). It works surprisingly good and I like your editor actually better than the default one. 

But, I found some issues / inconsistencies with the docs. It was very difficult to remove all interaction with the backend... 

- `#editor` -> `editor`: `document.getElementById('editor')` will be called.
- extend `apiCalls` docs, as it was not clear to me, what actually happens.
- extend `explore` docs to highlight interaction with the backend
- consistency: either `expert.pk` or `expert.id`
- consistency: `deleteCompletion` had different description than the other two.

### DIFF
--- a/docs/source/guide/frontend.md
+++ b/docs/source/guide/frontend.md
@@ -10,13 +10,13 @@ Instantiate a new Label Studio object with a selector for the div that should be
 
 ```xhtml
 <!-- Include Label Studio stylesheet -->
-<link href="https://unpkg.com/label-studio@0.2.1/build/static/css/main.05fec320.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


shevchenkonik:
@tammoippen 
Hey! Thank you for PR and your changes are important.

The current version of the documentation has some inaccuracies. There will be a more detailed description in future versions, ofc.

- Yes, LS will be called in HTML with ID of div (initialization example in Quickstart: https://labelstud.io/guide/frontend.html#Quickstart)
- `apiCalls` is a boolean parameter for using custom backend API (currently in draft status)
- `explore` will be described in the next version

PR #119, opened 2019-11-09
Streaming images from local dir

niklub:
Now we can point to local directory, e.g.
```
python server.py --input-path local/directory/with/images
```
and we recursively convert all images to labeling tasks

https://github.com/heartexlabs/label-studio/issues/107

### DIFF
--- a/backend/server.py
+++ b/backend/server.py
@@ -243,7 +243,7 @@ def api_instruction():
     return make_response(c['instruction'], 200)
 
 
-@app.route('/images/<path:filename>')
+@app.route('/data/<path:filename>')
 def get_image_file(filename):
     directory = request.args.get('d')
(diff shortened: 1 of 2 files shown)


backend/server.py
```diff
@@ -243,6 +243,12 @@ def api_instruction():
     return make_response(c['instruction'], 200)
 
 
+@app.route('/images/<path:filename>')
```
makseq: почему ты только любых картинок сделал? Кажется можно вообще любые файлы так отдавать.

makseq: Maybe we can support any file types?

niklub: Good point. I've included audio types as well checking when filetypes match current config.

makseq: I'd like to rename /images/ to /assets/ or something like and proxy any type of files over this

PR #157, opened 2020-01-01
Pip package, cmd tool & configuration prompt

niklub:


backend/setup.py
```diff
@@ -0,0 +1,30 @@
+import setuptools
+
+with open('README.md', 'r') as f:
+    long_description = f.read()
+
+
+with open('requirements.txt') as f:
+    requirements = f.read().splitlines()
+
+setuptools.setup(
```
makseq: А где examples? когда в promt указываешь input/output/config - там надо автоматом подставить путь на самый популярный example (наверное bboxes)

makseq: и статик файлы бекенда не заинклужены..

niklub: Prompt automatically collects paths specified in config.json
Static files are collected by `setuptools.find_packages()`, then are searchable using find_file/find_dir utils
Please double check this if I'm not missing any js/css/media files served by flask server (apparently there are endpoints for them, but now everything is settled in /static/ - should we get rid of editor/build_path dir in config?)

PR #174, opened 2020-01-10
fix broken cli

zehengl:
To fix issue https://github.com/heartexlabs/label-studio/issues/173

`required` is not a valid argument for `add_subparsers`


deppp:
Thanks for the PR @zehengl, merging in.

PR #190, opened 2020-01-18
Pretty examples

makseq:


label_studio/project.py
```diff
@@ -159,13 +160,15 @@ def validate_label_config_on_derived_input_schema(self, config_string_or_parsed_
...
                 raise ValidationError(
+                    'You have imported tasks already and they are not compatible with new config. '
                     'Can\'t find type "{item}" among already imported tasks with types {input_schema_types}'
                         .format(item=item, input_schema_types=list(input_schema_types)))
 
         # check input data values: they must be in schema
         for item in input_values:
             if item not in input_schema_values:
                 raise ValidationError(
+                    'You have imported tasks already and they are not compatible with new config. '
```
niklub: You have already imported tasks and they are incompatible with a new config

label_studio/project.py
```diff
@@ -551,6 +554,13 @@ def create_project_dir(cls, project_name, args):
         # create label config (config.xml)
         if not os.path.exists(default_label_config_file):
             default_label_config = '<View></View>'
+            try:
+                path = find_file('examples/image_polygons/config.xml')
+            except IOError:
```
niklub: when this exception could be thrown ?

PR #229, opened 2020-03-04
ML backend fixes

guneemwelloeux:
2 fixes:
* wrong default config generated, causing a crash when the model name parameter is read in the ML initialisation
* logging issue that would cause a crash when several schemas are returned by the ML /validate endpoint


label_studio/utils/schema/default_config.json
```diff
@@ -17,7 +17,7 @@
 
     "!ml_backend": {
         "url": "http://localhost:9090",
-        "model_name": "my_super_model"
+        "name": "my_super_model"
```
niklub: Hi, @guneemwelloeux ! Thank's for pointing to this. I believe that this parameter is unused due to the key `"!ml_backend"` (with exclamation). It was preliminary just for demo purpose, and could be deleted How do you get this error in runtime?

PR #238, opened 2020-03-16
add docker-compose file

CrazyOptimist:


Dockerfile
```diff
@@ -13,5 +13,4 @@ EXPOSE ${PORT}
 
 COPY . /label-studio
 
-RUN pip install -e .
-CMD ["label-studio", "start", "my_project", "--init"]
```
niklub: Thanks a lot for contributing! It would be nice if you can also update [Docker startup readme] (https://github.com/heartexlabs/label-studio#run-docker)

CrazyOptimist: Thank you very much for your kind reviewing!
I agree on removing this line and did it.
```bash
CMD ["label-studio", "start", "my_project", "--init"]
```
But this line should be there because I met some errors missing it.
```bash
RUN pip install -e .
```
Everything else would be fine.

CrazyOptimist:
I guess I need to make another Dockerfile for using docker-compose, because I see the one in the repository is being used for docker registry as well as for some kind of deployment.
Let me know your opinion.

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3"
+
+services: 
+  label_studio:
+    container_name: label_studio_server
+    build: .
+    working_dir: /label-studio
+    command: "label-studio start my_project --init --force"
```
niklub: Should we mount an external volume here?

niklub: You can do this by the following:
```yaml
command: "label-studio start my_project --init --root-dir /data"
volumes:
   - ./label-studio-projects:/data
```

CrazyOptimist: Sorry for late response, lots of works related to corona these days...
New changes have been tested including readme file.
Waiting your feedback.
Thank you!

niklub: @CrazyOptimist thanks for your contribution! hope the pandemic doesn't break your goals :)

CrazyOptimist: Many thanks @niklub.
And wish you the best all the contributors!

PR #243, opened 2020-03-25
Fix one click deploy

iCorv:
Hi all, this fixes the one click deployment to google cloud run.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,11 @@ WORKDIR /label-studio
 COPY requirements.txt /label-studio
 RUN pip install -r requirements.txt
 
-ENV PORT="8200"
+ENV PORT="8080"
 ENV collect_analytics=0
 EXPOSE ${PORT}
 
 COPY . /label-studio
 
 RUN pip install -e .
-CMD ["label-studio", "start", "my_project", "--init"]
+CMD ["label-studio", "start", "my_project", "--init", "--no-browser", "--port", "8080"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -7,11 +7,11 @@ WORKDIR /label-studio
...
+ENV PORT="8080"
 ENV collect_analytics=0
 EXPOSE ${PORT}
 
 COPY . /label-studio
 
 RUN pip install -e .
-CMD ["label-studio", "start", "my_project", "--init"]
\ No newline at end of file
+CMD ["label-studio", "start", "my_project", "--init", "--no-browser", "--port", "8080"]
```
niklub: Hi, @iCorv ! Thanks a lot for fixing this! Just wanting to know why it's needed to change the default app's port ?

niklub: If it's needed for running on GCP, can we use some environmental setup in `app.json` for a specific port?

iCorv: Actually it is recommended to listen to the GCP port variable instead of setting it yourself:

"Developers should listen on $PORT rather than explicitly setting it." (https://cloud.google.com/run/docs/configuring/containers)

If you want to set the port explicitly, everyone will have to first set up a Project on GCP and change the default port in the project settings before hitting the cloud run button. I guess this would not be very intuitive.

iCorv: Port 8080 is the default port. However, it may be better to only use the GCP provided environment variable if that is possible?

niklub: Ok, I see, that's good point. In that case, it's better to change the port to 8080 everywhere. I'll do it

PR #247, opened 2020-04-02
ML page, tutorial & cold-start predicts

niklub:


nabinkhadka:
What is the intention behind having "Import data" file uploader in the model setup page? @makseq @niklub

makseq:
@nabinkhadka you can check your model using it. Just drag & drop file there and prediction will be generated immediately.

nabinkhadka:
Will the data be a new task and saved? Will the predictions be saved?

makseq:
Yes, it will be a new task and prediction will be saved.

nabinkhadka:
Ah.. Basically it's a shortcut to import tasks, nothing to do with pred/model. As the predictions are generated only when moved to the tasks page. Right @makseq

makseq:
@nabinkhadka It imports a task and then forcibly run the prediction for this task.

Issue #217, opened 2020-02-17
Labelling page does not work with HTTPS redirect


guneemwelloeux:
**Describe the bug**
After deploying the app behind a HTTPS-redirecting proxy, the Labelling page does not work anymore.

**To Reproduce**
1. Deploy the app behind NGINX, configured to redirect HTTP to HTTPS requests
2. Go to Labelling
3. See error in console: 
`Mixed Content: The page at 'https://label.mydomain.dev/' was loaded over HTTPS, but requested an insecure resource 'http://label.mydomain.dev/api/projects/1/next/'. This request has been blocked; the content must be served over HTTPS.`

**Expected behavior**
The app should follow all the redirects and the api call should succeed, the scheme being conserved at each step of the way.

**Environment (please complete the following information):**
 - OS: macOS
 - Browser Chrome 
 - Version 80.0.3987.87

**Additional context**
I believe the redirect issued by label-studio upon calling "api/projects/1/next" is not taking the scheme into account, when it should

niklub:
There is no easy way to support HTTPS in Label Studio. I suggest you may try to do the following modifications in `server.py` file:

1. SSLify flask app using
```python
from flask_sslify import SSLify
...
sslify = SSLify(app, permanent=True)
```

2. Add certificates to SSL context while running app, i.e. when app.run is launched, add
```python
app.run(
            host='0.0.0.0',
            port=input_args.port,
            debug=input_args.debug,
            ssl_context=(
                'fullchain.pem',
                'privkey.pem'
            )
        )
```

Hope this helps!

rverma-jm:
This solution may work when we do ssl termination at the app, what if we want to do the ssl termination at loadbalancer?

vincenzomanzoni:
The solution proposed by @niklub works when the app does the SSL termination. However, since in my case is behind a reverse proxy (traefik), the optimal soution would be to conserve the scheme for every call. Any suggestions?

kretes:
I did a try on getting https work using: https://github.com/kretes/label-studio/tree/https#this-fork-just-contains-a-try-to-get-https-working-in-docker-compose-removed-postgres-btw by changing the nginx configuration. Any comments welcome.

Labelling works over https with this approach.

PR #261, opened 2020-04-17
Update path of /next endpoint, to match server-side naming

tim-jackson:
### Problem
We are running a container with the latest docker image of LabelStudio behind an ALB on AWS (HTTPS only).
We noticed that after adding tasks to LabelStudio, clicking on 'Labelling' or accessing the root of the application, a spinning wheel was shown and would not disappear, with the page failing to load.
Examining the Browser console, it was evident that upon accessing the root of the application, the front end was making a request to (HTTPS) /api/projects/1/next
LabelStudio responded with a 302 redirect, to (HTTP) /api/projects/1/next/
Note that the redirect was to a HTTP address, rather than HTTPS. This caused the browser (Chrome in our case), to prevent the request from being made - hence why the page failed to load.
It appears to be an underlying Flask dependency that is incorrectly producing the incorrect HTTP redirect URL.

### Fix
There are a couple of potential ways to fix this.
Option (1) is to update the endpoint path on the Flask side (server.py). However, I'm unsure of the endpoint naming conventions being followed - there appears to be a mixture of endpoints with a trailing slash, and some without.
Option (2) is to update the path of the API endpoint that the front end calls to load the page data.

This PR implements Option (2), which matches the front end API endpoint to what is implemented server side - avoiding the redirect in the first place.

### DIFF
--- a/label_studio/static/js/lsb.js
+++ b/label_studio/static/js/lsb.js
@@ -10,7 +10,7 @@ const API_URL = {
   COMPLETIONS: "/completions",
   CANCEL: "/cancel",
   PROJECTS: "/projects",
-  NEXT: "/next",
+  NEXT: "/next/",
   EXPERT_INSRUCTIONS: "/expert_instruction",
 };
 


tim-jackson:
I believe this should also resolve #217

Issue #273, opened 2020-04-28
absolute paths break reverse proxies (e.g. nginx)


ezavesky:
When attempting to stand-up running version behind a nginx reverse proxy, absolute path parameters cause failures.

**To Reproduce**
1. Launch labelstudio
2. Broken image and links will result
3. Inspection of the retrieval paths indicates incorrect path specification

**Expected behavior**
When using the below code for `nginx`, the application should run correctly at the new URL.

```
http {
    map $http_upgrade $connection_upgrade {
        default upgrade;
        '' close;
    }

    server {
        listen 5000;
        server_name localhost;

        location /my_prefix/another {
            rewrite /my_prefix/another/(.*) /$1  break;
            rewrite /my_prefix/another /  break;
            proxy_pass http://localhost:8200;
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection $connection_upgrade;
        }
    }
}
# code originally from https://github.com/streamlit/streamlit/issues/1139
```

**Screenshots**
Broken page and bad requests.
![page_error_01](https://user-images.githubusercontent.com/12480297/80541828-7aa27b80-8971-11ea-9814-d2e855d2fd09.jpg)
![page_error_02](https://user-images.githubusercontent.com/12480297/80541830-7bd3a880-8971-11ea-8d1f-09f45024fc81.jpg)

PR #274, opened 2020-04-28
fix URL paths for proxy safety

ezavesky:
- update JS, HTML pages for proxied safety
- attempts to fix issue #273; now works with or without a proxy


ezavesky:
a few more changes for proxy use, mirrors those suggested in #261

PR #277, opened 2020-05-05
Release 0.6.0

niklub:


Ziinc:
hi @niklub , I've noticed a config issue when starting up label-studio.

for a project started with init on v0.5.1 and before, there is no "host" key in the `config.json` set.

This results in an error when running this branch. 

The culprit line from the stacktrace is here:
https://github.com/heartexlabs/label-studio/blob/073e61e3c8d5ec3b3d54bec711e0492d7046e243/label_studio/server.py#L726 

Perhaps a fix would be to auto-set the config defaults if not set?

Ziinc:
Found another issue (though I wasn't able to find the exact culprit line).

When the `config.json` does not the `ml_backends` key, it results in an error when adding an ML backend through the app interface. A default empty array on the config dict on startup would resolve it

Ziinc:
Found another one, related to the `ml_backends` keys.

On start up, when any of the ml_backends servers are not running, the main label-studio process throws a ValueError and renders a 500 internal server error.
```
ValueError: ML backend {'url': 'http://localhost:9090', 'name': '777823aa'} is not connected.
```

Issue #291, opened 2020-05-13
Docker-compose up not working on Mac OS


EMCP:
**Describe the bug**
Readme.md states to run the following command to start the application via `docker-compose`, but when I do this the webapp is not running or not reachable.

**To Reproduce**
1. Clone the project on Mac OS
2. Execute the README.md's requested `INIT_COMMAND='--init' docker-compose up -d`
3. Wait for command to finish
4. Web app should be reachable from localhost:8080, but nothing is there

**Expected behavior**
The web app should come up, or we need to clean up the documentation if docker-compose no longer works

**Screenshots**
n/a

**Environment (please complete the following information):**
 - OS: Mac OS Catalina 
 - Browser Chrome
 - Version -  master branch as of todays date

**Additional context**
last lines of output from docker-compose are as follows 

```
Removing intermediate container 03664b9975b9
 ---> cddaaf963684
Step 10/10 : CMD ["label-studio", "start", "my_project", "--init", "--no-browser", "--port", "8080"]
 ---> Running in 394b0d7f9d48
Removing intermediate container 394b0d7f9d48
 ---> ba1dd352e473
Successfully built ba1dd352e473
Successfully tagged label-studio_label-studio:latest
WARNING: Image for service label-studio was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Creating label-studio ... done
```

I even tried again with another command cited in the readme.. no avail.

```
(base) emcp@EMCPs-Mac-mini label-studio % docker-compose up -d
WARNING: The INIT_COMMAND variable is not set. Defaulting to a blank string.
Recreating label-studio ... done
(base) emcp@EMCPs-Mac-mini label-studio % 

```

EMCP:
heres a `docker ps` showing the image is running.. but the webapp is not reachable with the commands given in the readme.. likely need to be modified

```
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS              PORTS                    NAMES
69d130308460        label-studio_label-studio   "label-studio start …"   4 minutes ago       Up 4 minutes        0.0.0.0:8080->8080/tcp   label-studio


```

EMCP:
regular docker command seems to work fine, happy to report

```
docker run --rm -p 8080:8080 -v `pwd`/my_project:/label-studio/my_project --name label-studio heartexlabs/label-studio:latest label-studio start my_project --init --host 0.0.0.0
```

[...]

PR #295, opened 2020-05-15
Add bind ip to docker-compose file

CrazyOptimist:
#291 
I've found out that `docker-compose up -d` is not working today and added bind ip.
Thanks for your hard work for improving gents!


makseq:
This issue was fixed in the last releases.

PR #320, opened 2020-06-16
Release 0.7.3

makseq:
Features: 
* display skipped completions in task manager
* implementation of tests
* prev/next buttons for label stream
* https://github.com/heartexlabs/label-studio/issues/288 (html upload as files)
* generate signed URLs on GCE instance (#326)
* relative path to js static
* image previews in tasks (if label config contains images)
* basic auth with username and password

Fixes:
* fix with Text and Header floats in import tasks
* fix uri resolver on s3 with custom credentials passed
* https://github.com/heartexlabs/label-studio/issues/330
* https://github.com/heartexlabs/label-studio/issues/329


PhilipMay:
I have an issue with 0.7.3. See here: #346

PR #351, opened 2020-07-10
Use gevent's WSGIServer instead of plain Flask

k4200:
## The issue we had

Some of my coworkers and I had an issue that some of the HTTP requests are blocked for several minutes (in my case 6 minutes), which leads to the system being practically unusable.

Below are screenshots of Chrome's developer tool:

welcome-network-1
welcome-network-2

The same issue occurs in Firefox and Safari as well.

This issue looks like happening inside Flask:
https://stackoverflow.com/q/54775894/564246

## What this PR does

Wraps around `app.run` with gevent's `WSGIServer` to enable the server to better handle concurrent requests.

## Details

I found some information related to this issue.

https://stackoverflow.com/a/14823968/564246

> `flask.Flask.run` accepts additional keyword arguments (`**options`) that it forwards to `werkzeug.serving.run_simple`
(snip)
`threaded` defaults to `True` as of Flask 1.0

https://stackoverflow.com/a/29887309/564246

> but keep in mind that the built-in WSGI server in Werkzeug (which is packaged into Flask and what it uses for `app.run()`) is only single-threaded.

## Misc.

I'm not a Python guy, so let me know if there's anything to improve. Thank you in advance.


makseq:
Thank you for your PR!
It's better to have an option to enable an alternative webserver.

k4200:
Thank you for your comment. I added an option `--use-gevent`. Let me know if there's anything else to improve.

PR #355, opened 2020-07-13
Improve/next task speed

lightforever:
I have 3000 tasks. next_task takes around 20 seconds.

The reason is `i not in self.target_storage` invokes collecting files every iteration.

There is no reason of using self.target_storage. `completed_tasks_ids` already has the ids

### DIFF
--- a/label_studio/project.py
+++ b/label_studio/project.py
@@ -406,7 +406,7 @@ class Project(object):
         sampling = self.config.get('sampling', 'sequential')
 
         # Tasks are ordered ascending by their "id" fields. This is default mode.
-        task_iter = filter(lambda i: i not in self.target_storage, self.source_storage.ids())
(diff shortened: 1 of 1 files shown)


label_studio/project.py
```diff
@@ -406,7 +406,7 @@ def next_task(self, completed_tasks_ids):
         sampling = self.config.get('sampling', 'sequential')
 
         # Tasks are ordered ascending by their "id" fields. This is default mode.
-        task_iter = filter(lambda i: i not in self.target_storage, self.source_storage.ids())
+        task_iter = filter(lambda i: i not in completed_tasks_ids, self.source_storage.ids())
```
makseq, line 409: @niklub Check it

makseq:
@lightforever Hi! Check out my latest commit in release/0.7.4, the speed has been improved significantly. Your feedback will be appreciated.

PR #394, opened 2020-08-11
Make deepcopy of values returned from ExternalTasksJSONStorage

hepaajan:
otherwise it will replace resolved data url in memory and store that
resolved url in completions. That will not work with ml backends because
presigned urls expire.

Also this makes ExternalTasksJSONStorage semantically same as other
cloudstorages. Specifically you need to use "set" to make changes.

### DIFF
--- a/label_studio/storage/filesystem.py
+++ b/label_studio/storage/filesystem.py
@@ -1,6 +1,7 @@
 import json
 import os
 import logging
+from copy import deepcopy
 
 from label_studio.utils.io import json_load, delete_dir_content, iter_files
 from .base import BaseStorage, BaseForm, CloudStorage
@@ -183,7 +184,7 @@ class ExternalTasksJSONStorage(CloudStorage):
         return self.path
 
     def _get_value(self, key):
-        return self.data[int(key)]
+        return deepcopy(self.data[int(key)])
 
     def _set_value(self, key, value):
         self.data[int(key)] = value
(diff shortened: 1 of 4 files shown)


label_studio/storage/filesystem.py
```diff
@@ -183,7 +184,7 @@ def readable_path(self):
         return self.path
 
     def _get_value(self, key):
-        return self.data[int(key)]
+        return deepcopy(self.data[int(key)])
```
niklub, line 187: Maybe it's better to put it right in `resolve_task_data_uri()`? I'm concerned about this unnecessary step in the case when URLs should not be resolved

hepaajan: That was my initial though also. Then I thought that ExternalTasksJSONStorage is used by default and it is semantically bit different compared to other storages which will fetch a copy of the data always. Putting it there would prevent similar issues being introduced later and task is currently fairly light weight object.

That said, either way will fix the issue I'm facing, so I'm definitely fine with putting it to resolve_task_data_uri() also.

makseq: Can I merge it? @niklub

niklub: @hepaajan I agree, guess it's better to unify storage design to avoid further issues, even at the cost of some minor performance reduction. @makseq let's merge it with https://github.com/heartexlabs/label-studio/pull/350

PR #414, opened 2020-08-31
Maintain ordering in source.json and task.json

theconsummate:
The issue:

When a new project is created, a source.json file is created from a task.json file. It would be convenient if the ordering of id's in the task.json file is also maintained in the source.json file. Using an ordered set accomplishes this goal

behaviour in previous versions:
```
task.json:
[id1, id2, id3..., idn]

source.json:
{0 -> id1, 1 -> id2, 2-> id3, n-1 -> idn ...}
```

behaviour in current version:
```
task.json:
[id1, id2, id3..., idn]

source.json (random ordering due to set):
{0 -> id4, 1 -> id10, 2-> id111, n-1 -> id3 ...}
```

The changes in this PR revert the current behaviour to the previous behaviour, i.e. maintain the ordering.

If random ordering is required during annotation, then the sampling configuration should be used rather than changing the composition of the source.json file.

### DIFF
--- a/label_studio/storage/base.py
+++ b/label_studio/storage/base.py
@@ -12,6 +12,7 @@ from flask_wtf import FlaskForm
 from wtforms import StringField, BooleanField
 from wtforms.validators import InputRequired, Optional, ValidationError
 from collections import OrderedDict
+from ordered_set import OrderedSet
 
 from label_studio.utils.io import json_load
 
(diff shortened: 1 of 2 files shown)


ftesser:
Hi all, I experienced the same issue and I am in favor of this PR.

PR #434, opened 2020-09-18
drop support for EOL python 3.5

graingert:
Python 3.5 hit EOL last week, 2020-09-13 https://devguide.python.org/#status-of-python-branches

flask dropped support for py3.5 here https://github.com/pallets/flask/pull/3554


makseq:
Sorry, but we need to support python 3.5.

PR #455, opened 2020-10-14
Release 0.7.6

makseq:


label_studio/utils/uploader.py
```diff
@@ -58,6 +58,10 @@ def tasks_from_file(filename, file, project):
         elif project is None:
             raise ValidationError('No tasks found in: ' + filename)
 
+        # no drag & drop support
```
niklub, line 61: is it duplicate?

.github/workflows/e2e-tests.yml
```diff
@@ -15,7 +15,9 @@ jobs:
       with:
         python-version: '3.x'
```
niklub, line 16: is it better to fix python version like 
`python-version: '3.8'` ?

PR #479, opened 2020-11-02
Fix parse_input_args to accept json config

### LINKED ISSUE #478: server.json is overwritten by parse_input_args?
**Describe the bug**
It looks like any server config provided in server.json is overwritten in server.main() by parse_input_args.

Detailed:
Here, server.json is parsed to global input_args:
https://github.com/heartexlabs/label-studio/blob/1f8974514ac76087eb34ffba855a570c84612da0/label_studio/server.py#L79-L86

However, here, it is overwritten in main():

https://github.com/heartexlabs/label-studio/blob/1f8974514ac76087eb34ffba855a570c84612da0/label_studio/server.py#L1100-L1104

I tested this and it looks like server.json, if provided, does nothing since it is simply overwritten.

In addition, when hosting label-studio from e.g. gunicorn instead over label-studio start, _all_ arguments have to be provided in server.json currently. This makes configuring over server.json complicated. 

I would try to tackle both issues and write a PR in the next days.

kasuteru:
This intends to fix #478 . Functionality should not have changed as long as no server.json is provided.

If a server.json is provided and successfully read, it is now used to supply the base config. Any arguments provided in cmd line as well overwrite server.json arguments.

Let me know if you have questions about the implementation, or require me to issue the PR to a different branch.

Commit message: 
Also, custom arguments can now be provided if using custom label-studio
host solution or for debugging.

Printing version info was moved to main for consistency.

As long as no custom arguments or existing arguments are provided,
functionality has not changed.

### DIFF
--- a/label_studio/utils/argparser.py
+++ b/label_studio/utils/argparser.py
@@ -1,20 +1,30 @@
 import os
 import json
+from types import SimpleNamespace
 
 from label_studio.utils.io import find_dir
 from label_studio.utils.misc import iter_config_templates
 
 
-def parse_input_args():
+def parse_input_args(existing_args=None, custom_args=None):
     """ Combine args with json config
 
-    :return: config dict
+    :existing_args: existing config as SimpleNamespace
+    :custom_args: overwrite sys.argv with list of simulated cmd line arguments
+
+    :return: config SimpleNamespace
     """
     import sys
     import argparse
     from label_studio.project import Project
 
-    if len(sys.argv) == 1:
+    # Select args:
+    if custom_args:
+        env_args = custom_args
+    else:
+        env_args = sys.argv[1:]
+
+    if len(env_args) == 0:
         print('\nQuick start usage: label-studio start my_project --init\n')
 
     available_templates = [os.path.basename(os.path.dirname(f)) for f in iter_config_templates()]
@@ -39,7 +49,7 @@ def parse_input_args():
         '--force', dest='force', action='store_true',
         help='Force overwrite existing files')
     root_parser.add_argument(
-        '--root-dir', dest='root_dir', default='.',
+        '--root-dir', dest='root_dir',
         help='Project root directory')
     root_parser.add_argument(
         '-v', '--verbose', dest='verbose', action='store_true',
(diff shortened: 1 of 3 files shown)


kasuteru:
Runs are failing, I will check this again.

kasuteru:
Whoops, forgot to commit renaming of a variable. All checks pass now.

label_studio/utils/argparser.py
```diff
@@ -39,7 +49,7 @@ def valid_filepath(filepath):
         '--force', dest='force', action='store_true',
         help='Force overwrite existing files')
     root_parser.add_argument(
-        '--root-dir', dest='root_dir', default='.',
```
niklub, line 42: why did you remove the defaults ?

label_studio/utils/argparser.py
```diff
@@ -172,4 +177,25 @@ def valid_filepath(filepath):
...
+        # then overwrite with the existing arguments, then overwrite these 
+        # with the arguments that were actually set (args_set):
+        args_merged = {**vars(args), **vars(existing_args), **args_set}
+        args = SimpleNamespace(**args_merged)
+
+    # To comply with legacy code, some arguments have to be set as default
+    # if they aren't defined yet:
+    args.root_dir = args.root_dir or "."
+    args.input_format = args.input_format or "json"
+    args.sampling = args.sampling or "sequential"
```
kasuteru, line 198: Defaults are now listed here and set to "None" above. This is because we need to differentiate between user-provided arguments and arguments that are set by default.

According to convention, the priority should be: (With cmd line the highest)
cmd line -> config file -> defaults

label_studio/utils/argparser.py
kasuteru: See comment below - if we set them here, we later cannot determine whether "." was provided by the user or by default. This prevents us from determining whether "." should be overwritten by a path provided in the config file (in case it is a default) or not (in case it was provided by the user via cmd line argument)

kasuteru: This is a known limit of Python's argparse and there are libraries to tackle this issue, e.g. https://pypi.org/project/ConfigArgParse/ . However, I think changing it like this is an acceptable way.

kasuteru:
Release 0.8.1 significantly changes the files involved in this PR. I will close this PR for now without merging and see whether I need to adapt this PR for 0.8.1

makseq:
@kasuteru Sorry, but there was a big changes because of blueprint integration. You need to move everything from your server.py to blueprint.py.

PR #489, opened 2020-11-12
Front end build x image multi select

davodesign84:


davodesign84:
oops, wrong repo. Sorry about that :)

PR #527, opened 2021-01-04
UI Automation base steps for Image labeling

ahovhannes:
All 4 steps finished.


test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        x = 50
+        y = 150
+        actions.move_to_element_with_offset(self.driver.find_element_by_tag_name('body'), 0,0)
+        actions.move_by_offset(x, y).click().perform()
+        actions.move_to_element_with_offset(self.driver.find_element_by_tag_name('body'), 0,0)
+        actions.move_by_offset(x, y+50).click().perform()
+        actions.move_to_element_with_offset(self.driver.find_element_by_tag_name('body'), 0,0)
+        actions.move_by_offset(x+50, y+30).click().perform()
+        actions.move_to_element_with_offset(self.driver.find_element_by_tag_name('body'), 0,0)
+        actions.move_by_offset(x, y).click().perform()
```
hlomzik, line 184: this can be simplified with just `move_by_offset` after each other, without resetting to body. also it's better to offset from image/canvas tag, it's more safe and logical.

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
...inlinecentered&quot;></div></div></div><script>window.onload=function(){vartask={&quot;id&quot;:42,&quot;data&quot;:{&quot;image&quot;:&quot;static\/samples\/sample.jpg&quot;},&quot;completions&quot;:null,&quot;predictions&quot;:null,&quot;project&quot;:1,&quot;created_at&quot;:&quot;2019-02-06T14:06:42.000420Z&quot;,&quot;updated_at&quot;:&quot;2019-02-06T14:06:42.000420Z&quot;};varLS=newLabelStudio(&quot;label-studio&quot;,{config:`<View><Headervalue=&quot;Selectlabelandstarttoclickonimage&quot;></Header><Imagename=&quot;image&quot;value=&quot;$image&quot;></Image><PolygonLabelsname=&quot;label&quot;toName=&quot;image&quot;strokeWidth=&quot;3&quot;pointSize=&quot;small&quot;opacity=&quot;0.9&quot;><Labelvalue=&quot;Airplane&quot;background=&quot;red&quot;></Label><Labelvalue=&quot;Car&quot;background=&quot;blue&quot;></Label></PolygonLabels></View>`,user:{pk:1,firstName:&quot;Awesome&quot;,lastName:&quot;User&quot;},task:_convertTask(task),interfaces:[&quot;basic&quot;,&quot;side-column&quot;,&quot;predictions&quot;],onLabelStudioLoad:function(LS){varc=LS.completionStore.addCompletion({userGenerate:true});LS.completionStore.selectCompletion(c.id);}});};</script></body></html>"]')
```
hlomzik, line 116: the whole code of iframe left intact. that's bad for lot of reasons.
and this even would not work in any next test because of hardcoded build name.

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        # action.move_to_element_with_offset(el, 500, 500)
+        # action.click()
+        # action.perform()
+        #
+        x = 700
+        y = 300
+        action.move_by_offset(x, y).click().perform()
+        action.move_by_offset(x, y+50).click().perform()
+        action.move_by_offset(x+30, y+30).click().perform()
+        action.move_by_offset(x, y).click().perform()
```
hlomzik, line 136: these offsets are relative after each other, so 4th are on 2830 by x axis, for away after screen's boundary :)
(as I know)

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        #
+        #Go to outsite from iframe
+        self.driver.switch_to_default_content()
+
+
+    #Step2 - Import tasks to project
+    def step2(self, testCaseName):
+        # Go to page Import
+        print("..... Going to page Import .....")
+        self.driver.find_element_by_xpath(self.welcome_Import).click()
```
hlomzik, line 146: I think it's better to extract checks on the page into distinct step/method and separate them from the way you get on this page, because this can be explicit like here (direct link) or implicit (like after submit and redirect).

hlomzik: and this can be a part of test — to check if the browser redirected automatically to another page after some action

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        #
+        # Choose image for dragAndDrop
+        print('Direktory is: ' + str(os.getcwd()))
+        self.import_img_src = str(os.getcwd()) +'/'+ self.import_img_src
+        print('Image is: ' +str(self.import_img_src))
+        self.driver.find_element_by_id(self.import_img_draganddrop_id).send_keys(self.import_img_src)
```
hlomzik, line 152: wow, great way to upload!

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+    def tearDown(self):
+        self.driver.close()
+
+    def test_cases(self):
+        print("..... Opening the site "+self.site_url+" .....")
+        self.driver.get(self.site_url)
+        print("..... Executing Test-Cases .....\n")
+        #
+        # self.step1('Setup your label config')
+        # self.goToWelcomePage()
```
hlomzik, line 72: why is this step skipped? there would not be any config on the run from scratch, as I understand

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        fp.set_preference("browser.download.folderList", 2)
+        fp.set_preference("browser.download.manager.showWhenStarting", False)
+        fp.set_preference("browser.download.dir", self.downloadDir)
+        # fp.set_preference("browser.helperApps.neverAsk.saveToDisk", "attachment/csv")
+        # fp.set_preference("browser.helperApps.neverAsk.saveToDisk", "attachment/json")
+        fp.set_preference("browser.helperApps.neverAsk.saveToDisk", "attachment/zip")
+        self.driver = webdriver.Firefox(firefox_profile=fp)
+        #
+        self.wait = WebDriverWait(self.driver, 10)
+        self.probel5 = '      '
```
hlomzik, line 36: ?

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        except NoSuchElementException:
+            return False
+        return True
+
+    #Step1 - Setup your label config
+    def step1(self, testCaseName):
+        # Go to page Setup
+        print("..... Going to page Setup .....")
+        self.driver.find_element_by_xpath(self.welcome_Setup).click()
+        time.sleep(2)
```
hlomzik, line 94: I prefer some explicit waiting for some element or action instead of just waiting, because it's more safe, understandable and predictable.

hlomzik: what the reason here to wait for 2 seconds?

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+    #Step4 - Export completions
+    def step4(self, testCaseName):
+        # Go to page Export
+        print("..... Going to page Export .....")
+        self.driver.find_element_by_xpath(self.welcome_Export).click()
+        #
+        # Click on Submit
+        self.driver.find_element_by_xpath(self.export_submit_btn).click()
+        print("Completion exported")
+        time.sleep(3)
```
hlomzik, line 201: what is this waiting for?

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        print("..... Executing Test-Cases .....\n")
+        #
+        # self.step1('Setup your label config')
+        # self.goToWelcomePage()
+        self.step2('Import tasks to project')
+        self.goToWelcomePage()
+        self.step3('Start labeling tasks')
+        self.goToWelcomePage()
+        self.step4('Export completions')
+        self.goToWelcomePage()
```
hlomzik, line 78: I think it's better to test the whole flow, so steps are going one after another, without jumping to welcome page

hlomzik (changes requested):

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        self.site_logo = '//a[@id="logo"]/img'
+        self.welcome_Setup  = '//a[contains(text(), "Setup")]'
+        self.welcome_Import = '//a[contains(text(), "Import")]'
+        self.welcome_Start  = '//a[contains(text(), "Start")]'
+        self.welcome_Export = '//a[contains(text(), "Export")]'
+        #
+        #self.Setup_label_car = '//span[contains(text(), "Car")]'
+        #self.Setup_label_car = '//span[@class="ant-tag"]'
+        #self.Setup_label_car = '//span[@class="ant-tag"][contains(text(), "Car")]'
+        self.Setup_label_car = '/html/body/div/div/div/div/div/div/div[1]/div/div/div[2]/span[2]'
```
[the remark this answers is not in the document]
ahovhannes: That will be changed during Setup step implementation.

test/ui_test_label_studio.py
ahovhannes: There were problem with iframe element finding process via selenium. After discussion with Nikolay, we decide to leave Setup step test implementation, because that page's iframe element can be changed to div element...

test/ui_test_label_studio.py
ahovhannes: This is only one test-chain. In future we will add a lot of test-chains)

test/ui_test_label_studio.py
ahovhannes: It were added just for video recording and will be deleted.

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,213 @@
...
+        #
+        self.import_img_draganddrop_id = 'file-input'
+        # self.import_img_draganddrop_text = '//p[contains(text(), "Drag and drop your files here or click for import")]'
+        self.import_img_src = 'img/imgForDragAndDrop.jpg'
+        # self.import_img_src = '/Users/ahovhannes/Desktop/Hovo/h/JOB/Job_Label_Studio/label_studio_project/label-studio/test/img/imgForDragAndDrop.jpg'
```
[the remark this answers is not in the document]
ahovhannes: Fixed)

test/ui_test_label_studio.py
ahovhannes: This is from Setup step, which will be rebuild )

test/ui_test_label_studio.py
ahovhannes: This is from Setup step, which will be rebuild, (depended from iframe... will it be removed or no )

test/ui_test_label_studio.py
ahovhannes: It were for logging and I removed it for now)

test/ui_test_label_studio.py
ahovhannes: This were for debugging)

test/ui_test_label_studio.py
ahovhannes: Without resetting we will have not wanted random dotes. 
Than is a good suggestion to offset by image/canvas, but when I tried offset with canvas: not succeeded.
Lets keep this implementation for now)

test/ui_test_label_studio.py
ahovhannes: I think we will add another test chain for that)

test/ui_test_label_studio.py
hlomzik: where do these random dots can come from? you'll only make a few clicks with explicitly given offsets.
about canvas... there can be some hidden ones maybe, so this should be checked on frontend side

test/ui_test_label_studio.py
```diff
@@ -0,0 +1,190 @@
...
+        # self.driver.switch_to.frame(iframe_)
+        # self.driver.find_element_by_xpath(self.setup_label_car)
+        # time.sleep(5)
+        #
+        # Click on Label Airplane
+        label_car_btn = self.wait.until(EC.element_to_be_clickable((By.CLASS_NAME, self.setup_label_car)))
```
hlomzik, line 100: so does it work with iframe now?
for Setup I think it's enough to just select some predefined template and click Save

hlomzik (changes requested):
Could you please add all required files to run this test? I'd like to check it in action :)
And I think it will fail on config part.

test/ui_test_label_studio.py
ahovhannes: About Setup page: thanks for suggestion) Now script clicks on Save button and saves default template.
After discussions I can add your wanted predefined template)

ahovhannes:
> Could you please add all required files to run this test? I'd like to check it in action :)
> And I think it will fail on config part.

You must download 'test' folder cd to that 'test' folder and just run command
python ui_test_label_studio.py

It will not fail, and you will see all 4 steps sequentially :)

test/ui_test_label_studio.py
ahovhannes: This is working chain. We can discuss about needed scenarios and add them here)

hlomzik:
It fails :)
I mean — how to configure all the setup? I never used selenium before and as I see, I have to install it, install some drivers (how? their readme isn't clear)
Could you please add the whole setup that I can just run? Thank you!

ahovhannes:
> It fails :)
> I mean — how to configure all the setup? I never used selenium before and as I see, I have to install it, install some drivers (how? their readme isn't clear)
> Could you please add the whole setup that I can just run? Thank you!

  Due to project changes, my test scripts will be changed too.
  As agreed with Nikolay and Max, I will implement test scenarios in parametrised way and a lot of things will be changed... so, after implementation, I will add special readme file for testing too)

Issue #526, opened 2021-01-02
Cloud storage slow with 1000+ tasks


keesschollaart81:
**Describe the bug**
When using either GCP or AWS cloud storage BBox tasks labeling is very slow, in this report I'll stick with S3 but the behavior for Google is very similar. Loading the tasks list takes minutes. 

**To Reproduce**

I'm migrating from another tool so I generate the JSON files for tasks and completions. I have 3 buckets now:

- labelstudio-tasks (containing 0.json, 1,json, .... 1300.json)
- labelstudio-completions (containing 0.json, 1.json, ... 1300.json)
- labelstudio-images (containing filename1.png, another-file.png, ... 1300thfile.png)

A tasks file (0.json) looks like this:
```json
{
  "data": {
    "image": "s3://labelstudio-images/evidence_637431187722247603.png"
  },
  "id": 0
}
```
A completion file looks like this:
```json
{
  "completions": [
    {
      "lead_time": 0,
      "result": [
        {
          "from_name": "label",
          "image_rotation": 0,
          "original_height": 480,
          "original_width": 480,
          "to_name": "image",
          "type": "rectanglelabels",
          "value": {
            "height": 19.4746828677309,
            "rectanglelabels": [
              "person"
            ],
            "rotation": 0,
            "width": 8.906873574686431,
            "x": 21.53292474344356,
            "y": 65.31989381413912
          }
        }
      ]
    }
  ],
  "data": {
    "image": "s3://labelstudio-images/evidence_637431187722247603.png"
  },
  "id": 0
}
```

I install label-studio and start it with the following command:
```
label-studio start my_project --init --source s3 --source-path labelstudio-tasks --source-params '{ \"use_blob_urls\": false, \"regex\": \".*\" }' --target s3-completions --target-path labelstudio-completions' 
```

The app start correctly and eventually after configuring and a few minutes waiting, it shows the tasks. When I open a tasks it loads quickly and shows the correct bounding box. 

I added some debug logging statements in the s3.py to see what is going on. There I observed that it downloads all of the task and completions files a bunch of times. The more files, the long it takes of course.  

**Expected behavior**
I would expect the tool to be me optimized in working with a larger (cloud based) dataset. Maybe an initial import (with progress indication?) keeping a local index of filenames and compare that? Given the pager in the tasks list, only downloading the files from the current page would already help a lot. 

**Environment (please complete the following information):**
 - OS: Both Windows and Mac, Both GCS and S3
[...]

PR #529, opened 2021-01-05
Feature/redis

### LINKED ISSUE #437: Add Azure Storage to Cloud Storage connectors.
This is pretty straightforward - we have support for s3 and gcs, but Azure Storage is missing from the list.

Add a connector for either Azure Blob Storage, Data Lake Storage or both.

I think I might be able to implement this myself, but if someone is already working on this, I don't want to duplicate the work. I am happy to help with testing.

kasuteru:
This closes #437 and might also help with #526. 

I have tested this with multiple configurations, both single and multi-session, and hope it works as intended. I suspect my documentation isn't extensive enough yet - let me know if any questions remain.

If wanted, I can extend this further to make it more configurable - currently, it works for me, and I am not sure whether anyone else is going to use it.

Hope this PR goes to the correct branch.


label_studio/storage/redis.py
```diff
@@ -0,0 +1,154 @@
...
+    my_project/1 (the "/" is added if not already in path)
+    
+    Values are python dicts encoded as utf-8 strings.
+    """
+
+    description = 'Redis database connection'
+    
+    form = RedisStorageForm
+
+    def __init__(self, project_path, path=None, db=1, **kwargs):
```
niklub, line 51: do we have some roadblocks for adding other crucial connection parameters at once, like host, port, and password? that would allow using more general Redis connections besides the default localhost

label_studio/storage/redis.py
```diff
@@ -0,0 +1,154 @@
...
+            path = project_path
+
+        # Append "/" to path if necessary:
+        path = path.rstrip("/")+"/"
+        
+        super().__init__(path=path, project_path=project_path, **kwargs)
+
+        # Get redis database object and test connection:
+        self.r = get_redis_connection(db=db)
+        if not self.r.ping():
```
niklub, line 77: Does it raise `redis.exceptions.ConnectionError` ?

niklub: And anyway, without the Redis server enabled we can't proceed to work with `self.r` variable further - perhaps it's better to leave an exception here in order to prevent using this type of storage at an early stage

kasuteru:
No, should be an easy task, I will add a few more after having a look at
what additional parameters might be useful.

niklub <notifications@github.com> schrieb am Di., 5. Jan. 2021, 22:58:

label_studio/storage/redis.py
```diff
@@ -0,0 +1,154 @@
...
+    # Since tasks are always text, we use StrictRedis with utf-8 decoding.
+    return redis.StrictRedis(db=db, charset="utf-8", decode_responses=True)
+
+
+class RedisStorageForm(BaseForm):
+    path = StringField('Path', description='Storage prefix (optional)')
+    bound_params = dict(path='path')
+
+
+class RedisStorage(BaseStorage):
```
Ryanglambert, line 25: Is the BaseStorage abstraction actually being used by project.Project? 

It doesn't appear to be.

label_studio/storage/redis.py
kasuteru: Good point, I will modify this to `raise redis.exceptions.ConnectionError("No redis db id passed!")`

label_studio/storage/redis.py
kasuteru: This was supposed to end up in this conversation:

No, should be an easy task, I will add a few more after having a look at
what additional parameters might be useful.

kasuteru: I will instead try to implement a way of forwarding **kwargs since there are so many parameters that also might change in the future, and have the GUI support a subset.

label_studio/storage/redis.py
kasuteru: tbh, I just put it there because all the other classes implement it that way and I don't know the code well enough to just leave it out. Even if it doesn't, it might in the future...

label_studio/storage/redis.py
kasuteru: Adding generic parameters proved difficult since I don't fully understand what is happening in the front-end. I instead added host, port and password as parameters to the form GUI & backend in the updated PR.

> I will instead try to implement a way of forwarding **kwargs since there are so many parameters that also might change in the future, and have the GUI support a subset.

label_studio/storage/redis.py
kasuteru: Upon further consideration, I decided to instead raise ValueError if no db is provided and redis.exceptions.ConnectionError if the Ping fails.

kasuteru:
I updated the PR. Please let me know whether there are still any issues with the new PR.

makseq:
@kasuteru Could you fix tests here?

makseq:
@kasuteru Could you fix tests here?

makseq:
And it will be great if you resolve conflicts, I switched the main branch to 0.9.0.post6.

niklub:
@kasuteru any updates here? we think of including this PR in the upcoming release, your help will be very appreciated :)

kasuteru:
Sorry for the delay. I will try to fix this until Monday.

kasuteru:
I found some time to do it right away. I fixed the conflicts and merge error, checks seem to pass now. Let me know if anything else is needed.

PR #531, opened 2021-01-06
Azure Blob support

keesschollaart81:
Adding support for Azure Blob Containers as a Cloud Storage backend. Implementation is very similar to GCS and S3.

The current implementation expect the Azure Storage Account Name and Key to be in an environment variable. The other 2 cloud backends handle this as part of the SDK. I could also store these (secret) values in the project (like optional with AWS) if you like...

Documentation for Cloud Storages might be subject for a slimmer structure as there is quite some duplication...?

In Azure the concept of a bucket is called 'container' so I added it to some places as most Azure-only developers will get confused otherwise.

Related (but unfinished) work: #437


makseq:
Merged to https://github.com/heartexlabs/label-studio/pull/571

makseq:
There were a lot of conflicts with 0.9.0.post6, so I merged it manually.

PR #560, opened 2021-01-19
Fix version checking hang in offline deployment

### LINKED ISSUE #558: Check latest version make app hang undefinitely when running offline
Hello, 

When I run Label Studio in an offline environment (without Internet Access), the app seems to hang indefinitely at start. I think it could be because of https://github.com/heartexlabs/label-studio/blob/707223a0a2059c63bc89aab5250450672eb746a9/label_studio/server.py#L35.

https://github.com/heartexlabs/label-studio/blob/707223a0a2059c63bc89aab5250450672eb746a9/label_studio/utils/misc.py#L386

Actually add a timeout in requests.get() may solve the issue.

LamDang:
Add timeout to version check

Fix #558

### DIFF
--- a/label_studio/utils/misc.py
+++ b/label_studio/utils/misc.py
@@ -383,7 +383,7 @@ def start_browser(ls_url, no_browser):
 def get_latest_version():
     pypi_url = 'https://pypi.org/pypi/%s/json' % label_studio.package_name
     try:
-        response = requests.get(pypi_url).text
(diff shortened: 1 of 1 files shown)


LamDang:
I think timeout=10 is a bit long but in some cases necessary, please feel free to suggest other values

LamDang:
Thanks @makseq for your prompt answer!

Issue #518, opened 2020-12-19
Any way to connect label-studio to Minio bucket storage?


louisoutin:
Hello,

I saw that you recently added a feature to connect label-studio storage to S3 (GCP or AWS) but do you know if there is a way to connect it to a [Minio](https://github.com/minio/minio) bucket?  It should be straightforward as Minio buckets are AWS S3 compatibles.

Thanks for the help!
Louis

makseq:
Hello, 

We use boto3 client (https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) for s3 storage sync. So, if it's supported by Minio, I think it will be very easy to use in LS. 

If Minio doesn't support boto3, you could implement Minio storage here:
https://github.com/heartexlabs/label-studio/tree/master/label_studio/storage

As example you can use s3 storage code (see S3Storage class): 
https://github.com/heartexlabs/label-studio/blob/master/label_studio/storage/s3.py#L40

PR #577, opened 2021-01-31
Feature setting S3 endpoint using environment variable

herber523:
Hello,

I try to setting S3 endpoint using environment variable.

This might help with #518 , since minio s3 needs to set the s3 endpoint.

Thanks.


makseq (changes requested):
:slightly_smiling_face:

PR #593, opened 2021-02-09
Add while loop to check for more busy ports

leonardovida:
The current code ` if not input_args.debug and check_port_in_use('localhost', port):` is checking whether the port (e.g. default :5000) is in use and then adds +1, but checks only once. If a user has a range of ports in use, for example 5000-5002, `label studio` does not manage to find an open port.

Modified by moving `check_port_in_use()` into a while loop to continue the check until there is a free port `label studio` can use.

I also took the freedom to create a PR without opening an issue first, hope that is fine with you.

### DIFF
--- a/label_studio/blueprint.py
+++ b/label_studio/blueprint.py
@@ -981,12 +981,13 @@ def main():
             ssl_context = (cert_file, key_file)
 
         # check port is busy
-        if not input_args.debug and check_port_in_use('localhost', port):
-            old_port = port
-            port = int(port) + 1
-            print('\n*** WARNING! ***\n* Port ' + str(old_port) + ' is in use.\n' +
-                  '* Trying to start at ' + str(port) +
-                  '\n****************\n')
+        if not input_args.debug:
+            while check_port_in_use('localhost', port):
+                old_port = port
+                port = int(port) + 1
+                print('\n*** WARNING! ***\n* Port ' + str(old_port) + ' is in use.\n' +
+                    '* Trying to start at ' + str(port) +
+                    '\n****************\n')
 
         # external hostname is used for data import paths, they must be absolute always,
         # otherwise machine learning backends couldn't access them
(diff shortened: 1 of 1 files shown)


label_studio/blueprint.py
```diff
@@ -981,12 +981,13 @@ def main():
...
-            print('\n*** WARNING! ***\n* Port ' + str(old_port) + ' is in use.\n' +
-                  '* Trying to start at ' + str(port) +
-                  '\n****************\n')
+        if not input_args.debug:
+            while check_port_in_use('localhost', port):
+                old_port = port
+                port = int(port) + 1
+                print('\n*** WARNING! ***\n* Port ' + str(old_port) + ' is in use.\n' +
+                    '* Trying to start at ' + str(port) +
+                    '\n****************\n')
```
makseq, line 990: Could you add here a breaking condition to avoid endless loop in the edge case?

makseq: @leonardovida

makseq (changes requested):

label_studio/blueprint.py
leonardovida: @makseq Sure, but if you mean when port > 65535, `socket.connect_ex()` already takes care of that raising an `OverflowError`. If you mean something else, could you please specify?

makseq: Maybe we can limit the range to +1000 ports? I think it could take a lot of time until 65535...

makseq: @leonardovida I think we must be hurry to be included into 0.9.1 release :smiley:

leonardovida: @makseq Makes super sense limiting to +1000. I'm raising a `ConnectionError()`. Let me know if you want a more informative message or just want to `break` the loop and continue execution!

leonardovida: Forgot the lack of support for fStrings. Sorry for the usage of actions

PR #594, opened 2021-02-10
removed hardcoded UTF8 decoding, and let requests convert post as_text

laynr:


makseq (commented):
@laynr Could you please make some comments: why utf8 was bad?

laynr:
@makseq uf8 is 'good enough'.   

I was the one who submitted `request.data.decode("utf-8")` in the first place and it works, but If you look at the [source](https://github.com/pallets/werkzeug/blob/2c7a3df2fcd33b4f5a3ea597c1f13348ed4690a0/src/werkzeug/wrappers/response.py#L294) of werkzeug you can see `as_text` dynamically determines the encoding:
`        if as_text:
            rv = rv.decode(self.charset)
        return rv`
So I think`request.get_data(as_text=True)` is cleaner and uses werkzeug as intended... but feel free to reject.

... now if label-studio isn't expecting a utf16 encoded config, that is a separate issue and probably better handled somewhere  an appropriate error can be raised and passed to the user.

makseq:
@laynr Thanks a lot!

PR #605, opened 2021-02-18
Rearchitect and update docs

smoreface:
Updating content to be more quickstart/overview oriented.

After reading this page I can explain to my colleague which parts of Label Studio I might want to extend
After reading this page I can map out the steps of a data labeling project that I need to prepare for.

### DIFF
--- a/docs/source/guide/FAQ.md
+++ b/docs/source/guide/FAQ.md
@@ -64,36 +64,41 @@ The most common mistake while resource loading is <b>CORS</b> (Cross-Origin Reso
 <br/> 
 
 ## How to display labels on bounding boxes, polygons and other regions
+
+Click the gear icon when labeling to configure the labeling interface to suit your labeling use case.
+
 <center>
   <img src='../images/lsf-settings.png'>
 </center>
 
 
-## How to run LSB with external domain name
+## How to run Label Studio with external domain name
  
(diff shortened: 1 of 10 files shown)


docs/source/guide/FAQ.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: FAQ
+title: Troubleshoot Label Studio
```
makseq, line 2: @smoreface Why FAQ is a not good variant?

smoreface: Most of the time FAQs aren't actually helpful to readers -- see [this article about why](https://content-guide.18f.gov/our-approach/structure-the-content/#dont-use-faqs). Most importantly for us, most of the content on this page isn't actually a question, even if it is frequently asked.

I plan to change this page into a troubleshooting page for errors (like CORS etc.) and then move the other helpful content (like the setting up pre-labeling) to dedicated pages or more relevant pages (like a second machine learning page about setting up and working with pre-labels, in addition to the existing page about setting up machine learning backends!)

smoreface:
Planning to merge this on Monday and work on 1.0.0 docs in a different branch!

atakanokan:
Would it be possible to address this issue before the merge too? 

https://github.com/heartexlabs/label-studio/issues/625

makseq:
@smoreface I think you should connect with @hlomzik (Andrey) to clarify the question about selectedColor.

smoreface:
@atakanokan I'll address that issue in a future PR, this one doesn't address the tags documentation at all.

niklub (approved):

PR #648, opened 2021-03-12
Automatic labeling (for each formate will be used different templates)

ahovhannes:
Automatic labeling (for each formate will be used different templates)


label_studio/ui_test/readme.txt
```diff
@@ -0,0 +1,72 @@
...
++jpg
++png
+-gif
+----------
+=====font=====
++txt
+----------
+=====common formats=====
++csv
+-tsv
```
niklub, line 24: why those are not tested?

label_studio/ui_test/readme.txt
```diff
@@ -0,0 +1,72 @@
...
+-txt
++json
+
```
niklub, line 27: Audio, HTML, PDF, time series are missed

label_studio/ui_test/ui_test_label_studio.py
```diff
@@ -0,0 +1,435 @@
...
+        #     if self.choosed_template == i:
+        #         all_templates_links[i].click()
+        #         break
+        # time.sleep(5)
+
+
+
+if __name__ == "__main__":
+    unittest.main()
+
```
niklub, line 435: Need for CI scripts (github action yml) with clear description of input parameters

label_studio/ui_test/ui_test_label_studio.py
```diff
@@ -0,0 +1,435 @@
...
+                    self.driver.find_element_by_xpath(self.tasks_afterSaveSubmit_btn).click()
+                # We will redirect to page: 'No more data available for labeling' , self.page_tasks_afterSaveSubmit_url = 'http://localhost:8080/tasks?tab=1&labeling=1'
+            elif self.driver.current_url == self.page_tasks_afterSaveSubmit_url:
+                # Go Back, due to No data in page
+                print('Task_page: 2@@@@@@@@@@')
+                print('Task_page: No more data available for labeling... so clicking on Back link')
+                self.driver.find_element_by_xpath(self.tasks_go_back_link).click()
+
+
+        # if self.driver.current_url == self.page_tasks_afterSaveSubmit_url:
```
niklub, line 307: Please remove unused comments

label_studio/ui_test/ui_test_label_studio.py
```diff
@@ -0,0 +1,435 @@
...
+        fp.set_preference("browser.helperApps.neverAsk.saveToDisk", "attachment/zip")
+        self.driver = webdriver.Firefox(firefox_profile=fp)
+        self.wait = WebDriverWait(self.driver, 10)
+        #
+        self.site_url = 'http://localhost:8080/welcome'
+        # self.current_page = self.site_url
+        self.site_logo = '//a[@id="logo"]/img'
+        self.page_import_url = 'http://localhost:8080/import'
+        #
+        # MENU
```
niklub, line 41: how to customize all those elements when xpaths are changed on page? could you please provide some instructions in comments/readme

niklub (changes requested):

label_studio/ui_test/readme.txt
ahovhannes: This gives us to testing in different ways.
Sure we can do test for all file types by setting '+' before each file type,
but also we can do testing for chooses formats only.
For example
* if we want to do test only for images, then we will set '+' before image formats(jpg, png,...)
* if we want to test only one format 'jpg' we will just set '+' before jpg and '-' before another formats (this can be used for case, when developers changed something only for jpg format, and there is a need to do testing only for that format)

label_studio/ui_test/readme.txt
ahovhannes: I planned to add them as I will add the corresponding codes for them.

label_studio/ui_test/ui_test_label_studio.py
ahovhannes: Done (Commit: 31fb2bf5f348d79caf5dc19d6421a3f1b5258b0c)

label_studio/ui_test/ui_test_label_studio.py
ahovhannes: All elements (from each page) defined into function   "def setUp(self)"
(xpathes, css-selectors, iIDs,...  all are here)
Each element is variable, which is used into test-script.
So, when xpath will changes, we only need to do changing only in one place (into function   "def setUp(self)" )
Sure, I will add more detailed instructions into readme file(on my next commit)

label_studio/ui_test/ui_test_label_studio.py
niklub: Hi! what about this? Looking forward to getting
- launchable CI with parameters
- customization on xpath (since current master version is outdated, we need to adapt these scripts to existing pages somehow)

PR #665, opened 2021-03-17
Fix building url for imported from azure images

triklozoid:


niklub:
Let's add this commit https://github.com/heartexlabs/label-studio/commit/fe32a796126d391a650e8d973be4495ed26913e0 and run test for the current fix

PR #677, opened 2021-03-18
Feature/localfiles storage

triklozoid:


label_studio/core/views.py
```diff
@@ -120,3 +122,13 @@ def samples_time_series(request):
...
     response['filename'] = filename
     return response
+
+
+def localfiles_data(request):
+    """Serving files for LocalFilesImportStorage"""
+    path = request.GET.get('d')
+    local_serving_allowed = get_bool_env('LOCAL_FILES_SERVING_ENABLED', default=False)
+    if local_serving_allowed and path and request.user.is_authenticated:
+        return serve(request, path, document_root='/')
```
makseq, line 132: I'd like to have here a more narrow path, maybe it's better to have it in env vars.

label_studio/io_storages/localfiles/models.py
```diff
@@ -0,0 +1,79 @@
...
+                if regex and not regex.match(key):
+                    logger.debug(key + ' is skipped by regex filter')
+                    continue
+                yield file.name
+
+    def get_data(self, key):
+        if self.use_blob_urls:
+            # include self-hosted links pointed to local resources via /data/filename?d=<path/to/local/dir>
+            path = str(Path(self.path) / key)
+            return {settings.DATA_UNDEFINED_NAME: f'/data?d={path}'}
```
makseq, line 48: As far as I remember we already have /data endpoint, it's better to make something like /data/local-files

label_studio/io_storages/urls.py
```diff
@@ -62,6 +63,13 @@
     path('export/redis/<int:pk>', RedisExportStorageDetailAPI.as_view(), name='export-storage-redis-detail'),
     path('export/redis/validate', RedisExportStorageValidateAPI.as_view(), name='export-storage-redis-validate'),
     path('export/redis/form', RedisExportStorageFormLayoutAPI.as_view(), name='export-storage-redis-form'),
+
+    # Local files
+    path('localfiles', LocalFilesImportStorageListAPI.as_view(), name='storage-localfiles-list'),
+    path('localfiles/<int:pk>', LocalFilesImportStorageDetailAPI.as_view(), name='storage-localfiles-detail'),
```
makseq, line 69: Maybe local-files?

makseq (changes requested):

label_studio/io_storages/localfiles/models.py
```diff
@@ -0,0 +1,79 @@
...
+        logger.debug(f'Creating new object on {self.__class__.__name__} Storage {self} for annotation {annotation}')
+        ser_annotation = AnnotationSerializer(annotation).data
+        with transaction.atomic():
+            # Create export storage link
+            link = LocalFilesExportStorageLink.create(annotation, self)
+            try:
+                with open(link.key, mode='w') as f:
+                    json.dump(ser_annotation, f, indent=2)
+            except Exception as exc:
+                logger.error(f"Can't export annotation {annotation} to GCS storage {self}. Reason: {exc}", exc_info=True)
```
smoreface, line 71: #
```suggestion
                logger.error(f"Can't export annotation {annotation} to local storage {self}. Reason: {exc}", exc_info=True)
```

label_studio/core/views.py
triklozoid: Added LOCAL_FILES_DOCUMENT_ROOT variable to config document root value.

label_studio/io_storages/localfiles/models.py
triklozoid: Changed to `/data/local-files`

label_studio/io_storages/urls.py
triklozoid: We can't do it, because we use dynamic url structure based on python module names and we'll get incompatibility between `local-files`(denied in module name) and `local_files`(denied in urls).

label_studio/io_storages/localfiles/models.py
triklozoid: Fixed, thanks!

PR #680, opened 2021-03-19
Fix migration predictions; check upload exists

triklozoid:


label_studio/core/old_ls_migration.py
```diff
@@ -67,17 +67,14 @@ def _migrate_tasks(project_path, project):
...
-                    for prediction in predictions_data['predictions']:
-                        task_prediction = Prediction(result=prediction['result'], task=task)
-                        with suppress_autotime(task_prediction, ['created_at']):
-                            task_prediction.created_at = datetime.datetime.fromtimestamp(
-                                prediction['created_at'], tz=datetime.datetime.now().astimezone().tzinfo
-                            )
-                            task_prediction.save()
+            predictions_data = task_data.get('predictions', [])
+            for prediction in predictions_data:
+                task_prediction = Prediction(result=prediction['result'], task=task)
```
niklub, line 72: `score=prediction.get('score')` should be also included

niklub (changes requested):

label_studio/core/old_ls_migration.py
triklozoid: Fixed.

PR #698, opened 2021-03-22
Feature/user active organization

triklozoid:


label_studio/core/utils/common.py
```diff
@@ -632,3 +632,16 @@ def collect_versions(force=False):
     return result
 
 
+def get_organization_from_request(request):
```
niklub, line 635: Do we remove this in the future once all sessions will be updated? If so, may be we can leave some TODO message to clear this function in a future release?

niklub (changes requested):

niklub:
Why did we add `tasks_and_annotations.json` ? And also do we have some tests that cover this feature?

label_studio/core/utils/common.py
triklozoid: Yes, we will remove it. Added TODO.

triklozoid:
I removed `tasks_and_annotations.json` and added it to `.gitignore` to prevent this thing. Also added simple test for active_organization.

niklub (approved):

label_studio/organizations/middleware.py
```diff
@@ -16,78 +14,12 @@ def __init__(self, get_response):
 
     def __call__(self, request):
         org = Organization.objects.first()
+        user = request.user
+        if user and user.is_authenticated and user.active_organization is None:
+            user.active_organization = org
+            user.save(update_fields=['active_organization'])
         if org is not None:
```
makseq, line 21: @triklozoid why didn't you remove this?

triklozoid: I want to do it in next release, when all users will get their active_organizations. 

In out case we can remove it now I think, but for some more complicated things with more than one backend running simultaneously it can be good thing to not break something during deploy.

PR #703, opened 2021-03-23
User invite via link

triklozoid:


label_studio/users/views.py
```diff
@@ -77,13 +78,19 @@ def user_signup(request):
...
     next_page = next_page if next_page else reverse('projects:project-index')
     user_form = forms.UserSignupForm()
     organization_form = OrganizationSignupForm()
 
     if user.is_authenticated:
         return redirect(next_page)
 
+    organization = Organization.objects.first()
+    if settings.DISABLE_SIGNUP_WITHOUT_LINK is True:
+        if not(token and organization and token == organization.token):
```
niklub, line 91: I think this is wrong. We should check token on auth (POST), not GET

niklub (changes requested):

label_studio/frontend/src/pages/PeoplePage/SelectedUser.js
```diff
@@ -0,0 +1,67 @@
...
+        </Elem>
+      ))}
+    </Elem>
+  );
+};
+
+export const SelectedUser = ({ user, onClose }) => {
+  const fullName = [user.first_name, user.last_name].filter(n => !!n).join(" ").trim();
+
+  return (
```
niklub, line 23: what about this compound design element?

image

niklub (changes requested):

label_studio/users/views.py
triklozoid: Moved check inside POST processing and fixed tests.

label_studio/frontend/src/pages/PeoplePage/PeoplePage.js
```diff
@@ -0,0 +1,143 @@
...
+        <Space spread>
+          <Space>
+            <Button style={{width: 170}} onClick={() => updateLink()}>
+              Reset Link
+            </Button>
+          </Space>
+          <Space>
+            <Button primary style={{width: 170}} onClick={copyLink}>
+              Copy Link
+              {copied && (<LsCheck style={{marginLeft: 10}}/>)}
```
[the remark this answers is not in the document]
nicholasrq: i've added "Copied!" text to the button itself. it replaces "Copy Link" for 1.5s and then changes back. initially I was thinking to add the text before the button, but realized that according to the design we will have some other buttons there and this modal will be incompatible with LSE version of the same modal (we should always keep that in mind during the design and implementation)

niklub (approved):

PR #708, opened 2021-03-24
Make sample task creation and config validation through POST instead of GET

nicholasrq:
Sample task generation as well as label config validation are done with GET. It can break potentially with large config. Move those requests to POST.


label_studio/projects/api.py
```diff
@@ -500,9 +499,10 @@ class ProjectLabelConfigValidateAPI(generics.RetrieveAPIView):
     @swagger_auto_schema(tags=['Projects'], operation_summary='Validate a label config', manual_parameters=[
                             openapi.Parameter(name='label_config', type=openapi.TYPE_STRING, in_=openapi.IN_QUERY,
                                               description='labeling config')])
-    def get(self, request, *args, **kwargs):
+    def post(self, request, *args, **kwargs):
         project = self.get_object()
-        label_config = self.request.query_params.get('label_config')
+        json_data = json.loads(request.body)
```
niklub, line 506: why not 
```
label_config = self.request.data.get('label_config')
```

label_studio/projects/api.py
```diff
@@ -645,7 +645,9 @@ class ProjectSampleTask(generics.RetrieveAPIView):
     serializer_class = ProjectSerializer
     swagger_schema = None
 
-    def retrieve(self, request, *args, **kwargs):
-        config = request.GET.get('label_config')
+    def post(self, request, *args, **kwargs):
+        json_data = json.loads(request.body)
```
niklub, line 651: same here as above

niklub (changes requested):

label_studio/projects/api.py
makseq: I'll fix it in release/1.0.1 branch after merge

PR #723, opened 2021-03-26
Using utf-8 as default encoding

secsilm:
If not using utf-8, this will cause error on some platforms, for example Windows especially in China.


makseq:
Thank you very much for your PR!

PR #736, opened 2021-03-30
Fix/relative urls

hlomzik:
- [x] icons/images url
- [x] config api fetch
- [x] import api fetch
- [x] redirects
- [x] static error pages
- [x] check annotation templates


niklub:
@makseq can we include nginx.conf with relative URLs in this PR and test it with a running server against proposed changes ?

niklub (changes requested):
need to be tested

niklub (approved):

niklub:
lgtm

PR #740, opened 2021-04-01
Add localfiles validation; fix form layout

triklozoid:


niklub (changes requested):
Need to provide a more meaningful message about "Treat every bucket object as.." when failing to deserialize object in `get_data` (check s3.models as an example)
image

niklub (approved):

PR #741, opened 2021-04-01
Docs page meta

niklub:


smoreface (approved):
changes look good to me, and I can start adding meta_title and meta_descriptions to other pages on this branch or as I update documentation generally, lmk

smoreface:
@niklub this should have everything covered now! Not sure exactly where the meta tags for labelstud.io itself might go, so I think that's the only one I didn't catch.

PR #758, opened 2021-04-06
Fix heroku deployment

niklub:
\+ prioritize command-line argument for port wrt PORT env var


Dockerfile
```diff
@@ -18,12 +18,11 @@ RUN pip3 install -r requirements.txt && pip install uwsgi
 
 ENV DJANGO_SETTINGS_MODULE=core.settings.label_studio
 ENV LABEL_STUDIO_BASE_DATA_DIR=/label-studio/data
-ENV LABEL_STUDIO_PORT="8080"
 
 COPY . /label-studio
 RUN python3.8 setup.py develop
 
-EXPOSE ${LABEL_STUDIO_PORT}
+EXPOSE 8080
```
Arrow7000, line 25: Is this still needed? Does it do anything?

Dockerfile.heroku
```diff
@@ -0,0 +1,2 @@
+FROM heartexlabs/label-studio:latest
```
Arrow7000, line 1: If this is what Heroku always runs doesn't that mean that no matter what branch or fork you're trying to run it will always try to run the latest heartex master branch?

Maybe I'm misunderstanding how it works.

Dockerfile
```diff
@@ -18,12 +18,11 @@ RUN pip3 install -r requirements.txt && pip install uwsgi
 
 ENV DJANGO_SETTINGS_MODULE=core.settings.label_studio
 ENV LABEL_STUDIO_BASE_DATA_DIR=/label-studio/data
-ENV LABEL_STUDIO_PORT="8080"
 
 COPY . /label-studio
 RUN python3.8 setup.py develop
 
-EXPOSE ${LABEL_STUDIO_PORT}
```
Arrow7000, line 26: A I right in thinking that the this environment variable is no longer used anywhere then? In which case should the docs [here](https://github.com/heartexlabs/label-studio/blob/master/docs/source/guide/start.md#L45) and [here](https://github.com/heartexlabs/label-studio/blob/master/docs/source/guide/start.md#L68) be updated to no longer use `LABEL_STUDIO_PORT` but to use `PORT` instead?

Dockerfile
niklub: I guess it should be there in case you run docker outside heroku environment where expose step is ignored. e.g. by launching it with [readme commands](https://github.com/heartexlabs/label-studio#install-locally-with-docker)

Dockerfile.heroku
niklub: yes, and that's exactly how it worked before. For now, it won't build from the ground up during Heroku deployment.

Dockerfile
niklub: [LABEL_STUDIO_PORT](https://github.com/heartexlabs/label-studio/blob/master/docs/source/guide/start.md#run-label-studio-on-localhost-with-a-different-port) is used to run the app outside the container, so it is a primary variable to explicitely set the port during the runtime. On the other hand, `PORT` variable is somewhat specific only to Heroku deployment (may be also some other cloud providers, but haven't checked it yet), so that's why a separate `Dockerfile.heroku` is created. As I mentioned before, LS app is looking for `LABEL_STUDIO_XXX` variables first to avoid catching the wrong global configurations.

Dockerfile.heroku
Arrow7000: Sorry does that mean it's not possible to run a fork or branch on Heroku at all, only the master branch of this repo? Can I ask why?

niklub: No significant reason, if you think it should be the case, I return it back

Arrow7000: Well yes I do, otherwise it's impossible to deploy this project on Heroku with anything other than the main project's master branch, which seems like an unnecessarily harsh restriction 🤷‍♂️

Dockerfile
Arrow7000: Are you sure about that? I can't find any reference to that env var anywhere in the project other than the start.md doc
image

niklub: Yes, [there is a logic](https://github.com/heartexlabs/label-studio/blob/master/label_studio/core/utils/params.py#L75) to resolve all env variables with prefixes

Arrow7000: Ah ok thanks for clarifying!

So am I understanding it correctly that now _either_ `LABEL_STUDIO_PORT` _or_ `PORT` will work? (In which case it should work on Heroku again)

niklub: yes, but `LABEL_STUDIO_PORT` is taken first

Arrow7000 (approved):

makseq (approved):

PR #767, opened 2021-04-07
Support predictions loading from storages

makseq:
Temporary branch for the webinar
https://label-studio.slack.com/archives/CQ8LYPPJS/p1617394734275500


label_studio/io_storages/base_models.py
```diff
@@ -74,12 +74,30 @@ def _scan_and_create_links(self, link_class):
...
                 tasks_created += 1
 
+                if not task.predictions.exists():
+                    for p in predictions:
+                        prediction = Prediction(result=p['result'], score=p.get('score'),
+                                                model_version=p.get('model_version'),
+                                                task=task)
+                        prediction.save()
+                    if predictions:
+                        logger.debug(f'Create {len(predictions)} predictions for task={task}')
```
niklub, line 99: we typically print debug messages before running a code

niklub (changes requested):

label_studio/io_storages/base_models.py
makseq: If it prints after, you will never be sure wether this command has been executed or not

label_studio/io_storages/base_models.py
```diff
@@ -74,12 +74,30 @@ def _scan_and_create_links(self, link_class):
...
+                                     'you must put "data" field in the task too')
+                data = data['data']
+
             with transaction.atomic():
                 task = Task.objects.create(data=data, project=self.project)
                 link_class.create(task, key, self)
                 logger.debug(f'Create {self.__class__.__name__} link with key={key} for task={task}')
                 tasks_created += 1
 
+                if not task.predictions.exists():
```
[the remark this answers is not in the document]
makseq: if the task exists already and this is the second sync, predictions will be duplicated without this constraint

label_studio/io_storages/base_models.py
```diff
@@ -74,12 +74,30 @@ def _scan_and_create_links(self, link_class):
...
+                    for p in predictions:
+                        prediction = Prediction(result=p['result'], score=p.get('score'),
```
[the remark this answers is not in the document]
makseq: No, we need task id inside of the each prediction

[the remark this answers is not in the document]
makseq: it's better to rewrite it later with TaskSerializerBulk

label_studio/io_storages/base_models.py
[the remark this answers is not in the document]
niklub: but it seems that duplication is handled by https://github.com/heartexlabs/label-studio/blob/365dc78a50d7d029c2bb6087def4641f94bf4fd5/label_studio/io_storages/base_models.py#L72 otherwise even task creation won't work

label_studio/data_manager/actions/basic.py
```diff
@@ -37,7 +37,7 @@ def delete_tasks_annotations(project, queryset, **kwargs):
     annotations = Annotation.objects.filter(task__id__in=task_ids)
     count = annotations.count()
     annotations.delete()
-    return {'processed_items': count,
+    return {'processed_items': count, 'reload': True,
```
niklub, line 40: what's this option used for?

label_studio/io_storages/base_models.py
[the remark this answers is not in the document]
niklub: I think validating preannotations is a crucial thing, so I would move all `Prediction` instance creation in a single place to avoid code duplication in the near future

label_studio/io_storages/localfiles/models.py
```diff
@@ -49,8 +49,15 @@ def get_data(self, key):
             document_root = Path(get_env('LOCAL_FILES_DOCUMENT_ROOT', default='/'))
             relative_path = str(path.relative_to(document_root))
             return {settings.DATA_UNDEFINED_NAME: f'/data/local-files/?d={relative_path}'}
-        with open(path) as f:
-            value = json.load(f)
+
+        try:
+            with open(path, encoding='utf8') as f:
+                value = json.load(f)
+        except (UnicodeDecodeError, json.decoder.JSONDecodeError):
```
niklub, line 56: It has been already added in [a separate PR](https://github.com/heartexlabs/label-studio/pull/801)

label_studio/io_storages/s3/models.py
```diff
@@ -143,8 +143,8 @@ def get_data(self, key):
         # read task json from bucket and validate it
         _, s3 = self.get_client_and_resource()
         bucket = s3.Bucket(self.bucket)
+        obj = s3.Object(bucket.name, key).get()['Body'].read().decode('utf-8')
```
niklub, line 146: why moving this outside?

label_studio/data_manager/actions/basic.py
makseq: for dm reloading

niklub (approved):

PR #769, opened 2021-04-07
Add default values for arguments description

triklozoid:


smoreface (approved):
Updated Defaults to Default and encoded 1 additional default, looks great! thank you!

PR #781, opened 2021-04-09
Add few doc fixes on ML & local files storage setup

niklub:


smoreface (approved):
Updated text of the warning for local file storage, otherwise looks good!

PR #797, opened 2021-04-12
Add info about local files and Docker setup

smoreface:
* Environment variable updates 
* Docker examples
* Cross-references
* Improved info about local file storage
* Reorganized flow of local file directory import


docs/source/guide/storage.md
```diff
@@ -295,11 +304,11 @@ In the Label Studio UI, do the following to set up the connection:
 2. For a specific project, open **Settings > Cloud Storage**.
 3. Click **Add Source Storage**.  
 4. In the dialog box that appears, select **Local Files** as the storage type.
-5. Specify the name of the local directory.
+5. Specify the name of the local directory. Label Studio assumes a root directory of `\`. To specify a different directory, see [Run Label Studio on Docker and use local storage](start.html#Run_Label_Studio_on_Docker_and_use_local_storage).
```
niklub, line 307: again `/` , and maybe mentioning here that "local directory" is relative to the root directory? Or does it already stem from it?

docs/source/guide/storage.md
```diff
@@ -311,5 +320,13 @@ You can specify additional parameters from the Label Studio UI.
 | regex | Specify a regular expression to filter directory objects. Use ".*" to collect all objects. | Skips all directory objects. |
 | use_blob_urls | If true, treat every directory object as a source file. Use for resources like JPG, MP3, or similar file types. If false, directory objects are interpreted as tasks in Label Studio JSON format with one object per task. | false |
 
+### Set up local storage with Docker
```
niklub, line 323: why it nearly duplicates what's written in start.md ?

niklub (changes requested):

docs/source/guide/storage.md
smoreface: I think it's worth making that clearer, will update thanks

docs/source/guide/storage.md
smoreface: good catch, I'll link to the start.md file instead here.

docs/source/guide/start.md
```diff
@@ -91,6 +93,17 @@ nginx:
 
 To run Label Studio on Docker with a host and subdomain, you can refer to the example `deploy/dockerfiles/subpath.example.yml` Docker YAML file. To customize it for your environment, manually modify the `nginx/subdomain.example.simple.conf` NGINX configuration file, the relevant Label Studio environment variables, and the PostgreSQL database settings for your environment. 
 
+## Run Label Studio on Docker and use local storage
+To run Label Studio on Docker and reference persistent local storage directories, mount those directories as volumes when you start Label Studio and specify any environment variables you need.
+
+```bash
+docker run -it -p 8080:8080 --env LABEL_STUDIO_LOCAL_FILES_SERVING_ENABLED=true -v <path/to/local/file/directory>:/label-studio/my-files heartexlabs/label-studio:latest label-studio
```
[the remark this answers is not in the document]
smoreface: rewrote these examples entirely, please check again

docs/source/guide/start.md
```diff
@@ -91,6 +93,24 @@ nginx:
...
+```bash
+docker run -it -p 8080:8080 --env LABEL_STUDIO_LOCAL_FILES_SERVING_ENABLED=true,LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT="/var/lib/docker/volumes/" --mount type=volume,source=ls-task-files,target=/label-studio/data-store heartexlabs/label-studio:latest label-studio
+```
+Label Studio assumes the root of the local file directory path is `/`. To specify a different file directory root, use the `LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT` environment variable. In this example, the root is updated to specify the path to where Docker stores volumes. 
+
+You can also reference a local file directory of `/etc/apps/label-studio/task-files` with the source argument like this example, assuming you are in the `opt` directory when you run the following command:
+```bash
+docker run -it -p 8080:8080 --env LABEL_STUDIO_LOCAL_FILES_SERVING_ENABLED=true --mount type=bind,source="$(pwd)/etc/apps/label-studio/task-files",target=/label-studio/data-store heartexlabs/label-studio:latest label-studio
```
smoreface, line 107: @niklub a bind mount is required to access the local file system from Docker. I've tested this out on my local machine (referencing a folder on my desktop) but haven't been able to get the e2e flow working successfully. Would appreciate it if you test this out as well.

niklub: The right command is
```
docker run -it -p 8080:8080 -v `pwd`/mydata:/label-studio/data \
 --env LABEL_STUDIO_LOCAL_FILES_SERVING_ENABLED=true --env LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT=/label-studio/files -v `pwd`/myfiles:/label-studio/files \
heartexlabs/label-studio:latest label-studio
```

so it basically mounts the local `./myfiles` directory from the host machine to the internal `/label-studio/files`. And then `LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT=/label-studio/files` tells to scan only this directory

niklub: Also worth mentioning about that it's highly recommended to always secure docker container by explicit LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT

docs/source/guide/tasks.md
```diff
@@ -191,10 +192,21 @@ this is a second task
...
 
 ## Import data from a local directory
+
 To import data from a local directory, you have two options:
-- Add the file directory as a source or target [local storage](storage.html#Local-storage) connection in the Label Studio UI. 
 - Run a web server to generate URLs for the files, then upload a file that references the URLs to Label Studio. 
+- Add the file directory as a source or target [local storage](storage.html#Local-storage) connection in the Label Studio UI.
+
+Reference tasks in a local directory using the following syntax in the Label Studio JSON format. For example, to reference an image file `1.png`: 
+```
```
niklub, line 201: why do we need this duplicated with https://github.com/heartexlabs/label-studio/pull/797/files#diff-446481c51c42df4652b3b424cf7c3f1e69f4cfea206d82a3ec68e51e7f7474cdR296 ?

docs/source/guide/tasks.md
```diff
@@ -204,7 +216,11 @@ Use that script to do the following:
    The script collects the links to the files provided by that HTTP server and saves them to a `files.txt` file with one URL per line. 
 3. Import the file with URLs into Label Studio using the Label Studio UI. 
 
-To import tasks from a file directory on a remote server, make sure the remote server is accessible to the server running Label Studio and set up the directory as a [local storage](storage.html#Local-storage) connection. 
+
+### Add the file directory as source storage in the Label Studio UI
+To import tasks from a file directory on a remote or local server, make sure the remote server is accessible to the server running Label Studio and set up the directory as a [local storage](storage.html#Local-storage) connection. See [local storage](storage.html#Local-storage).
```
niklub, line 221: I don't think this could be a real use case with connecting file storage from a remote server. Better to avoid that

niklub (changes requested):

docs/source/guide/tasks.md
smoreface: good point, I'll remove it from this spot. I think I had it here so that people using the local web server method would have this reference, but I think that will be less common anyway.

docs/source/guide/tasks.md
smoreface: I'll remove it! thanks

niklub (approved):

PR #799, opened 2021-04-13
Unexpected tabs reset. Broken label_config saving. Another small fixes

makseq:


label_studio/projects/api.py
```diff
@@ -195,11 +195,14 @@ def delete(self, request, *args, **kwargs):
...
     def patch(self, request, *args, **kwargs):
         project = self.get_object()
-        label_config = self.request.query_params.get('label_config')
+        label_config = self.request.data.get('label_config')
 
         # config changes can break view, so we need to reset them
-        if parse_config(label_config) != parse_config(project.label_config):
-            View.objects.filter(project=project).all().delete()
+        try:
+            if label_config and parse_config(label_config) != parse_config(project.label_config):
```
niklub, line 202: there is another function in this file called `config_essential_data_has_changed` - it's better to modify and reuse it here

label_studio/tasks/models.py
```diff
@@ -255,7 +255,6 @@ class Annotation(models.Model):
     objects = AnnotationManager()
     tracker = FieldTracker(fields=['ground_truth', 'result'])
 
-    state = JSONField('state', null=True, default=dict, help_text='Editor state (system data)')
```
niklub, line 258: why?

label_studio/tasks/api.py
```diff
@@ -139,38 +136,6 @@ def put(self, request, *args, **kwargs):
         return super(TaskAPI, self).put(request, *args, **kwargs)
 
 
-class TaskCancelAPI(APIView):
```
niklub, line 142: hm, and how Skip is handled now?

niklub (changes requested):

label_studio/core/old_ls_migration.py
```diff
@@ -42,15 +42,15 @@ def suppress_autotime(model, fields):
 def _migrate_tasks(project_path, project):
     """ Migrate tasks from json file to database objects"""
     tasks_path = project_path / 'tasks.json'
-    with io.open(os.path.abspath(tasks_path)) as t:
+    with io.open(os.path.abspath(tasks_path), encoding='utf-8') as t:
```
niklub, line 45: do we have a github issue patched by this fix?

label_studio/tasks/models.py
makseq: Because we don't need it for more over a year

label_studio/tasks/api.py
makseq: AnnotationListAPI

label_studio/core/old_ls_migration.py
makseq: I didn't see it, but old projects with non utf8 chars won't be migrated

label_studio/tasks/models.py
niklub: I meant that we can remove later all useless stuff in one migration, no rush for that now

niklub (approved):

PR #802, opened 2021-04-13
Fix API doc generation issues

niklub:


label_studio/io_storages/azure_blob/api.py
```diff
@@ -9,33 +9,105 @@
 
 
 class AzureBlobImportStorageListAPI(ImportStorageListAPI):
+    """
+    get:
+    Get all Azure import storages
```
niklub, line 14: Maybe we can group titles based on prefixes like `Azure: Import storage list`

label_studio/users/api.py
```diff
@@ -45,30 +45,52 @@ def avatar(self, request, pk):
             request.user.save()
             return Response(status=204)
 
-    @swagger_auto_schema(tags=['Users'], operation_summary='Save user details')
+    @swagger_auto_schema(
+        tags=['Users'],
+        operation_summary='Save user details',
+        operation_description='Save user details',
```
niklub, line 51: Please provide more meaningful `operation_description` whenever needed

label_studio/io_storages/s3/api.py
```diff
@@ -12,33 +14,108 @@ class S3ImportStorageListAPI(ImportStorageListAPI):
     queryset = S3ImportStorage.objects.all()
     serializer_class = S3ImportStorageSerializer
 
+    @swagger_auto_schema(
```
makseq, line 17: @niklub Seems we should stop to write `@swagger_auto_schema` before all empty methods like get, post, ... and do it before class implementation, like here: 
https://github.com/heartexlabs/label-studio/blob/master/label_studio/data_manager/api.py#L53

label_studio/users/api.py
```diff
@@ -45,30 +45,53 @@ def avatar(self, request, pk):
...
-    @swagger_auto_schema(auto_schema=None)
+    @swagger_auto_schema(
+        tags=['Users'],
+        operation_summary='Get user info',
+        operation_description='Get info about a specific Label Studio user.'
+    )
     def retrieve(self, request, *args, **kwargs):
         return super(UserAPI, self).retrieve(request, *args, **kwargs)
 
     @swagger_auto_schema(auto_schema=None)
```
makseq, line 86: I can't see PATCH method for User. It's better to have swagger description for partial_update() [PATCH] instead of update() [PUT] method, or duplicate it. 
image

label_studio/io_storages/azure_blob/api.py
```diff
@@ -9,33 +9,105 @@
 
 
 class AzureBlobImportStorageListAPI(ImportStorageListAPI):
```
makseq, line 11: Azure storage docs doesn't appear. The same problems with 
* GCS
* Redis
* Localfiles
image

makseq (changes requested):

label_studio/io_storages/azure_blob/api.py
niklub: are you sure using the latest commit in this branch? because all these titles have been fixed here

label_studio/io_storages/s3/api.py
niklub: and what are the benefits?

label_studio/users/api.py
niklub: I kept them both replacing "Save" -> "Update" in PATCH method (but still not sure about the purpose of PUT). Can you describe it then @smoreface will edit the descriptions if needed

makseq (approved):

PR #821, opened 2021-04-19
Use default viewset logic for user patch

triklozoid:


.gitignore
```diff
@@ -41,6 +41,7 @@ archive
 *.rtl.min.css
 *.idea/*
 .vscode/*
+pyrightconfig.json
```
niklub, line 44: let's follow ignore-everything-except pattern (not necessarily in this commit)

niklub (approved):

PR #829, opened 2021-04-20
FIX: S3 Import issue causing nested jsons

yashwantkeswani:
```Fix: JSONs which were present in S3 tasks were not being parsed correctly. They were getting created as nested JSONs where the entire data went into the data field. Also there would have been bugs, had `predictions` or `annotations` field been found in the json field. ```

### DIFF
--- a/label_studio/io_storages/base_models.py
+++ b/label_studio/io_storages/base_models.py
@@ -93,7 +93,6 @@ class ImportStorage(Storage):
                 if 'data' not in data:
                     raise ValueError('If you use "predictions" field in the task, '
                                      'you must put "data" field in the task too')
-                data = data['data']
 
             # annotations
             annotations = data.get('annotations', [])
@@ -101,6 +100,8 @@ class ImportStorage(Storage):
                 if 'data' not in data:
                     raise ValueError('If you use "annotations" field in the task, '
                                      'you must put "data" field in the task too')
+
+            if 'data' in data and isinstance(data['data'], dict):
                 data = data['data']
 
             with transaction.atomic():

### CODE REVIEW

File: label_studio/io_storages/base_models.py [hunk 1]
```diff
@@ -93,14 +93,15 @@ def _scan_and_create_links(self, link_class):
...
                                      'you must put "data" field in the task too')
-                data = data['data']
 
             # annotations
             annotations = data.get('annotations', [])
             if annotations:
                 if 'data' not in data:
                     raise ValueError('If you use "annotations" field in the task, '
                                      'you must put "data" field in the task too')
+
```
makseq, line 103: I'd like to place it before predictions (line 91)

yashwantkeswani: It would possibly break the flow in that case. 
since the expected format as stated in the docs is as follows

```
{
  # "data" must contain the "my_text" field defined by labeling config,
  # and can optionally include other fields
  "data": {
    "my_text": "Opossums are great",
    "ref_id": 456,
    "meta_info": {
      "timestamp": "2020-03-09 18:15:28.212882",
      "location": "North Pole"
    } 
  },

  # annotations are the list of annotation results matching the labeling config schema
  "annotations": [{
    "result": [{
      "from_name": "sentiment_class",
      "to_name": "message",
      "type": "choices",
      "value": {
        "choices": ["Positive"]
      }
    }]
  }],

  # "predictions" are pretty similar to "annotations" 
  # except that they also include some ML-related fields like a prediction "score"
  "predictions": [{
    "result": [{
      "from_name": "sentiment_class",
      "to_name": "message",
      "type": "choices",
      "value": {
        "choices": ["Neutral"]
      }
    }],
  # score is used for active learning sampling mode
    "score": 0.95
  }]
}```

makseq: Sorry, you are right.

makseq (approved):

PR #835, opened 2021-04-21
Fix localfiles get data link

jackklpan:
When `LABEL_STUDIO_HOST` is set with a subpath, the local file importer's `get_data` is not working due to the absolute path `/data/`.
截圖 2021-04-21 下午3 43 29

Change it to a relative path can work.

Signed-off-by: Jack Pan <jackpan@infuseai.io>


label_studio/io_storages/localfiles/models.py
```diff
@@ -50,10 +50,10 @@ def iterkeys(self):
     def get_data(self, key):
         path = Path(key)
         if self.use_blob_urls:
-            # include self-hosted links pointed to local resources via /data/filename?d=<path/to/local/dir>
+            # include self-hosted links pointed to local resources via data/local-files?d=<path/to/local/dir>
             document_root = Path(get_env('LOCAL_FILES_DOCUMENT_ROOT', default='/'))
             relative_path = str(path.relative_to(document_root))
-            return {settings.DATA_UNDEFINED_NAME: f'/data/local-files/?d={relative_path}'}
+            return {settings.DATA_UNDEFINED_NAME: f'data/local-files/?d={relative_path}'}
```
makseq, line 56: I'm afraid `data/local-files/?d={relative_path}` could be non working for some web location of LS. For example, the browser will generate link for `http://your.domain.com/prefix/projects/1` => `http://your.domain.com/prefix/projects/1/data/local-files/?d=...`. Am I wrong? 

What do you think about this variant?
```
f'{settings.HOSTNAME}/data/local-files/?d={relative_path}'
```

makseq (changes requested):

label_studio/io_storages/localfiles/models.py
jackklpan: You are right. Thanks for your feedback. I have made the change.

makseq (approved):

PR #836, opened 2021-04-21
fix template assets host path

even-wei:
1. patch template assets host path
2. fix typo of `rel="shortcut icon"` in `base.html`

Signed-off-by: Even Wei <evenwei@infuseai.io>


makseq (approved):

makseq:
Thank you very much for your help!

PR #853, opened 2021-04-27
Updating components and resources

nicholasrq:


niklub (changes requested):
need to check multi annotations UI issues (editor window comes out of screen)
image

niklub:
I can select and Submit predictions, and nothing happens - which is very confusing
image

niklub:
Looks better now - but how can I select one of the predictions to serve as an annotation?
image

niklub:
Also minor but panel text is slightly truncated
image

niklub (approved):

PR #858, opened 2021-04-28
Fix ml tutorials

makseq:


label_studio/ml/examples/mmdetection/mmdetection.py
```diff
@@ -3,9 +3,9 @@
 
 from mmdet.apis import init_detector, inference_detector
 
-from label_studio.ml import LabelStudioMLBase
```
niklub, line 6: `ml/examples` dir should be totally removed from repo, made in another branch but could be done here instead

docs/source/tutorials/object-detector.md
```diff
@@ -1 +0,0 @@
-../../../label_studio/ml/examples/mmdetection/README.md
```
niklub, line 1: These tutorials are outdated - need to sync with what we have in label-studio-ml-backend/examples

niklub (changes requested):

label_studio/ml/examples/mmdetection/mmdetection.py
makseq: I'll do it here

smoreface (approved):

niklub (approved):

PR #872, opened 2021-05-03
Create mlflow_example.py

KonstantinKorotaev:
https://heartex.atlassian.net/browse/HTX-1251
Adding example of working with MLFlow


niklub (changes requested):
2 things needed to be done:

1. Use example data
2. Make 2nd pass to demonstrate the full ML cycle: raw data -> label -> train -> eval -> predict unlabeled -> relabel -> train -> eval

Issue #918, opened 2021-05-13
Inconsistent build from source commands github vs docs


JustasB:
**Describe the bug**
In [docs](https://labelstud.io/guide/install.html#Install-from-source) it says to use this command to build:

`python setup.py develop`

But in [github readme](https://github.com/heartexlabs/label-studio#install-for-local-development), it says to use this command:

`pip install -e .`

**Expected behavior**
Unless this is by design, the commands should be the same

makseq:
@JustasB Thank you very much for your suggestion. 

@smoreface Could you leave only `pip install -e .` in the docs please?

smoreface:
Yes, I'll fix this up! thanks @JustasB

PR #929, opened 2021-05-17
Update install docs to fix issue #918

smoreface:
Fix issue #918


docs/source/guide/install.md
```diff
@@ -110,17 +110,12 @@ If you want to use nightly builds or extend the functionality, consider download
 ```bash
 git clone https://github.com/heartexlabs/label-studio.git
 cd label-studio
-python setup.py develop
+# Install all package dependencies
+pip install -e .
+# Start the server in development mode at http://localhost:8000
```
niklub, line 115: It will start at http://localhost:8080

niklub (changes requested):

docs/source/guide/install.md
niklub: also we need to run DB migrations
```
python label_studio/manage.py migrate
```
before runserver

smoreface: Updated README and install doc with this feedback

niklub (approved):

PR #956, opened 2021-05-20
Various developer improvements

### LINKED ISSUE #633: Add Microsoft ADLS2
**Is your feature request related to a problem? Please describe.**
No

Add Microsoft ADLS2 as a storage provider, which combines ADLS Gen1 and Azure Blob Storage.
Apparently requires [azure-storage-file-datalake](https://pypi.org/project/azure-storage-file-datalake/) instead of [azure-storage-blob](https://pypi.org/project/azure-storage-blob/).

**Describe alternatives you've considered**
I haven't tested the current implementation as I need this for an upcoming project.

**Additional context**

twsl:
Started to fix #633, but I noticed, that I misunderstood something: since ADLS2 is based on blob containers, the current interface works just fine with ADLS2 storage for accessing files.
But as I started setting up the dev environment before actually checking the function, I still have some code to contribute.
I think, some people might find VS Code Devcontainers useful for developing further features without huge struggles to setup the dev env. Feel free to cherry pick.


twsl:
@niklub @nicholasrq @makseq whats the review process like? Do I need to provide anything else?

twsl:
???

makseq:
@farioas please, review it.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
```
farioas, line 1: Not required. We have strict list of dependencies which are well known and covered by tests. Any new update initiated by dependabot requires code/tests change and etc.

Dockerfile
```diff
@@ -1,14 +1,15 @@
 # Building the main container
-FROM ubuntu:20.04
+FROM python:3.8-buster
```
farioas, line 2: Python image based on debian, while we prefer ubuntu, because it updates much faster do we receive update for critical vulnerabilities in packages frequently.

docker-compose.yml
```diff
@@ -47,13 +47,14 @@ services:
           nocopy: true
 
   db:
-    image: postgres:11.5
+    image: postgres:latest
```
farioas, line 50: we do prefer strict version definition

docker-compose.yml
```diff
@@ -47,13 +47,14 @@ services:
...
     hostname: db
     restart: always
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
     volumes:
-      - ${POSTGRES_DATA_DIR:-./postgres-data}:/var/lib/postgresql/data
```
farioas, line 56: This is breaking change, all existing data will be lost. Don't do that please.

farioas (changes requested):
Looks like only changes from .devcontainer/ folder are feasible.

docker-compose.yml
twsl: Valid objection, however I do not see any other solution supporting windows. Any ideas how to move forward?

docker-compose.yml
twsl: According to [postgres](https://www.postgresql.org/support/versioning/) the latest minor is recommended, which is equal to the `latest` tag regarding the latest major.

Dockerfile
twsl: talking about the security of debian vs ubuntu is a discussion of its own imo. When looking at docker images I'm personally also concerned about speed, stability and weight of the image itself. To me it seems counter intuitive not to use the official python image for something based on python. But this is a valid [dicussion](https://pythonspeed.com/articles/base-image-python-docker-images/) and I'm no expert by any means on this topic.

.github/dependabot.yml
twsl: I use dependabot to notify me of updates and move forward if tests don't fail as not every update requires changes. Breaking changes should be catched by the pipeline. To me this is a good way to automatically move forward with new versions and reduce technical debt. I thought adding it would be consistent as it is enabled on the frontend repo.

twsl:
The devcontainer is the most important change in this PR to me as it enables codespaces/devcontainers. As stated in the original message: feel free to cherry pick 😄

twsl:
If the devcontainer is the only thing you'd like to merge, it's probably easier to create a new PR with just this change, right?

.github/dependabot.yml
farioas: Frontend lifecycle not strict to backend lifecycle. This how the technologies evolves. We cannot invest time of our backend engineers to fix non stable versions each and every time after minor release of some small pypi dependency.

Dockerfile
farioas: You can compare it by yourself https://docs.docker.com/engine/scan/:
```
docker scan python:3.8-buster
docker scan ubuntu:20.04
```

My results are:
```
python:3.8-buster -- Tested 431 dependencies for known vulnerabilities, found 381 vulnerabilities.
ubuntu:20.04 -- Tested 93 dependencies for known vulnerabilities, found 16 vulnerabilities.
```

farioas:
I'd recommend to adjust the existing pull-request, since we already have a lot of abandoned pull-requests 😄

Dockerfile
twsl: I do understand that your main priority seems to be the security aspect and it's perfectly fine to prefer ubuntu as a base image. I'd just like to add, that basing your decision on number of CVEs alone may may miss [this](https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves) aspect.

farioas: I've just showed you an example that is free to use, we have other "paid" security scanners which shows other results, but the overall comparison is the same.

twsl:
@farioas might have butchered the commit history a bit while trying to rebase and merge. Please use squash.

farioas (approved):

PR #974, opened 2021-05-25
Update API docs with more help

smoreface:
Adding a few examples and more detail to some more commonly-used endpoints


label_studio/core/settings/base.py
```diff
@@ -255,7 +255,9 @@
     'SECURITY_DEFINITIONS': {
         'token': {
             'type': 'token',
-            'url': '/business/account/#/token'
+            'name': 'Token',
+            'in': 'header',
+            'url': '/user/account'
```
smoreface, line 260: Updated the URL to match where in the product the token gets retrieved and added more details here in the hopes that it would make the https://labelstud.io/api/#section/Authentication page more useful. I'm not able to build LS locally due to the Silicon chip so please make sure I didn't break anything by changing this!

label_studio/data_export/api.py
```diff
@@ -59,7 +59,13 @@ def get_queryset(self):
...
         tags=['Export'],
         operation_summary='Export annotations',
-        operation_description='Export annotated tasks as a file in a specific format.',
+        operation_description='Export annotated tasks as a file in a specific format.'
+                              'For example, to export JSON annotations for a project to a file called annotations.json,'
+                              'run the following from the command line:'
+                              '```bash'
+                              'curl -X GET https://localhost:8080/api/projects/{id}/export?exportType=JSON '
+                              '-H \'Authorization: Token abc123\' --output annotations.json'
+                              '```',
```
smoreface, line 68: No idea if the syntax will show up properly when it's inside the description like this vs part of a comment but 🤞

label_studio/users/api.py
```diff
@@ -66,7 +66,7 @@ def list(self, request, *args, **kwargs):
     @swagger_auto_schema(
         tags=['Users'],
         operation_summary='Create new user',
-        operation_description='Create a new user for Label Studio.',
+        operation_description='Create a user in Label Studio.',
         request_body=UserSerializer
```
smoreface, line 70: request_body is incorrect for this API endpoint, says requires org_membership, so something in the serializers needs to be updated, @makseq

label_studio/users/api.py
```diff
@@ -75,16 +75,16 @@ def create(self, request, *args, **kwargs):
...
         operation_summary='Get user info',
-        operation_description='Get info about a specific Label Studio user.'
+        operation_description='Get info about a specific Label Studio user, based on the user ID.'
     )
     def retrieve(self, request, *args, **kwargs):
         return super(UserAPI, self).retrieve(request, *args, **kwargs)
 
     @swagger_auto_schema(
         tags=['Users'],
         operation_summary='Update user details',
```
smoreface, line 85: As an administrator, I get an error that I don't have permissions to make these changes on a different user, so then who does have permissions?

label_studio/users/api.py
```diff
@@ -48,8 +48,8 @@ def avatar(self, request, pk):
     @swagger_auto_schema(
         tags=['Users'],
         operation_summary='Save user details',
```
smoreface, line 50: As an administrator, I get an error that I don't have permissions to make these changes on a different user, so then who does have permissions?

docs/source/guide/api.md
```diff
@@ -18,20 +18,39 @@ See the [API reference documentation](/api) for further guidance and interactive
...
+Create a project and set up the labeling interface in Label Studio using the API. See the [Create new project API endpoint documentation](/api#operation/projects_create).
+
+If you want to make sure the configuration for your labeling interface is valid before submitting it using the API, you can use the [validate label config](/api#operation/projects_validate_create) API endpoint.
+
+### Import tasks using the API
+
+To import tasks using the API, make sure you know the project ID that you want to add tasks to. See additional examples and parameter descriptions in the [import data endpoint documentation](/api#operation/projects_import_create)
+
+### Retrieve tasks
+Retrieve a paginated list of tasks for a specific project. See details and parameters in the [list project tasks endpoint documentation](/api#operation/projects_tasks_list).
```
niklub, line 43: Here we can get all tasks along with annotations, which is an alternative to export in a specific format and could be useful for downstream scenarios

label_studio/projects/api.py
```diff
@@ -84,18 +85,33 @@
 )
 
 
-class ProjectListAPI(generics.ListCreateAPIView):
-    """
-    get:
-    List your projects
-
+@method_decorator(name='get', decorator=swagger_auto_schema(
```
niklub, line 88: @smoreface let's rewrite docstrings-like helpers to these `@method_decorator`, similarly in other methods like data_import/api, it will allow us to dynamically change URL parameters  like hostname
Also note that by using multiline strings you can freely copy-paste all md content
```
"""
everything is 
multiline here and ```bash work as well
"""
```

niklub (changes requested):

docs/source/guide/api.md
```diff
@@ -18,20 +18,39 @@ See the [API reference documentation](/api) for further guidance and interactive
...
 
-For example, ofor a Label Studio instance hosted at localhost:
-```bash
-curl -X GET https://localhost:8000/api/projects/ -H 'Authorization: Token 123456789abcdefghijklmnop123456789'
-```
+See [API documentation for authentication](/api#section/Authentication).
+
+### List all projects
+
+To perform most tasks with the Label Studio API, you must specify the project ID, sometimes referred to as the `pk`. See the [List your projects API endpoint documentation](/api#operation/api_projects_list).
```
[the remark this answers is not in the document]
smoreface: Yeah I didn't make it super clear, but this is really saying "hey, you might need a pk, here's how to get all the pks that you might need". I'll add a clarifying sentence to tie together the link and my point about that!

label_studio/users/api.py
smoreface: I was using the wrong API, ahhhh.

label_studio/users/api.py
smoreface: This was using the enterprise API instead of open source but might still be an issue?

label_studio/core/settings/base.py
smoreface: cc: @makseq please check this 😬

smoreface:
@niklub / @makseq this is ready for review

label_studio/core/settings/base.py
smoreface: @niklub you said that this change doesn't work, correct?

smoreface:
@niklub / @makseq a customer just ran into an issue fixed in this PR, can I get a review on this please? I'm also happy to split it up if that makes it easier to review. Thanks!

niklub (approved):

PR #985, opened 2021-05-28
Fix loop of death on server disconnection

hlomzik:


makseq (changes requested):
Looks like it doesn't work 

https://user-images.githubusercontent.com/501780/120805204-2ee7db80-c54e-11eb-9f27-55c51f049156.mp4

makseq (approved):

hlomzik:
Frontend should be built to see this changes :(
Will be included in master

PR #990, opened 2021-05-31
Sign up turned off

DimaVinnitsa:
1. Applied black
2. Added possibility to register user with e-mail from admin panel
3. Changed HTML of the homepage


makseq:
@DimaVinnitsa Sorry, but your PR looks huge, there are plenty of file movements and other unclear things.

PR #1010, opened 2021-06-02
Doc SEO updates

smoreface:
Depends on the LSE docs! 

- Improving meta_titles and descriptions
- Removing manual scrub of label studio mentions in TOC because it reduces the likelihood of semantic links, lowering the overall SEO score for each documentation page (use a Lighthouse test in the developer console to confirm this score goes from 80 to 100 with the new topic titles)
- Updating some topic titles to reduce superfluous mentions of Label Studio
- Replacing analytics.js reference with gtag.js script per Google Analytics best practices


docs/themes/htx/layout/partials/toc.ejs
```diff
@@ -61,7 +61,7 @@
       <% } %>
     <li>
       <a href="<%- url_for(p.path) %>" class="sidebar-link<%- page.title === p.title ? ' current' : '' %><%- p.is_new ? ' new' : '' %>">
-        <%- removeLabelStudioMention(p.title) %>
```
makseq, line 64: @smoreface why did you remove this? LS table of contents looks ugly without it.

e.g.: 
image

makseq: P.s. also I'd move the "enterprise" badge after names.

makseq (changes requested):

docs/themes/htx/layout/partials/toc.ejs
smoreface: As I put in the PR description, the manual scrub of label studio mentions in TOC reduces the likelihood of semantic links, which lowers the overall SEO score for each documentation page. Since we're trying to improve our SEO, I thought it important to make this change. I used the Lighthouse feature in developer console to check the SEO score, and it goes from 80 (with Label Studio mentions removed) to 100 with this PR.  

Also, @deppp and I discussed the location of Enterprise badge, and I tried it with it before and after the topic titles, and we agreed that it should go before the topic titles.

makseq: Do we need page titles with "Label studio" only? and do we can skip it from table of content? 
In this case we have to implement different texts in .md files for content and page titles.

smoreface: Those are actually the ones that need it most, the "Secure Label Studio" and "Start Label Studio", for example, so that they are semantically meaningful. I tried to update the page titles as well to reduce unnecessary mentions of Label Studio, and also rewrite the meta_titles to be more helpful.

smoreface: Updated even more titles!

makseq (approved):

PR #1020, opened 2021-06-04
Expand tasks size. Remove tag. Move moto to requirements-test.txt

makseq:


label_studio/core/settings/base.py
```diff
@@ -310,8 +310,8 @@
 os.makedirs(EXPORT_DIR, exist_ok=True)
 
 # file / task size limits
-DATA_UPLOAD_MAX_MEMORY_SIZE = int(get_env('DATA_UPLOAD_MAX_MEMORY_SIZE', 50 * 1024 * 1024))
-TASKS_MAX_NUMBER = 250000
+DATA_UPLOAD_MAX_MEMORY_SIZE = int(get_env('DATA_UPLOAD_MAX_MEMORY_SIZE', 250 * 1024 * 1024))
+TASKS_MAX_NUMBER = 1000000
```
niklub, line 314: I don't think we need to do this by default. It's better to wrap it into `TASKS_MAX_NUMER=get_env("TASKS_MAX_NUMBER", default=250000)` then writing the message about increasing this limit (with "not recommended") right after the corresponding error pops up

niklub (changes requested):

deploy/requirements-test.txt
```diff
@@ -6,4 +6,5 @@ requests-mock==1.5.2
 pyyaml>=5.3.1
 moto==1.3.16.dev122
 tavern==1.14.0
-fakeredis==1.5.0
\ No newline at end of file
+fakeredis==1.5.0
+moto==1.3.16.dev122
```
niklub, line 10: also please make here `attrs>=19.2.0`

niklub (changes requested):

label_studio/core/settings/base.py
makseq: A lot of people flare facing this issue. But I thought about get_env too

niklub: still not sure if we enable this - could we run into another performance issue? I would shrink that now until performance enhancements

niklub (approved):

deploy/requirements-test.txt
niklub, line 10: moto is doubled

PR #1036, opened 2021-06-08
Feature/webhooks backend

chiganov:
## 2 models:
- Webhook
- WebhookAction

## API:
- `/api/webhooks/` to create/list
- `/api/webhooks/<pk>/` to delete/update/get
- `/api/webhooks/info/` to get info

## tests
- CRUD tavern tests for webhooks
- unit tests to catch webhooks

## other
`run_webhook()` - run one hook
`emit_webhooks_for_instanses()` - run hooks for an organization and send serialized instances
`@api_webhook` - decorator to wrap Create/Update methods of APIView
`@api_webhook_for_delete` - decorator to Delete method


makseq (changes requested):

label_studio/webhooks/serializers_for_hooks.py
```diff
@@ -0,0 +1,30 @@
...
+        fields: ('id',)
+
+
+class ProjectWebhookSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = '__all__'
+
+
+class TaskWebhookSerializer(serializers.ModelSerializer):
```
makseq, line 21: We need to copy logic from this serializer ExportDataSerializer: https://github.com/heartexlabs/label-studio/blob/master/label_studio/data_export/serializers.py#L26

label_studio/webhooks/utils.py
```diff
@@ -0,0 +1,82 @@
...
+                request.user.active_organization,
+                action,
+                [WebhookAction.ACTIONS[action]['model'].objects.get(id=responce.data.get('id'))]
+            )
+            return responce
+        return wrap
+    return decorator
+
+
+def api_delete_webhook(action):
```
makseq, line 69: Does it delete a webhook? Or it's a special kind of webhook for a deletion.. Maybe it's not a good name.

label_studio/webhooks/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            Q(id__in=WebhookAction.objects.filter(
+                webhook__organization=organization,
+                action=action
+            ).values_list('webhook_id', flat=True))
+        )
+    )
+    if not webhooks.exists():
+        return
+    payload = None
+    if instanses and webhooks.filter(send_payload=True).exists():
```
makseq, line 45: I'd write a comment for blocks where something non-obvious is going on :-)

label_studio/data_manager/actions/basic.py
```diff
@@ -25,9 +28,11 @@ def delete_tasks(project, queryset, **kwargs):
     :param queryset: filtered tasks db queryset
     """
     count = queryset.count()
+    tasks_ids = list(queryset.values('id'))
```
makseq, line 31: Why not `.values_list()` ?

label_studio/data_manager/actions/basic.py
```diff
@@ -48,7 +53,9 @@ def delete_tasks_annotations(project, queryset, **kwargs):
     task_ids = queryset.values_list('id', flat=True)
     annotations = Annotation.objects.filter(task__id__in=task_ids)
     count = annotations.count()
+    annotations_ids = list(annotations.values('id'))
```
makseq, line 56: Why not .values_list?

label_studio/webhooks/api.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # permission_classes = [IsAdminUser]
+
+
+@method_decorator(name='get', decorator=swagger_auto_schema(tags=['Webhooks'],))
+@method_decorator(name='put', decorator=swagger_auto_schema(tags=['Webhooks'],))
+@method_decorator(name='patch', decorator=swagger_auto_schema(tags=['Webhooks'],))
+@method_decorator(name='delete', decorator=swagger_auto_schema(tags=['Webhooks'],))
+class WebhookAPI(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Webhook.objects.all()
```
makseq, line 43: I think we should have a filter by organization here, better to implement it via get_queryset. By default organization_id = 1.

label_studio/webhooks/api.py
```diff
@@ -0,0 +1,47 @@
...
+)
+@method_decorator(
+    name='post',
+    decorator=swagger_auto_schema(
+        tags=['Webhooks'],
+        operation_description="Create a webhooks"
+    )
+)
+class WebhookListAPI(generics.ListCreateAPIView):
+    queryset = Webhook.objects.all()
```
makseq, line 26: I think we should have a filter by organization here, better to implement it via `get_queryset`. By default organization_id = 1.

label_studio/data_manager/actions/basic.py
chiganov: `.values_list()` returns a tuple, and it can't be serialized by DRF serializer. `values()` returns a dict and it can be.

label_studio/data_import/api.py
```diff
@@ -160,7 +163,9 @@ def post(self, *args, **kwargs):
     def _save(self, tasks):
         serializer = self.get_serializer(data=tasks, many=True)
         serializer.is_valid(raise_exception=True)
-        return serializer.save(project_id=self.kwargs['pk']), serializer
+        task_instances = serializer.save(project_id=self.kwargs['pk'])
+        emit_webhooks_for_instanses(self.request.user.active_organization, WebhookAction.TASK_CREATED, task_instances,)
```
makseq, line 167: Excess `,` ?

label_studio/webhooks/api.py
```diff
@@ -0,0 +1,87 @@
...
+    serializer_class = WebhookSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        return Webhook.objects.filter(organization=self.request.user.active_organization)
+
+    def perform_create(self, serializer):
+        serializer.save(organization=self.request.user.active_organization)
+
+    @swagger_auto_schema(
```
makseq, line 27: We are trying to start writing swagger_auto_schema before a class implementation, like here: 

image

or 

image

in these cases you don't need to implement dummy def post(), def get(), ... and all descriptions are collected in one place.

PR #1042, opened 2021-06-09
support subfolder for s3 when sync with prefix set.

xhuaustc:
When setting a bucket prefix, only sync the files under the folder, not files under the sub folders.
This patch is to support syncing files under the sub folders.

### DIFF
--- a/label_studio/io_storages/s3/models.py
+++ b/label_studio/io_storages/s3/models.py
@@ -106,7 +106,7 @@ class S3ImportStorage(S3StorageMixin, ImportStorage):
     def iterkeys(self):
         client, bucket = self.get_client_and_bucket()
         if self.prefix:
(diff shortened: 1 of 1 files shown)


makseq:
@xhuaustc Hi! There is something wrong with your last commit (no differences with master), please check it.

xhuaustc:
> @xhuaustc Hi! There is something wrong with your last commit (no differences with master), please check it.

Thanks.
I checked, in the commit, `, Delimiter='/'` is deleted in Line 109. It's different from master.

Please check : https://github.com/heartexlabs/label-studio/pull/1042/files

triklozoid:
@xhuaustc did you try "Recursive scan" storage option? With it storage should sync files under subfolders too.

PR #1080, opened 2021-06-18
Adding tasks agreement calculation remark

KonstantinKorotaev:


docs/source/guide/stats.md
```diff
@@ -22,6 +22,8 @@ Task agreement shows the consensus between multiple annotators when labeling the
 
 You can also see how the annotations from a specific annotator compare to the prediction scores for a task, or how they compare to the ground truth labels for a task.  
 
+Task agreement is calculated asymmetrically. For example, if one annotator adds 2 labels to a specific task, and another annotator only adds 1 label to the task, the annotators and tasks have different scores in the agreement matrices.
```
smoreface, line 25: @KonstantinKorotaev I'm a bit confused about how this relates or doesn't relate to the content about matching scores in the rest of the document, which seems to address this case. 
Also, please check that I updated which scores are different in the agreement matrices -- it is unclear to me whether this is talking about inter-annotator agreement, the per-task agreement score, both, or something else. 
Is this addressing a customer question?

smoreface (changes requested):

docs/source/guide/stats.md
KonstantinKorotaev: Canceling change as asymmetrical calculation is changed to symmetrical

PR #1102, opened 2021-06-28
Predictions CRUD api

triklozoid:


makseq (changes requested):
Also we have to change some docs here: 
https://labelstud.io/guide/ml.html#Get-predictions-from-a-model

we can't do task patch with predictions and we must use prediction patch/post here instead. 

image

makseq (approved):

PR #1132, opened 2021-07-02
Try fixing connection reset by peer uwsgi error

niklub:


label_studio/projects/api.py
```diff
@@ -613,6 +613,7 @@ class TemplateListAPI(generics.ListAPIView):
     swagger_schema = None
 
     def list(self, request, *args, **kwargs):
+        data = request.data
```
piraka9011, line 616: nit: since data isn't used anywhere

```suggestion
        _ = request.data
```

piraka9011: This is just some pep styling I've seen else where, idk what the conventions are for this project, so do what you will.

piraka9011 (approved):

PR #1182, opened 2021-07-14
Remove RequestDebugLogMixin from APIs

makseq:


chiganov (approved):
We need fixing tests but it's ok for hotfix

PR #1184, opened 2021-07-14
Validate annotators emails when import previous exports

niklub:
This PR solves the "Unknown completed_by ID" validation error when we try to import previously exported JSON tasks, by match existing users given emails:
- if completed_by is omitted - it will assign annotations to the user who imports or user who created this project
- if completed_by is dict object - it will look for `"email"` field inside and raise validation error instead
- if completed_by is integer (deprecated format) - it will try to match existing annotator's ID and raise validation error if not found

### DIFF
--- a/label_studio/tasks/serializers.py
+++ b/label_studio/tasks/serializers.py
@@ -229,6 +229,40 @@ class TaskSerializerBulk(serializers.ListSerializer):
 
         return default
 
+    @staticmethod
+    def get_completed_by_email(annotation, default=None):
+        completed_by = annotation.get('completed_by', None)
+        # user id as is
+        if completed_by and isinstance(completed_by, int):
+            return
+        # user dict
+        if isinstance(completed_by, dict):
+            return completed_by.get('email')
+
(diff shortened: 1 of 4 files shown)


label_studio/tasks/serializers.py
```diff
@@ -217,17 +217,29 @@ def to_internal_value(self, data):
...
-        if completed_by and isinstance(completed_by, dict):
-            return completed_by.get('id')
-
-        return default
+    def _insert_valid_completed_by_id_or_raise(self, annotations, members_email_to_id, members_ids, current_user):
+        for annotation in annotations:
+            completed_by = annotation.get('completed_by')
+            # no completed_by info found - just skip it, will be assigned to the user who imports
+            if completed_by is None:
+                annotation['completed_by_id'] = current_user.id if current_user else None
```
makseq, line 225: if current_user else None --- maybe project.created_by? I'm not sure we process completed_by = None in all places correctly..

makseq (approved):

PR #1190, opened 2021-07-16
Add sentry transactions cleanup

makseq:


label_studio/core/utils/sentry.py
```diff
@@ -1,13 +1,36 @@
 from django.conf import *
 
 
+def event_processor(event, hint):
+    # skip specified exceptions
+    if event.get('exception', {}).get('values', [{}])[-1].get('type') in [
+        'Http404', 'NotAuthenticated', 'AuthenticationFailed', 'NotFound', 'XMLSyntaxError',
```
niklub, line 7: Are we sure that any error of these types could be really ignored?

label_studio/projects/models.py
```diff
@@ -722,7 +722,9 @@ def _get_annotation_key(self, result):
             return None
         if 'from_name' not in result or 'to_name' not in result:
             logger.error(
-                'Unexpected annotation.result format: "from_name" or "to_name" not found in %r' % result)
+                'Unexpected annotation.result format: "from_name" or "to_name" not found in %r' % result,
```
niklub, line 725: to squash sentry errors, it's better to use
```
logger.error('Unexpected annotation.result format: "from_name" or "to_name" not found in %r', result, ...)
```

niklub (approved):

PR #1225, opened 2021-07-27
[fix] Export with resolved uri to coco, yolo, voc

makseq:


niklub:
How are we gonna solve the problem with presigned URL lifetime here?

makseq:
> How are we gonna solve the problem with presigned URL lifetime here?

task.data['$resolved$'] is only for COCO/YOLO/and other export formats for image (and other media) downloading. After export this key $resolved$ will be removed.

PR #1226, opened 2021-07-27
[fix] API tasks with project

makseq:


label_studio/tasks/api.py
```diff
@@ -35,12 +36,24 @@
...
+    def get_queryset(self):
+        pk = int_from_request(self.request.GET, 'project', 0) or int_from_request(self.request.data, 'project', 0)
+        if not pk:
+            raise ValueError('Project is not set')
+
+        project = get_object_with_check_and_log(self.request, Project, pk=pk)
+        if not project.has_permission(self.request.user):
+            raise DRFPermissionDenied('No project access')
+
+        return Task.objects.filter(project__organization=self.request.user.active_organization)
```
triklozoid, line 55: I think we need only this line there.

triklozoid (approved):

PR #1237, opened 2021-08-04
[fix] Fix postgre version in docker-compose for POSTGRES_HOST_AUTH_METHOD

makseq:


docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
     command: [ "./deploy/wait-for-postgres.sh", "bash", "/label-studio/deploy/start_label_studio.sh" ]
 
   db:
-    image: postgres:11.5
+    image: postgres:10.16
```
niklub, line 40: why?

makseq: Previously I had a problems with pg_hba.conf patch, but it works now with 11.5.

PR #1251, opened 2021-08-05
[fix] Fixing is_labeled for task if enforced maximum reached

KonstantinKorotaev:


label_studio/tasks/models.py
```diff
@@ -191,8 +191,12 @@ def update_is_labeled(self):
         """Set is_labeled field according to annotations*.count > overlap
         """
         n = self.annotations.filter(Q_finished_annotations & Q(ground_truth=False)).count()
-        # self.is_labeled = n >= self.project.maximum_annotations
-        self.is_labeled = n >= self.overlap
+        finished_tasks = Task.objects.filter(project=self.project, is_labeled=True).count()
```
niklub, line 194:
```
finished_tasks = self.project.get_labeled_count
```

label_studio/tasks/models.py
```diff
@@ -191,8 +191,12 @@ def update_is_labeled(self):
...
+        if finished_tasks < (self.project.tasks.count() * self.project.overlap_cohort_percentage / 100):
```
niklub, line 195: I don't believe this scheme will work. Image we have 2 tasks:
task1.overlap = 1
task2.overlap = 2
and `overlap_cohort_percentage=50%`

Given sequential mode enabled, I take task1 and fall into this branch. task1.is_labeled becomes True.
Then I get task2, and it ends up with 2nd `else` condition => overlap sets to 1  => task2.is_labeled becomes also True, despite it requires 2 annotations
Am I missing something?

niklub (changes requested):

label_studio/tasks/models.py
KonstantinKorotaev: Cancelling PR as overlap update logic is in label_studio.project.models.update_tasks_states

PR #1302, opened 2021-08-17
Reduce sentry errors

makseq:


label_studio/core/label_config.py
```diff
@@ -63,7 +65,10 @@ def _get_parent_output_tag_name(tag, outputs):
             if name in outputs:
                 return name
 
-    xml_tree = etree.fromstring(config_string)
+    try:
+        xml_tree = etree.fromstring(config_string)
+    except etree.XMLSyntaxError as e:
+        raise LabelStudioXMLSyntaxErrorSentryIgnored(str(e))
```
niklub, line 71: don't we lose a traceback here?

niklub (approved):

label_studio/core/label_config.py
makseq: working as usual
image
image

PR #1307, opened 2021-08-17
[ext] Add deprecation warning about meta field removing from task model

makseq:
…model


niklub (approved):

label_studio/tasks/serializers.py
```diff
@@ -283,6 +283,10 @@ def create(self, validated_data):
                          file_upload_id=task.get('file_upload_id'))
                 db_tasks.append(t)
 
+            # deprecated meta warning
+            if 'meta' in task:
+                logger.warning('You task data has field "meta" which is deprecated and it will be removed in future')
```
smoreface, line 288:
```suggestion
                logger.warning('Your task data contains the "meta" field, which is deprecated and will be removed in the future')
```

PR #1315, opened 2021-08-18
[ext] Allow exporting specific task IDs

niklub:
- use `curl -X GET http://localhost:8080/api/projects/123/export?ids[]=38319\&ids[]=38321` to export only specific tasks
- set JSON exportType by default

### DIFF
--- a/label_studio/data_export/api.py
+++ b/label_studio/data_export/api.py
@@ -28,16 +28,16 @@ logger = logging.getLogger(__name__)
     operation_summary='Get export formats',
     operation_description='Retrieve the available export formats for the current project.',
     responses={200: openapi.Response(
-                description='Export formats',
-                schema=openapi.Schema(
-                    title='Format list',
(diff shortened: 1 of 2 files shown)


label_studio/data_export/api.py
```diff
@@ -56,15 +56,27 @@ def get(self, request, *args, **kwargs):
...
+        openapi.Parameter(name='ids',
+                          type=openapi.TYPE_ARRAY,
+                          items=openapi.Schema(
+                              title='Task ID',
+                              description='Individual task ID',
+                              type=openapi.TYPE_INTEGER
+                          ),
+                          in_=openapi.IN_QUERY,
+                          description="""
+                          To retrieve only subset of tasks, specify the list of target tasks IDs
```
niklub, line 76: @smoreface take a look into API descriptions in this PR

triklozoid (approved):

PR #1344, opened 2021-08-22
Rename label-studio apps from 'app_name' to `label_studio.app_name'

bendidi:
Original slack thread: https://label-studio.slack.com/archives/C01SKFX54QK/p1629105692025200

To increase the reusability of label-studio apps, it seems like a good idea to rename all apps from classic `app_name` format to `label_studio.app_name`, this way they can be reused downstream easily, for example if I want to create my custom LS backend:

```sh
pip install label-studio
```

then in my custom `settings.py`

```
INSTALLED_APPS = [
    .....

    'label_studio.core',
    'my_custom.users',
    'my_custom.organizations',
    'my_custom.data_import',

    'label_studio.projects',
    'label_studio.tasks',
    'label_studio.data_manager',
    'label_studio.io_storages',
    'my_custom.ml',
    'label_studio.webhooks',
]
```
A point where I'm confused is: 

```py
AUTH_USER_MODEL = 'users.User'
```
I'm not sure how to handle the `AUTH_USER_MODEL` settings, since django doesn't recommend to change it mid-project and even if I change it to `label_studio.user.User`, django isn't happy with having this format (having 2 `.` in the string) and mainly something like `<app_name_with_no_point>.<model_name>` 

All tests seem to be passing, apart from 2 `webhook` tests

```sh
======================================================================= short test summary info =======================================================================
FAILED tests/webhooks/test_webhooks.py::test_webhooks_for_projects - assert 2 == 1
FAILED tests/webhooks/test_webhooks.py::test_webhooks_for_tasks - assert 1 == 0
```

I've created an issue [here](https://github.com/heartexlabs/label-studio/issues/1342) since it seems to also be failing on master (before my changes)


label_studio/data_manager/functions.py
```diff
@@ -6,13 +6,13 @@
...
-from data_manager.models import View
-from tasks.models import Task
+from label_studio.core.utils.common import int_from_request
+from label_studio.data_manager.prepare_params import PrepareParams
+from label_studio.data_manager.models import View
+from label_studio.tasks.models import Task
 
 
-TASKS = 'tasks:'
+TASKS = 'label_studio.tasks:'
```
makseq, line 15: This should be unchanged, it's not relative to import

label_studio/io_storages/filesystem.py
```diff
@@ -264,8 +264,8 @@ def remove_all(self, ids=None):
                 self._remove_id_from_keys_map(id)
             self._save_ids()
 
-            logger.debug('Remove all data from tasks.json')
-            # remove record from tasks.json
+            logger.debug('Remove all data from label_studio.tasks.json')
```
makseq, line 267: Wrong replacement

label_studio/manage.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 if __name__ == '__main__':
-    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'core.settings.label_studio')
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'label_studio.core.settings.label_studio')
```
makseq, line 8: Not sure about this replacement

makseq (changes requested):

label_studio/data_manager/functions.py
bendidi: hum, I've also renamed the `app_name` in `label_studio.tasks.urls` to `label_studio.tasks` ([here](https://github.com/heartexlabs/label-studio/pull/1344/files#diff-892c91f02e5dd7ba251171e84c88f4342900422c2fa42e27b76a1cc3caabae4dR8)) 

so if I'm not mistaken the goal is to use it to reverse find urls ? based on app_name ? or should I also revert the `app_name` in `label_sudio.tasks.urls` to just `tasks` ?

label_studio/io_storages/filesystem.py
bendidi: fixed

bendidi:
This can be closed, since the change is too large and may have unintended side effects that are hard to fully test

PR #1355, opened 2021-08-25
Draft: [ext] 🛠 Fixed docker-compose.yml

k4black:
### Problem:
  After version update with docker distributed label-studio, there was no actual update 😅.  
  Because the logic files had been fixed in this volume and were not updated.

### Solution:
  Removing volume
  `- static:/label-studio/label_studio:rw`


farioas:
Fixed in https://github.com/heartexlabs/label-studio/pull/1363

PR #1395, opened 2021-09-01
Specify DB name in initialization script

piraka9011:
If the DB name is not `root`, then we cannot to the DB.
Add the bash substitution arg to `POSTGRE_NAME` so we can set the name of the DB using env. vars

### DIFF
--- a/deploy/docker-entrypoint.d/10-wait-for-postgres.sh
+++ b/deploy/docker-entrypoint.d/10-wait-for-postgres.sh
@@ -3,7 +3,7 @@ set -e
 
 if [ -n "${POSTGRE_HOST:-}" ]; then
   echo >&3 "=> Waiting for postgres..."
-  until PGPASSWORD=$POSTGRE_PASSWORD psql -h "$POSTGRE_HOST" -p $POSTGRE_PORT -U "$POSTGRE_USER" -c '\q'; do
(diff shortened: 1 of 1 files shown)


deploy/docker-entrypoint.d/10-wait-for-postgres.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 if [ -n "${POSTGRE_HOST:-}" ]; then
   echo >&3 "=> Waiting for postgres..."
-  until PGPASSWORD=$POSTGRE_PASSWORD psql -h "$POSTGRE_HOST" -p $POSTGRE_PORT -U "$POSTGRE_USER" -c '\q'; do
+  until PGPASSWORD=$POSTGRE_PASSWORD psql -h "$POSTGRE_HOST" -p $POSTGRE_PORT -U "$POSTGRE_USER" -d "${POSTGRE_NAME:=root}" -c '\q'; do
```
farioas, line 6: POSTGRE_NAME:=postgres
because it should be in sync with https://github.com/heartexlabs/label-studio/blob/master/docker-compose.yml#L33

farioas (changes requested):

farioas (approved):

PR #1398, opened 2021-09-03
[docs] Add copy button, new code highlight colors

makseq:


smoreface:
Can we also have the copy icon for json? And will it show up for manual codeblock formatting too? I'll test this tomorrow morning!

smoreface (commented):
Works everywhere I tested except for the "includes" option --> so it doesn't appear on the example here:
https://labelstud.io/guide/predictions.html#Units-of-image-annotations

makseq:
> Works everywhere I tested except for the "includes" option --> so it doesn't appear on the example here:
> https://labelstud.io/guide/predictions.html#Units-of-image-annotations

@smoreface Could you check my latest commit please?

hlomzik (approved):

PR #1406, opened 2021-09-04
Fixing duplicate key simple in ApiProvider

mpomarole:
This is a minor fix, but the file modal sets the key `simple` twice.


nicholasrq (approved):

nicholasrq:
thanks!

PR #1426, opened 2021-09-08
[fix] Change UploadFile names "project_id-hash-orig_fname"

makseq:
…iginal_name>

### DIFF
--- a/label_studio/data_import/api.py
+++ b/label_studio/data_import/api.py
@@ -391,6 +391,10 @@ class UploadedFileResponse(generics.RetrieveAPIView):
         file = settings.UPLOAD_DIR + ('/' if not settings.UPLOAD_DIR.endswith('/') else '') + filename
         logger.debug(f'Fetch uploaded file by user {request.user} => {file}')
         file_upload = FileUpload.objects.filter(file=file).last()
+
+        if not file_upload.project.has_permission(request.user):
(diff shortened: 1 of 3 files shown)


label_studio/data_import/models.py
```diff
@@ -18,10 +19,16 @@
 logger = logging.getLogger(__name__)
 
 
+def upload_name_generator(instance, filename):
+    return os.path.join(
+        settings.UPLOAD_DIR, str(instance.project_id) + '-' + str(uuid.uuid4())[0:8] + '-' + filename
```
niklub, line 24: can we do here `{str(instance.project)}/{str(uuid.uuid4())[0:8] + '-' + filename}` ?

label_studio/data_import/api.py
```diff
@@ -391,6 +391,10 @@ def get(self, *args, **kwargs):
         file = settings.UPLOAD_DIR + ('/' if not settings.UPLOAD_DIR.endswith('/') else '') + filename
         logger.debug(f'Fetch uploaded file by user {request.user} => {file}')
         file_upload = FileUpload.objects.filter(file=file).last()
+
+        if not file_upload.project.has_permission(request.user):
```
niklub, line 395: better to create has_permission function in `FileUpload`

niklub (approved):

PR #1438, opened 2021-09-13
[fix] Annotation result contains filter optimized

triklozoid:


label_studio/tasks/migrations/0009_auto_20210913_0739.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tasks', '0008_auto_20210903_1332'),
+    ]
+
+    operations = [
+        TrigramExtension(),
+        migrations.RunSQL(
```
makseq, line 15: Could you create the similar migration for task.data please?

makseq (changes requested):

label_studio/tasks/migrations/0009_auto_20210913_0739.py
triklozoid: Discussed and decided to make more investigation before doing that.

makseq (approved):

PR #1485, opened 2021-09-20
Update ML and Export API docs

smoreface:
- Update export API descriptions for delayed export options
- Add docs for listing ML backends
- Update descrs for train ML backend and requesting interactive annotations


label_studio/data_export/api.py
```diff
@@ -223,9 +228,9 @@ def get(self, request, *args, **kwargs):
     name='get',
     decorator=swagger_auto_schema(
         tags=['Export'],
-        operation_summary='List your exports',
+        operation_summary='Get all export files',
```
niklub, line 231: calling exports as `"files"` in LIST/GET/POST/DELETE for exports is somewhat inconsistent with what we get from http://localhost:8080/docs/api#operation/api_projects_exports_download_read

The former is about creating the "snapshot". The latter is about converting the "snapshot" to the actual downloadable file. It's ok to name shapshots as "files", but in this case, we have to rewrite the title/desc for `api_projects_exports_download_read`

niklub (changes requested):

label_studio/data_export/api.py
smoreface: Good point, this is about getting a list of files/info about files that have been exported, and the other GET request is about actually retrieving the files themselves. I'll clarify.

niklub (approved):

PR #1499, opened 2021-09-21
[docs] Clarify role permissions and S3 endpoint

smoreface:
- Clarify what an S3 endpoint is/whether it's needed
- Update role permissions info for regular S3
- Add more details about access key ID etc.

### DIFF
--- a/docs/source/guide/storage.md
+++ b/docs/source/guide/storage.md
@@ -28,9 +28,11 @@ Connect your [Amazon S3]((https://aws.amazon.com/s3)) bucket to Label Studio to 
 For details about how Label Studio secures access to cloud storage, see [Secure access to cloud storage](security.html#Secure-access-to-cloud-storage).
 
 ### Prerequisites
-- You must have programmatic access enabled. [See the Amazon Boto3 configuration documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration) for more on how to set up access to your S3 bucket.
-- You must get temporary security credentials to use to grant access to your S3 bucket. See the AWS Identity and Access Management documentation on [Requesting temporary security credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html). 
-- (Optional) If you want to use a designated account for retrieving source tasks and storing annotations in S3, assign the following role policy to an account you set up for this purpose, replacing `<your_bucket_name>` with your bucket name:
+These prerequisites assume that you're using the same S3 bucket for both source and target storage with Label Studio. If you only use S3 for source storage, Label Studio does not need PUT access to the bucket. 
+
+1. Enable programmatic access to your bucket. [See the Amazon Boto3 configuration documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration) for more on how to set up access to your S3 bucket. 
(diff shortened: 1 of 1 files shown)


docs/source/guide/storage.md
```diff
@@ -28,8 +28,29 @@ Connect your [Amazon S3]((https://aws.amazon.com/s3)) bucket to Label Studio to
 For details about how Label Studio secures access to cloud storage, see [Secure access to cloud storage](security.html#Secure-access-to-cloud-storage).
 
 ### Prerequisites
-
-You must have programmatic access enabled. [See the Amazon Boto3 configuration documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration) for more on how to set up access to your S3 bucket.
+- You must have programmatic access enabled. [See the Amazon Boto3 configuration documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration) for more on how to set up access to your S3 bucket.
+- You must get temporary security credentials to use to grant access to your S3 bucket. See the AWS Identity and Access Management documentation on [Requesting temporary security credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html). 
+- (Optional) If you want to use a designated account for retrieving source tasks and storing annotations in S3, assign the following role policy to an account you set up for this purpose, replacing `<your_bucket_name>` with your bucket name:
```
niklub, line 33: 1. It's not an optional step.
2. Let's create a separate section on bucket permissions setup:
3. We need to clarify that for source storage, the IAM policy should be without `"s3:PutObject"` action since we don't require to store objects in bucket, and the depicted one is for the target storage
4. As the last step in permissions section, we have to provide the link to the CORS setup

niklub: also I don't think we have to duplicate permissions in `Set up an S3 connection with IAM role access` section - they evenly work for both types of AWS S3 storage

niklub (changes requested):

docs/source/guide/storage.md
smoreface: @niklub 
1. fixed.
2. Why a separate step separate from the prerequisites? If it isn't optional, then everyone will need to do it, and the existing prerequisite is about bucket access already so I feel like it fits well in this spot. 
3. I was expecting that people would use 1 user role to grant Label Studio access to the bucket, rather than 2 separate roles, one for each bucket prefix, especially if it's typical that source and target storage would use the same bucket with different prefixes. If that assumption isn't consistent with how people actually set up RBAC with the buckets, let me know and I'll update the steps. 
4. I'll add a link to CORS setup, that also seems like a prerequisite with programmatic access and separate from user role permissions. 
5. I duplicated the permissions in the IAM role access section so that users following those steps wouldn't need to go back and forth between two separate sets of instructions. 





> also I don't think we have to duplicate permissions in Set up an S3 connection with IAM role access section - they evenly work for both types of AWS S3 storage

I agree, but I wasn't sure if it would be too disruptive to have the readers in that section hop back up to the prerequisites and then back down to their task to finish their task.

niklub:
```
[
    {
        "AllowedHeaders": [
            "*"
        ],
        "AllowedMethods": [
            "GET"
        ],
        "AllowedOrigins": [
            "*"
        ],
        "ExposeHeaders": [
            "x-amz-server-side-encryption",
            "x-amz-request-id",
            "x-amz-id-2"
        ],
        "MaxAgeSeconds": 3000
    }
]
```

niklub (approved):

PR #1503, opened 2021-09-22
[ext] Add org members api pagination

triklozoid:


label_studio/organizations/api.py
```diff
@@ -59,6 +60,10 @@ def post(self, request, *args, **kwargs):
         return super(OrganizationListAPI, self).post(request, *args, **kwargs)
 
 
+class OrganizationMemberPagination(PageNumberPagination):
+    page_size = 20
```
makseq, line 64: Let's take it from args, e.g.: `members?page_size=X`

makseq (approved):

PR #1546, opened 2021-09-30
Update CSRF cookie settings to allow insecure host

dsanders11:
Out of the box with the Docker image I was getting a CSRF error for sign up due to it trying to set the CSRF cookie as a secure cookie even though the hostname is `http://`. This PR updates the settings with more friendly defaults.


label_studio/core/settings/label_studio.py
```diff
@@ -16,7 +16,10 @@
 
 DEBUG_PROPAGATE_EXCEPTIONS = get_bool_env('DEBUG_PROPAGATE_EXCEPTIONS', False)
 
+# Sessions and CSRF
```
makseq, line 19: Looks like it's better to write False on this line, what do you think about it? 
https://github.com/heartexlabs/label-studio/blob/master/label_studio/core/settings/base.py#L342

makseq (changes requested):

label_studio/core/settings/label_studio.py
dsanders11: Are you sure that's a change that should go into `settings/base.py`? Seems like it could affect other environments and settings like production deployment if changed there. That was my logic for making the change in `settings/label_studio.py` instead. But if you don't think that's a problem, I can change it.

makseq: To be honest, we have to merge label_studio.py and base.py :-) it's excess split, besides almost all params can be passed over environment variables. So, let's move it to base.py, because we hadn't properly enabled CSRF before yesterday (and it didn't take in account an insecure host option at all) and now it breaks production deployments until we keep CSRF_COOKIE_HTTPONLY = True.

dsanders11: Ok, I've updated this PR to make the change as you've suggested.

makseq (approved):

PR #1559, opened 2021-10-03
Fix offsets in Text with emoji

### LINKED ISSUE #1008: Offsets incorrectly displaying in annotation
image

image

In the attached pictures you can see the proper text for "Ecuador" as an entity, and the annotation tool showing it incorrectly.
My first guess is that it is related to the emoji being used, but if so, it isn't all emoji:
image
As you can see in the third image, offsets continue correctly in that particular case after being exposed to some emoji.

hlomzik:
Store code points instead of js length

closes #1008


makseq:
was merged in another PR.

makseq:
https://github.com/heartexlabs/label-studio-frontend/pull/308

Issue #1241, opened 2021-08-05
Disappearing brush stoke on semantic segmentation annotations


royschwartz2:
**Describe the bug**
When making a brush stroke, part of the brush stroke disappears while working on the stroke, only to appear again when releasing the mouse click. 

**To Reproduce**
Create a brush stroke of more than just a click. 

**Expected behavior**
The brush stroke (mask) should appear throughout the annotation process. Otherwise, this is quite confusing. 

**Screenshots**

**Environment (please complete the following information):**
 - OS: MacOS
 - Label Studio Version 1.1

**Additional context**

hlomzik:
Hi, @royschwartz2 !
Can you check PR #1609 ? The problem was in drafts saving in the middle of drawing process, so now I changed the way it triggers and stores, so it should not interfere with drawing no more. Will be merged soon after checks.

PR #1609, opened 2021-10-13
[fix] glitches during brush drawing

hlomzik:
- save touches and original RLE to drafts, it's much faster
- don't autosave during drawing


hlomzik:
Closes #1241

PR #1618, opened 2021-10-15
Reorganize Storage API

smoreface:
- Update storage API tags to use suffixes for storage type
- Update export and import storage list APIs to include query string parameter for project


smoreface:
Tested and confirmed this works as intended and I didn't break anything :)
Screen Shot 2021-10-18 at 2 34 35 PM
Screen Shot 2021-10-18 at 2 30 10 PM

PR #1640, opened 2021-10-20
[WIP] Reorganize and list agreement metrics

smoreface:
- List all agreement metrics with descriptions, tags, and labeling types
- Provide examples for common agreement metrics
- Update terminology of matching function --> agreement metrics


smoreface:
Updated with extensive examples for all the agreement types. I referred to the code in the evalme library but it's likely I still got something wrong, please check it!

PR #1650, opened 2021-10-22
Various improvements over Preannotations / ML

niklub:
- Bulk import for predictions
- Return tasks IDs after import
- Model versions with counters
- DM API for tasks


label_studio/data_manager/api.py
```diff
@@ -173,6 +178,11 @@ class TaskListAPI(generics.ListAPIView):
         DELETE=all_permissions.tasks_delete,
     )
 
+    def get_serializer_class(self):
```
makseq, line 181: This serializer changes should be propagated to LSE too

makseq (approved):

PR #1688, opened 2021-11-01
[ext] Fast Create annotations from predictions. Add task bulk create callback to import serializer

makseq:


label_studio/core/utils/common.py
```diff
@@ -609,3 +609,7 @@ def round_floats(o):
     if isinstance(o, (list, tuple)):
         return [round_floats(x) for x in o]
     return o
+
+
+def dummy_func(*args, **kwargs):
```
triklozoid, line 614: We don't need now, right?

label_studio/tasks/serializers.py
```diff
@@ -1,6 +1,8 @@
 """This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
 """
 import logging
+from abc import ABC
```
triklozoid, line 4: I can't find where it used.

triklozoid (approved):

PR #1702, opened 2021-11-05
[fix] Set task overlap for storage import

makseq:


label_studio/io_storages/base_models.py
```diff
@@ -107,7 +109,7 @@ def _scan_and_create_links(self, link_class):
                 data = data['data']
 
             with transaction.atomic():
-                task = Task.objects.create(data=data, project=self.project)
+                task = Task.objects.create(data=data, project=self.project, overlap=maximum_annotations)
```
triklozoid, line 112: Just a minor question: Why we can not access the project attribute in-place? 
```
Task.objects.create(data=data, project=self.project, overlap=self.project.maximum_annotations)
```

triklozoid (approved):

PR #1704, opened 2021-11-05
DEV-882 fix is_labeled calculation

triklozoid:


label_studio/tasks/models.py
```diff
@@ -200,77 +200,13 @@ def storage(self):
...
-    @staticmethod
-    def bulk_reset_updates(project):
-        """ Bulk reset updates to default, it's a fast way to reset all tasks in project
-        """
-        for field in Task._meta.fields:
-            if field.name in Task.updates:
-                project.tasks.update(**{field.name: field.default})
-
-    @staticmethod
-    def bulk_update_is_labeled(project):
```
makseq, line 227: I'm not sure it's a good idea to remove it. Yes, it's not used anywhere, but it must be used in all bulk imports for tasks and now it's a bug.

makseq (approved):

makseq (approved):

PR #1716, opened 2021-11-09
Update Dockerfile

Muaado:


farioas:
Experimental API is not strict compared to the fixed version. It may change any time, we don't want our users to experience compatibility issues

PR #1721, opened 2021-11-10
Update frontend

niklub:


Etto91:
hi guys, if I want to use the latest version of lsf on label_studio, do I have to wait for this pr to be merged? i've tried to replace the lsf folder but the frontend seems broken. do i need to replace dm2 either? thanks

PR #1722, opened 2021-11-10
[docs] Update with chart repo usecases

farioas:


smoreface (approved):
Fine tuned some language, otherwise looks great!

PR #1733, opened 2021-11-12
[fix] Empty annotations and predictions in DM TaskSerializer

makseq:


label_studio/data_manager/serializers.py
```diff
@@ -230,14 +240,12 @@ def get_annotations_results(self, task):
     def get_predictions_results(self, task):
         return self._pretty_results(task, 'predictions_results')
 
-    def get_annotations(self, task):
-        if not self.context.get('annotations'):
-            return []
+    @staticmethod
+    def get_annotations(task):
```
niklub, line 244: why did they become static methods?

makseq: they don't use any self fields now

PR #1747, opened 2021-11-17
Get only task ids via drf-flexfields (DEV-1102)

triklozoid:


Makefile
```diff
@@ -28,4 +28,4 @@ frontend-build:
 
 # Run tests
 test:
-	cd label_studio && pytest -v -m "not integration_tests"
+	cd label_studio && DJANGO_DB=sqlite pytest -v -m "not integration_tests"
```
makseq, line 31: Why should we do this?

triklozoid: Do you use this Makefile command? 
For me, it doesn't work at all before this change.

label_studio/core/settings/base.py
```diff
@@ -448,3 +448,7 @@ def collect_versions_dummy(**kwargs):
 
 mimetypes.add_type("application/javascript", ".js", True)
 mimetypes.add_type("image/png", ".png", True)
+
+
+# fields name was used in DM api before
```
niklub, line 453: comment is misleading - is it about all drf serializers, not only DM ones ?

niklub (approved):

Issue #1749, opened 2021-11-17
CI and development improvements


twsl:
**Is your feature request related to a problem? Please describe.**
The CI chain and general development environment could use some improvements to make some things easier.
Things I noticed so far:
- [ ] no pre-commit
- [ ] no linting
- [ ] no stale option for issues/prs
- [ ] ci improvements
- [ ] docker deployment on github actions
- [ ] security checks

**Describe the solution you'd like**
The solution should address multiple things like devcontainer and github actions improvements.
I encourage everyone to contribute more ideas.

**Describe alternatives you've considered**
-

**Additional context**

PR #1751, opened 2021-11-18
Improve devcontainer

twsl:
As part of #1749 this PR improves the devcontainer and introduces a pre-commit hook to enforce style


farioas (changes requested):
Please split this PR to two. Move all the changes not related to devcontainers to another PR and name it accordingly

PR #1753, opened 2021-11-18
Enforce PEP8 and black

twsl:
As part of #1749 this PR enforces the style guidelines of PEP8 and black


label_studio/__init__.py
```diff
@@ -1,10 +1,9 @@
-"""This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
-"""
+"""This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license."""  # noqa: E501
 # Package name
-package_name = 'label-studio'
+package_name = "label-studio"
```
makseq, line 3: Sorry, but we use ' instead of " deliberately. There is a special option in black to skip this check.

makseq (changes requested):

label_studio/__init__.py
twsl: The black docs say `This is meant as an adoption helper, avoid using this for new projects.`.
However, if you already made that decision to keep the single quotes, I can reverse that.
That's why it's a draft 😄

makseq: Yeh, LS is not a new project ;)

makseq: And if we talk about future: PRs with more than 100 lines are very difficult to review. Please, note this! :slightly_smiling_face:

PR #1784, opened 2021-11-24
feat: DEV-1205: Add task.updated_at column

triklozoid:


label_studio/tasks/models.py
```diff
@@ -317,6 +317,12 @@ def decrease_project_summary_counters(self):
             summary = self.task.project.summary
             summary.remove_created_annotations_and_labels([self])
 
+    def save(self, *args, **kwargs):
```
makseq, line 320: We have to check does it work on annotations delete action in DM, the most likely it won't work. Also there might be more bulk operations we should care about..

triklozoid: Fixed annotations delete from DM. 
With bulk actions, I found only "create annotations from predictions" where we need this logic. In other places, we use bulk create to create tasks with annotations, so we don't need to set updated_at there.

label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -42,6 +46,7 @@ def predictions_to_annotations(project, queryset, **kwargs):
     logger.debug(f'{count} predictions will be converter to annotations')
     db_annotations = [Annotation(**annotation) for annotation in annotations]
     db_annotations = Annotation.objects.bulk_create(db_annotations)
+    Task.objects.filter(id__in=tasks_ids).update(updated_at=now())
```
makseq, line 49: Maybe it's better to move into `TaskSerializerBulk.post_process_annotations` ?

triklozoid: It will add a database update query, that we don't need on import. If it's ok, I can move.

makseq: Got it.

makseq (dismissed):

label_studio/data_manager/actions/predictions_to_annotations.py
makseq: Why do we need this line if we already have set_updated_at = True for `perform_action`?

You can put task_ids in `return {'response_code': 200, 'detail': f'Created {count} annotations',  ==> 'task_ids': xxx}`

and then use them [here ](https://github.com/heartexlabs/label-studio/pull/1784/files#diff-8ff0ffe1df554b71bae2a1554d8ce537efd879e691ea46211339893d561a1232R113)

don't forget to do `result.pop('task_ids')`

makseq (changes requested):

makseq (approved):

label_studio/data_manager/actions/experimental.py
```diff
@@ -103,6 +103,7 @@ def remove_duplicates(project, queryset, **kwargs):
         'entry_point': remove_duplicates,
         'permission': all_permissions.tasks_change,
         'title': 'Remove Duplicated Tasks',
+        'set_updated_at': False,
```
makseq, line 106: Do we need this?

makseq (changes requested):

makseq (approved):

PR #1786, opened 2021-11-25
DEV-1103: Export API specs are incomplete

chiganov:


triklozoid (approved):

smoreface:
sorry for the delay -- it would be helpful to provide help text for these options, and also defaults if they exist. But our existing API documentation doesn't usually have help text so this would be above and beyond. LGTM besides.

PR #1800, opened 2021-12-01
iRobot changes: Resolve S3 URLs for tasks. Cache boto3 sessions.

wgillett:
Resolving S3 URLs in tasks is key for our workflows. Caching boto3 sessions is an essential performance improvement. Tested manually and through repeated use.

### DIFF
--- a/label_studio/tasks/serializers.py
+++ b/label_studio/tasks/serializers.py
@@ -23,6 +23,8 @@ from core.label_config import replace_task_data_undefined_with_config_field
 from users.serializers import UserSerializer
 from core.utils.common import load_func
 
+from io_storages.s3.utils import get_client_and_resource, resolve_s3_url
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


label_studio/tasks/serializers.py
```diff
@@ -162,6 +164,13 @@ def to_representation(self, instance):
             if self.context.get('resolve_uri', False):
                 instance.data = instance.resolve_uri(instance.data, proxy=self.context.get('proxy', False))
 
+            # resolve S3 URLs
```
triklozoid, line 167: We already have a generic URL resolving function on the previous line. And it is also call `resolve_s3_url`. So in some cases, we will get double URL resolving. You can fix the logic of retrieving URL from `instance.data` there: https://github.com/wgillett/label-studio/blob/irobot-2021-11-30/label_studio/io_storages/utils.py#L12

niklub:
Hi, @wgillett ! We've checked the proposed changes, and there are few questions:
1. We already have a [naive S3 client caching mechanism implemented](https://github.com/wgillett/label-studio/blob/irobot-2021-11-30/label_studio/io_storages/s3/models.py#L60), I'm not quite sure that LRU will bring an additional performance gain on top of it. Your implementation looks definitely more streamlined than ours, but I'm just wondering whether we're missing something and what might cause the improvements on your side you've mentioned earlier?
2. About resolving `s3://` type links: we have [this functionality currently available](https://github.com/heartexlabs/label-studio/blob/master/label_studio/tasks/models.py#L174) - do you have some specific URL that aren't resolved by using this method? Do you mind sharing the example?

wgillett:
We discussed in a call today. For S3:// links, our current code does not trigger S3 URL resolution, but I learned about USE_DEFAULT_S3_STORAGE and will try that. Re client caching, I thought that wasn't enough but will take a fresh look. Need to put this on hold for a bit, back on it next week. Thanks for your help.

makseq:
I'm closing this as outdated and solved (as it follows here: "We already have a [naive S3 client caching mechanism implemented](https://github.com/wgillett/label-studio/blob/irobot-2021-11-30/label_studio/io_storages/s3/models.py#L60)"). 
Feel free to re-open, if it still requires an attention.

wgillett:
This ticket is obsolete for sure, thanks Max Tkachenko

Walter
________________________________

PR #1801, opened 2021-12-01
[ext] Repeater support on backend (DEV-1073)

makseq:


nicholasrq (dismissed):

label_studio/core/label_config.py
```diff
@@ -356,6 +356,22 @@ def generate_sample_task_without_check(label_config, mode='upload', secure_mode=
             # not found by name, try get example by type
             task[value] = examples.get(p.tag, 'Something')
 
+        # support for Repeater tag
```
niklub, line 359: in future - let's make similar workarounds as a feature flag, nothing prevents here to break config when adding `value=$something[0]`

PR #1820, opened 2021-12-07
DEV-405: Update import details and troubleshooting for HTML minification

smoreface:


docs/source/guide/FAQ.md
```diff
@@ -93,5 +93,18 @@ If you find that after annotating audio data, the visible audio wave doesn't mat
 ffmpeg -y -i audio.mp3 -ar 8k -ac 1 audio.wav
 ```
 
+## HTML label offsets are in the wrong places
```
makseq, line 96: Unfortunately, this is not the only reason for incorrect offsets.. It might be caused by complex CSS and some tricky html tags.. @hlomzik correct me please if I'm wrong OR add these cases here

makseq (changes requested):

docs/source/guide/FAQ.md
makseq: Or maybe it's better to disable this minification completely.

smoreface: pinging @niklub to comment on the potential of disabling minification

smoreface: I added a few sentences to acknowledge that this isn't the only cause, but the most common one

PR #1865, opened 2021-12-22
DEV-584: Fixes for view all

Gondragos:
- rework view all rendering
- fix displaying image regions in view all (isReady logic)
- fix displaying richtext regions in view all


hlomzik:
merged in https://github.com/heartexlabs/label-studio/pull/1917

PR #1910, opened 2022-01-11
fix: DEV-1476: Resolving performance for project storages

makseq:


label_studio/projects/models.py
```diff
@@ -683,6 +684,16 @@ def get_model_versions(self, with_counters=False):
         else:
             return list(output)
 
+    @lru_cache(maxsize=None)
```
niklub, line 687: I think using LRU cache here is wrong - it'll keep returning the same storage object per project, even if this storage has been removed

niklub (changes requested):

niklub (approved):

PR #1963, opened 2022-01-26
fix: DEV-1220: DEV-1555: update LSF

hlomzik:
just update LSF after these PRs merged


hlomzik:
outdated. these PRs already in master

PR #1965, opened 2022-01-26
fix: DEV-1525: update label-studio-converter with conll, pillow & nltk version bumps. Remove python 3.6 from tests

makseq:
…k version bumps


makseq:
https://github.com/heartexlabs/label-studio/issues/1959

PR #1978, opened 2022-01-28
Fix ML API timeouts

makseq:


label_studio/ml/api_connector.py
```diff
@@ -163,7 +180,7 @@ def train(self, project, use_ground_truth=False):
                 'password': project.task_data_password
             }
         }
-        return self._request('train', request, verbose=False)
+        return self._request('train', request, verbose=False, timeout=30.0)
```
farioas, line 183: timeout=TIMEOUT_TRAIN ?

PR #1988, opened 2022-02-01
fix: DEV-1564: DEV-1565: Fix shortcuts focus and cursor problems

Gondragos:


hlomzik:
https://github.com/HumanSignal/label-studio-frontend/pull/461 is closed

PR #2014, opened 2022-02-10
fix: DEV-1694: Make update_tasks_states async

KonstantinKorotaev:


label_studio/data_manager/actions/basic.py
```diff
@@ -54,8 +53,13 @@ def delete_tasks(project, queryset, **kwargs):
     # delete only specific tasks
     else:
         # this signal re-save the task back
-        with temporary_disconnect_signal(signals.post_delete, update_is_labeled_after_removing_annotation, Annotation):
+        with temporary_disconnect_all_signals():
             queryset.delete()
+        project.update_tasks_states(
```
niklub, line 58: that will not recompute project.summary counters

label_studio/data_manager/actions/basic.py
```diff
@@ -54,8 +53,13 @@ def delete_tasks(project, queryset, **kwargs):
     # delete only specific tasks
     else:
```
niklub, line 54: we don't need to keep if..else branch if behaviour is  identical

niklub (changes requested):

label_studio/data_manager/actions/basic.py
```diff
@@ -54,8 +53,13 @@ def delete_tasks(project, queryset, **kwargs):
...
         # this signal re-save the task back
-        with temporary_disconnect_signal(signals.post_delete, update_is_labeled_after_removing_annotation, Annotation):
+        with temporary_disconnect_all_signals():
```
makseq, line 56: I'm not sure we can disable all signals here, because we storages use signals to remove annotations

makseq (approved):

label_studio/data_manager/actions/basic.py
```diff
@@ -43,20 +43,22 @@ def delete_tasks(project, queryset, **kwargs):
...
-            overlap_cohort_percentage_changed=False,
-            tasks_number_changed=True
-        )
 
     # delete only specific tasks
     else:
-        # this signal re-save the task back
-        with temporary_disconnect_signal(signals.post_delete, update_is_labeled_after_removing_annotation, Annotation):
+        # update project summary
+        project.summary.remove_created_annotations_and_labels(Annotation.objects.filter(task__in=queryset))
```
niklub, line 51: could we move all summary recalculation into (a)sync job?

niklub (changes requested):

label_studio/data_manager/actions/basic.py
KonstantinKorotaev: Fixed

label_studio/data_manager/actions/basic.py
KonstantinKorotaev: Fixed

niklub (changes requested):

label_studio/core/utils/common.py
```diff
@@ -605,3 +605,39 @@ def round_floats(o):
     if isinstance(o, (list, tuple)):
         return [round_floats(x) for x in o]
     return o
+
+
+class temporary_disconnect_list_signal:
+    """ Temporarily disconnect a list of signals
+        Each signal tuple: (signal_type, signal_method, object)
+        Example:
+            with temporary_disconnect_all_signals(
```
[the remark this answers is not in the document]
KonstantinKorotaev: Fixed

PR #2052, opened 2022-02-21
fix: DEV-1725: Add ANNOTATIONS_CREATED webhook action to predictions to annotations action

KonstantinKorotaev:
…to annotations action

### DIFF
--- a/label_studio/webhooks/utils.py
+++ b/label_studio/webhooks/utils.py
@@ -4,11 +4,9 @@ from functools import wraps
 import requests
 from core.utils.common import load_func
 from django.conf import settings
-from django.db import models
 from django.db.models import Q
 
 from .models import Webhook, WebhookAction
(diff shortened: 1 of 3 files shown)


label_studio/webhooks/utils.py
```diff
@@ -176,9 +174,18 @@ def wrap(self, request, *args, **kwargs):
...
+    :param field: Field to lookup
+    :return: List or single instance of looked up field
+    """
     if field == '__self__':
         return value
     fields = field.split('__')
     for fld in fields:
-        value = getattr(value, fld)
-    return value
+        if isinstance(value, list):
```
makseq, line 187: this can be calculated outside of the loop

makseq (changes requested):

makseq (approved):

PR #2097, opened 2022-03-08
fix: SRE-6: Limit page_size with 100 tasks

triklozoid:


label_studio/core/utils/common.py
```diff
@@ -140,6 +143,9 @@ def paginator(objects, request, default_page=1, default_size=50):
     :return: paginated objects
     """
     page_size = request.GET.get('page_size', request.GET.get('length', default_size))
+    if settings.VERSION_EDITION != 'Community' and (int(page_size) > 100 or page_size == '-1'):
```
niklub, line 146: I guess we need to change help message In `paginator_help`

niklub (changes requested):

label_studio/core/utils/common.py
triklozoid: Fixed.

makseq (approved):

PR #2117, opened 2022-03-11
fix: DEV-1324: DEV-1233: fix undo/redo for Audio

hlomzik:
- undo doesn't work for audio
  - if there was a draft, moved region will disappear after undo
  - if there was no drafts, nothing happens
- redo do nothing for audio (connected with previous)


Gondragos (approved):

hlomzik:
closed in #2198

PR #2127, opened 2022-03-14
fix: DEV-510: fix Azure Single Click template

nikitabelonogov:


azuredeploy.json
```diff
@@ -85,21 +96,39 @@
...
+              "name": "LABEL_STUDIO_ONE_CLICK_DEPLOY",
+              "value": "1"
+            }
           ],
-          "properties": {
-            "pythonVersion": "3.7"
-          }
-        }
-      ]
+          "linuxFxVersion": "DOCKER|heartexlabs/label-studio:latest",
```
farioas, line 127: Let's make image tag configurable field like username and password.
Default should be set to `latest`

farioas (changes requested):

azuredeploy.json
nikitabelonogov: done

PR #2129, opened 2022-03-14
fix: DEV-1906: launchdarkly to switch to file if cannot connect to redis

qs:
- Fix launchdarkly to switch to file if cannot connect to redis
 - Add request serializer for debugging

### DIFF
--- a/label_studio/core/feature_flags/base.py
+++ b/label_studio/core/feature_flags/base.py
@@ -38,7 +38,7 @@ elif settings.FEATURE_FLAGS_OFFLINE:
     client = ldclient.get()
 else:
     # Production usage
-    if hasattr(settings, 'REDIS_LOCATION'):
+    try:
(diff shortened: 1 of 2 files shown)


label_studio/core/feature_flags/base.py
```diff
@@ -49,7 +49,7 @@
             feature_store=store,
             http=HTTPConfig(connect_timeout=5)
         ))
-    else:
+    except:
```
makseq, line 52: please don't use so broad exceptions, let's do 

```
except Exception as exc: 
 logger.debug('...: {exc}')
```

label_studio/core/utils/request.py
```diff
@@ -0,0 +1,7 @@
+def serialize_request(request):
+    try:
+        data = {key: value for key, value in request.META.items() if isinstance(value, (str, bool, int, float))}
+        data[request.method] = getattr(request, request.method)
+        return data
+    except:
```
makseq, line 6:
```
except Exception as exc: 
 logger.debug('...: {exc}')
```

PR #2133, opened 2022-03-15
fix: dev-1631: handle-open-polygon

Travis1282:
handle toolswitch on open polygon


hlomzik:
merged already by https://github.com/heartexlabs/label-studio/commit/867ffaadb036682c81558b41ccc3413ddcfa60d2

PR #2153, opened 2022-03-18
fix: dev-1908: Modified code to include async delete

HeartexDL:


label_studio/projects/api.py
```diff
@@ -193,7 +193,9 @@ def get(self, request, *args, **kwargs):
 
     @api_webhook_for_delete(WebhookAction.PROJECT_DELETED)
     def delete(self, request, *args, **kwargs):
-        return super(ProjectAPI, self).delete(request, *args, **kwargs)
+        with ThreadPoolExecutor(max_workers=1) as executor:
```
makseq, line 196: It's a wrong solution. You will always return 204 despite to real errors in ProjectAPI. You need to take a look here: https://github.com/heartexlabs/label-studio/blob/a757398bd3bf21296540e16b8d02b5107ea7b409/label_studio/projects/api.py#L218 and understand what is going wrong there.

makseq (changes requested):

PR #2154, opened 2022-03-18
fix: dev-239: Modified code to include jsonl format file

HeartexDL:


label_studio/io_storages/s3/models.py
```diff
@@ -157,12 +157,22 @@ def get_data(self, key):
         _, s3 = self.get_client_and_resource()
         bucket = s3.Bucket(self.bucket)
         obj = s3.Object(bucket.name, key).get()['Body'].read().decode('utf-8')
-        value = json.loads(obj)
-        if not isinstance(value, dict):
-            raise ValueError(f"Error on key {key}: For S3 your JSON file must be a dictionary with one task")
 
-        value = self._get_validated_task(value, key)
-        return value
+        if key.endswith(".jsonl"):
```
makseq, line 161: we need this not for s3 only, but for all storages.

makseq (changes requested):

label_studio/io_storages/s3/models.py
HeartexDL: Implemented for all the storage types

PR #2158, opened 2022-03-21
fix: DEV-1920: Review performance and other speed ups

makseq:


triklozoid (approved):

triklozoid:
I think it's better to create indexes only for fields, where we really need them.

PR #2168, opened 2022-03-25
fix: DEV-2024: Add control weight description with example

KonstantinKorotaev:


label_studio/projects/models.py
```diff
@@ -188,7 +188,12 @@ class SkipQueue(models.TextChoices):
         help_text='Minimum number of completed tasks after which model training is started',
     )
 
-    control_weights = JSONField(_('control weights'), null=True, default=dict, help_text='Weights for control tags')
+    control_weights = JSONField(_('control weights'), null=True, default=dict, help_text="Dict of weights for each control tag in metric calculation. Each control tag (e.g. label or choice) will "
+                                                                                         "have it's own key in control weight dict with type, weight for each label and overall weight." 
+                                                                                         'For example for default object detection with bounding boxes control tags dict will look like this: '
```
niklub, line 193: why do we need "type" here?
Also, I would rewrite the example as following:
```
For example, if bounding box annotation with control tag named `"my_bbox"` should be included with 0.33 weight in agreement calculation, and the first label `"Car"` should be twice more important than `"Airplaine"`, then you have to need the specify: {"my_bbox": {"labels": {"Car": 1.0, "Airplaine": 0.5}, "overall": 0.33}
```

niklub (changes requested):

label_studio/projects/models.py
KonstantinKorotaev: Type removed, but added to example.
Example fixed.

niklub (approved):

PR #2183, opened 2022-03-30
fix: dev-2057: width is auto adjustable

Travis1282:
calculate row width based on the biggest visible taxonomy item


hlomzik:
merged a while ago

PR #2185, opened 2022-03-31
fix: DEV-1556: Drafts left untouched after update of other user's annotation

makseq:


label_studio/tasks/models.py
```diff
@@ -610,8 +610,6 @@ def update_is_labeled_after_removing_annotation(sender, instance, **kwargs):
 def delete_draft(sender, instance, **kwargs):
     task = instance.task
     query_args = {'task': instance.task, 'annotation': instance}
-    if instance.completed_by is not None:
-        query_args['user'] = instance.completed_by
```
triklozoid, line 614: Can we add a simple test scenario? 
* Create draft by user A
* Create draft by user B
* Update annotation by user A
* Check that there is no drafts for annotation.

PR #2190, opened 2022-04-01
WIP for DEV-1996

dvwright:
WIP for [DEV-1996](https://heartex.atlassian.net/browse/DEV-1996)

* Fix tavern tests mac osx (+windows)
*

### DIFF
--- a/label_studio/tests/data_manager/api_tasks.tavern.yml
+++ b/label_studio/tests/data_manager/api_tasks.tavern.yml
@@ -102,238 +102,307 @@ stages:
         "total": 4,
         "tasks": [
         {
-          "cancelled_annotations": 0,
-
-          "annotations_results": "",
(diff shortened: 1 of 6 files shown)


label_studio/tests/data_manager/api_tasks.tavern.yml
```diff
@@ -102,238 +102,307 @@ stages:
         "total": 4,
```
dvwright, line 102: XXX: very odd behavior  in my environment. 
if this test is executed standalone
```pytest -k tasks-all-fields-postgre -vv tests/data_manager/api_tasks.tavern.yml```
it passes

if it's executed as part of the test suite it fails! it creates one less prediction
```
    "total_annotations": 4,
    "total_predictions": 2,
    "total": 3,
```

PR #2194, opened 2022-04-04
fix: DEV-1996: EmptyPage contains no result shouldn't return 500 error

dvwright:
WIP for [DEV-1996](https://heartex.atlassian.net/browse/DEV-1996)

* Fix tavern tests mac osx (+windows)
* skip tavern postgres test mac osx (passes standalone, not part of suite)
* handle empty paging


label_studio/core/utils/common.py
```diff
@@ -148,16 +148,22 @@ def paginator(objects, request, default_page=1, default_size=50):
```
dvwright, line 148: why does this method contain a `DEPRECATED` and a TODO to use djp paginator? looks like it does use it. 
can the deprecation text be removed?

triklozoid: Comment about django rest framework pagination. 
https://www.django-rest-framework.org/api-guide/pagination/

We already use it: https://github.com/heartexlabs/label-studio/blob/master/label_studio/projects/api.py#L88
and I think we should use it for all cases.

label_studio/tests/data_manager/api_tasks.tavern.yml
```diff
@@ -4,7 +4,7 @@ strict: false
 marks:
   - usefixtures:
       - django_live_url
-  - skipif: "'default' not in '{tavern.env_vars.DJANGO_DB}'"
+  - skipif: "'default' not in '{tavern.env_vars.DJANGO_DB}' or 'Darwin' in platform.system()"
```
triklozoid, line 7: I'm not sure about it. 
LabelStudio support PostgreSQL and sqlite database backends. Idea of this skip was to check PostgreSQL related code only when we setup project to use it. 

If you don't want to run these tests on `Darwin` system, you should control it over environment variables on tests run. 

How we run tests in our CI: https://github.com/heartexlabs/label-studio/blob/develop/.github/workflows/tests.yml#L238

Place in the code, where we read this env variable: https://github.com/heartexlabs/label-studio/blob/develop/label_studio/core/settings/base.py#L101

triklozoid (approved):

label_studio/tests/data_manager/api_tasks.tavern.yml
dvwright: > If you don't want to run these tests on Darwin system, you should control it over environment variables on tests run.

I would like to run them, however,  this test fails on my system. Apple M1. I spent much time trying to resolve it. A updated test which passes on my system fails as part when run from ci on linux.

I agree it's better to not skip but since Linux is our target production architecture I thought it would not matter so much, there are tests we skip on Windows as well.

label_studio/core/utils/common.py
dvwright: My point was the deprecated comment was confusing since this method does already use Django pagination.

I get it, you feel this code can be removed and replaced with the new code.

ok, for now, I am just making this change. I will leave the TODO in place for a code refactor at another time

makseq (approved):

PR #2231, opened 2022-04-13
feat: DEV-2147: Calculate only requested fields

triklozoid:


label_studio/projects/models.py
```diff
@@ -48,40 +48,65 @@ def for_user(self, user):
...
 
     @staticmethod
-    def with_counts_annotate(queryset):
-        return queryset.annotate(
-            task_number=Count('tasks', distinct=True),
-            finished_task_number=Count('tasks', distinct=True, filter=Q(tasks__is_labeled=True)),
-            total_predictions_number=Count('tasks__predictions', distinct=True),
-            total_annotations_number=Count(
+    def with_counts_annotate(queryset, fields=None):
+        def annotate_task_number(queryset):
```
makseq, line 56: I would move this part to functions.py

makseq (changes requested):

label_studio/core/utils/common.py
```diff
@@ -657,3 +657,10 @@ def __exit__(self, type_, value, traceback):
                 sender=sender,
                 dispatch_uid=dispatch_uid
             )
+
+
+def get_fields_from_request(request):
+    fields = request.query_params.get('include', None)
```
makseq, line 663: maybe `include` => 'fields` ? because we have a similar field name in api/tasks.

label_studio/projects/api.py
```diff
@@ -187,7 +187,8 @@ class ProjectAPI(generics.RetrieveUpdateDestroyAPIView):
     redirect_kwarg = 'pk'
 
     def get_queryset(self):
-        return Project.objects.with_counts().filter(organization=self.request.user.active_organization)
+        fields = get_fields_from_request(self.request)
```
makseq, line 190: Why didn't you get query_param using using serializer?

makseq (changes requested):

label_studio/core/utils/common.py
triklozoid: `fields` was default for drf-flexfields, but we already had this parameter with different meaning (`fields=all`)

label_studio/projects/models.py
triklozoid: Fixed.

label_studio/projects/api.py
triklozoid: Fixed.

makseq (approved):

PR #2259, opened 2022-04-19
performance redis exploratory 

dvwright:
* not sure these changes are needed here
* it doesn't appear that rqworker are used


triklozoid:
not sure these changes are needed here
    it doesn't appear that rqworker are used

We [should] support both cases in open source label studio. 
Users can start it as simple way as possible, without additional services, like redis, but also, users should have the ability to set up connection with redis and improve performance that way.

triklozoid (approved):

PR #2262, opened 2022-04-21
fix: DEV-2184: Next task query optimization

triklozoid:


label_studio/projects/functions/next_task.py
```diff
@@ -13,30 +13,13 @@
...
-            except Exception as exc:
-                logger.error(exc, exc_info=True)
-            else:
-                try:
-                    task = Task.objects.select_for_update(skip_locked=True).get(pk=task.id)
-                    if not task.has_lock(user):
-                        return task
-                except Task.DoesNotExist:
-                    logger.debug('Task with id {} locked'.format(task.id))
+    for task in task_query.order_by('?').only('id')[:settings.RANDOM_NEXT_TASK_SAMPLE_SIZE]:
```
makseq, line 16: are you sure we can skip 'overlap'?

triklozoid: Looks like we call `has_lock` on the task object with all field: 
```
                    task = Task.objects.select_for_update(skip_locked=True).get(pk=task.id)
                    if not task.has_lock(user):
```

triklozoid:
Django version updated because of: https://code.djangoproject.com/ticket/26390

PR #2264, opened 2022-04-21
fix: DEV-2213: Refactor and Optimize Redis Queues

dvwright:
define redis queues


dvwright:
I'm not sure why the `bandit` tests are failing but they seem to be failing on the default branch as well

makseq:
@dvwright failing bandit is ok, we are fixing this in another PR.

makseq (approved):

PR #2273, opened 2022-04-26
fix: DEV-2236: Stored XSS via SVG file

dvwright:
* strip harmful tags from SVG files


label_studio/data_import/api.py
```diff
@@ -459,7 +459,8 @@ def get(self, *args, **kwargs):
         file = file_upload.file
         if file.storage.exists(file.name):
             content_type, encoding = mimetypes.guess_type(str(file.name))
-            content_type = content_type or 'application/octet-stream'
+            if content_type not in ['image/jpeg', 'image/gif', 'image/png', 'application/pdf']:
```
makseq, line 462: Looks like it may break a playback for audio, video and other media formats that browser supports natively - please check the list of formats here: https://developers.google.com/cast/docs/media

makseq (changes requested):

label_studio/data_import/api.py
dvwright: Does label studio allow labeling audio and video content? 

I thought this endpoint `UploadedFileResponse` was only a convenience for viewing uploaded content? 

Does it have a function for the application? I don't think we need to allow this end point at all...

makseq: > Does label studio allow labeling audio and video content? 

Sure, check all the templates on the Create Project dialog. 

> I thought this endpoint UploadedFileResponse was only a convenience for viewing uploaded content?

This endpoint provides uploaded data not only for viewing, but for annotators during labeling too. It's important part especially for our opensource community users. Enterprise clients use s3/gcs/azure storages more, but sometimes they like to switch to uploaded files. 

btw - after this fix users won't be able to use SVG files in their labeling workflows. It seems to fix this issue we need to parse uploaded SVG file and remove its dangerous parts...

dvwright:
needs more testing, does work for the example provided by the user

label_studio/data_import/api.py
```diff
@@ -459,8 +460,52 @@ def get(self, *args, **kwargs):
...
+                'polyline',
+                'rect'
+        ]
+
+        cleaner = clean.Cleaner(
+                allow_tags=allow_tags,
+                remove_unknown_tags=False
+        )
+
+        fd_dirty = open(dirty_file.path, 'r')
```
makseq: Better to write 
```
with open(dirty_file.path) as fd_dirty:
   ...
```

label_studio/data_import/api.py
```diff
@@ -459,8 +460,52 @@ def get(self, *args, **kwargs):
...
+        clean_xml = cleaner.clean_html(dirty_xml)
+        fd_dirty.close()
+
+        fd_clean = tempfile.NamedTemporaryFile(delete=False)
+        try:
+            fd_clean.write(clean_xml.encode())
+            fd_clean.seek(0)
+            return fd_clean
+        except IOError as e:
+            logger.debug(f'Sanitize svg file upload error {e}')
```
makseq: Better to use `logger.error()` or maybe raise an exception.

label_studio/data_import/api.py
```diff
@@ -459,8 +460,52 @@ def get(self, *args, **kwargs):
...
-                content_type = 'text/plain'
+            content_type = content_type or 'application/octet-stream'
+            mime_type_filter = [
+                    'image/svg+xml',
+                    # 'application/xhtml+xml',
+                    # 'text/html',
+                    # 'text/javascript'
+            ]
+            if content_type in mime_type_filter:
+                return RangedFileResponse(
```
makseq, line 471: I would make all checks on the import step and keep this place untouched.

label_studio/data_import/models.py
```diff
@@ -67,6 +67,8 @@ def content(self):
             body = getattr(self, '_file_body')
         else:
             body = self.file.read().decode('utf-8')
+            # TODO: ? filter out malicious/harmful content from file.
```
makseq, line 70: Could we call _allowlist_svg here?

label_studio/data_import/api.py
```diff
@@ -449,7 +450,7 @@ class UploadedFileResponse(generics.RetrieveAPIView):
     def get(self, *args, **kwargs):
         request = self.request
         filename = kwargs['filename']
-        file = settings.UPLOAD_DIR + ('/' if not settings.UPLOAD_DIR.endswith('/') else '') + filename
+        file = os.path.join(settings.UPLOAD_DIR, filename)
```
makseq, line 453: on windows os.path.join generates `\`, so the next line 
```
FileUpload.objects.filter(file=file).last()
```
stops working

label_studio/data_import/api.py
makseq: Maybe we can just raise an exception here? and say to users that their SVGs are not secure and we can import them?

makseq (changes requested):

label_studio/data_import/api.py
dvwright: sure, if that is what you want

label_studio/data_import/api.py
dvwright: sure

label_studio/data_import/models.py
dvwright: yes, but then we are modifying the original file, which can lead to unexpected results. what if our filter takes out too much, users (and us) may be confused.

label_studio/data_import/api.py
dvwright: personally, I believe this is the better approach. This way we don't overwrite users uploaded files, which can cause confusion. However i have moved these changes to happen on the file intake process as requested. cc @makseq

makseq: > This way we don't overwrite users uploaded files

yeah, it's a good point I missed. However, I have taken into account the performance: every time GET API for uploaded file has to clean the SVG. So, let's wrap import with cleanup into `SVG_SECURITY_CLEANUP` env variable. If some users have troubles, they can disable it.

label_studio/data_import/uploader.py
```diff
@@ -58,10 +59,55 @@ def check_file_sizes_and_number(files):
 
 def create_file_upload(request, project, file):
     instance = FileUpload(user=request.user, project=project, file=file)
+    content_type, encoding = mimetypes.guess_type(str(instance.file.name))
+    if content_type in ['image/svg+xml']:
```
makseq, line 63: We can add env var `SVG_SECURITY_CLEANUP` here

makseq (changes requested):

label_studio/data_import/api.py
dvwright: >  have taken into account the performance: every time GET API for uploaded file has to clean the SVG

yes, but only SVG files and this is a fast process, not a performance issue for label studio. 

I'll add the env var

label_studio/core/settings/base.py
```diff
@@ -486,3 +486,5 @@ def collect_versions_dummy(**kwargs):
 # default value for feature flags (if not overrided by environment or client)
 FEATURE_FLAGS_DEFAULT_VALUE = False
 
+# Strip harmful content from SVG files by default
+SVG_SECURITY_CLEANUP = True
```
makseq, line 490: SVG_SECURITY_CLEANUP = get_bool_env('SVG_SECURITY_CLEANUP', True)

makseq: our users should be able to do this without code modifications, we often ask them to fix things like this through the console

makseq (changes requested):

label_studio/core/settings/base.py
dvwright: > (if not overrided by environment or client)

The existing comment above makes how one should define/set these misleading. Why would it differ from how we set ```FEATURE_FLAGS_DEFAULT_VALUE```

also, i see we override settings with a conf file. i.e. ```core/settings/base.py```, i thought many uses would have their own.

i'll make the change you requested, just letting you know how i arrived at the decision.

makseq: FEATURE_FLAGS_DEFAULT_VALUE should be `FEATURE_FLAGS_DEFAULT_VALUE = get_env('FEATURE_FLAGS_DEFAULT_VALUE', True)` too, we missed it. 

> i thought many usesr would have their own.
pip users won't have it :-)

makseq (approved):

PR #2277, opened 2022-04-27
fix: DEV-2243: Fix creating labels config using just the dynamic data

KonstantinKorotaev:


label_studio/core/utils/schema/label_config_schema.json
```diff
@@ -140,7 +140,7 @@
     "Labels": {
       "type": "object",
       "anyOf": [
-        {"required": ["@name", "@toName", "Label"]},
+        {"required": ["@name", "@toName"]},
         {"required": ["@name", "@toName", "View"]}
```
niklub, line 144: I think this string unneeded then

niklub, line 144: I think this string unneeded then

niklub (approved):

label_studio/core/utils/schema/label_config_schema.json
KonstantinKorotaev: Fixed

PR #2283, opened 2022-04-29
fix: DEV-2237: Add config preview data for dynamic choices

Gondragos:


label_studio/core/label_config.py
```diff
@@ -292,6 +292,12 @@ def generate_sample_task_without_check(label_config, mode='upload', secure_mode=
                 task[value] = examples['HyperText']
         elif p.tag.lower().endswith('labels'):
             task[value] = examples['Labels']
+        elif p.tag.lower() == "choices":
+            allow_nested = p.get('allowNested') or p.get('allownested') or "false"
+            if allow_nested == "true":
```
makseq, line 297: allow_nested.lower() == "true"

makseq (changes requested):

label_studio/core/label_config.py
Gondragos: There may be only true or false due to validation.

makseq (approved):

PR #2286, opened 2022-04-29
fix: DEV-2285: Improve performance of list of projects

dvwright:
* use django cache for projects list
   - for now uses file based but could use any available, memory, redis, etc. it's just a config change.

### DIFF
--- a/label_studio/core/settings/base.py
+++ b/label_studio/core/settings/base.py
@@ -288,6 +288,14 @@ RQ_QUEUES = {
     },
 }
 
+CACHES = {
+    'default': {
+        # 'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+        'TIMEOUT': 86400,  # 24 hours
+    }
+}
+
 # Swagger: automatic API documentation
 SWAGGER_SETTINGS = {
     'SECURITY_DEFINITIONS': {
--- a/label_studio/projects/api.py
+++ b/label_studio/projects/api.py
@@ -146,7 +146,18 @@ class ProjectListAPI(generics.ListCreateAPIView):
             raise LabelStudioDatabaseException('Database error during project creation. Try again.')
 
     def get(self, request, *args, **kwargs):
-        return super(ProjectListAPI, self).get(request, *args, **kwargs)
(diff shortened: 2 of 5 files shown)


dvwright:
@makseq can you review this when you have time - https://heartex.atlassian.net/browse/DEV-2285 thanks!

label_studio/core/settings/base.py
```diff
@@ -288,6 +288,14 @@
     },
 }
 
+CACHES = {
+    'default': {
+        # 'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
```
makseq, line 293: I would use LocMemCache, because a lot of users run our app in k8s and k8s has some problems with its storage persistence.

dvwright: I was trying to avoid any memory related approaches

dvwright:
this isn't going in

label_studio/projects/api.py
```diff
@@ -146,7 +146,18 @@ def perform_create(self, ser):
             raise LabelStudioDatabaseException('Database error during project creation. Try again.')
 
     def get(self, request, *args, **kwargs):
-        return super(ProjectListAPI, self).get(request, *args, **kwargs)
+        from projects.cache import cached_projects_get, cached_projects_set
+
+        user_id = self.request.user.id
+        response = cached_projects_get(user_id)
```
makseq, line 152: Unfortunately, we cannot cache projects, they change intensively every second. We can cache only particular fields like `project.members` for 10-20 min on project cards. How often this cache will be refreshed?

makseq (changes requested):

label_studio/projects/api.py
dvwright: Sergei echoed the same response when I first brought this up. It sounded like you guys had other ideas which is why I stopped working this. However I think it's worth some more time exploring, which I will do on my down time.

I think the key to having this work is to focus on caching each counter. Each counter would have its own cache. They would be be updated/deleted on any model save and after a certain amount of time.

> How often this cache will be refreshed?

We can specify any amount of time. 5 minutes is the default. I also set the cache to be deleted *ANY* time a project is saved (for any reason). See ```label_studio/projects/models.py -> cached_projects_remove(self.created_by_id)```

@makseq

PR #2303, opened 2022-05-03
feature: relative prefix instead of only absolute hostname

juliusvonkohout:
Add the possibility to add a relative prefix instead of an absolute hostname. This is useful if you cannot easily get the full Hostname as for example on Kubernetes with an Istio ingressgateway.

### DIFF
--- a/label_studio/core/settings/base.py
+++ b/label_studio/core/settings/base.py
@@ -39,7 +39,7 @@ if HOSTNAME:
             FORCE_SCRIPT_NAME = match.group(3)
             if FORCE_SCRIPT_NAME:
                 logger.info("=> Django URL prefix is set to: %s", FORCE_SCRIPT_NAME)
-    elif HOSTNAME.startswith('/')
+    elif HOSTNAME.startswith('/'):
         logger.info(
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/patch-1*:refs/remotes/origin/patch-1* +refs/tags/patch-1*:refs/tags/patch-1*
  The process '/usr/bin/git' failed with exit code 1
```

this looks as if your checks are somewhat damaged.

makseq:
Could you please check and say how it works with user avatars, uploaded images using GUI ?

juliusvonkohout:
> Could you please check and say how it works with user avatars, uploaded images using GUI ?

it only works partially so far. I can login, see static images etc. What does not work is accessing /projects. /projects just returns a blank page

makseq:
yes, it might be, because it's a very tricky thing (HOSTNAME) and it affects a lot of deep things.

juliusvonkohout:
> yes, it might be, because it's a very tricky thing (HOSTNAME) and it affects a lot of deep things.

some script is not using it properly on the projects page
grafik

juliusvonkohout:
closing, since this shall be the task of the maintainers.

PR #2307, opened 2022-05-04
fix: DEV-2298: Increase long RQ jobs timeout

KonstantinKorotaev:


label_studio/core/settings/base.py
```diff
@@ -300,12 +300,6 @@
         'DB': 0,
         'DEFAULT_TIMEOUT': 180,
     },
-    'low': {
```
triklozoid, line 303: Why this part is removed? 
I thought job decorator parameter has a higher priority and can override default, isn't it?

KonstantinKorotaev: Reverted

triklozoid (approved):

AndrejOros (approved):

PR #2359, opened 2022-05-16
fix: DEV-2374: fix Runtime Error during import with no total_annotations and other

KonstantinKorotaev:


label_studio/tasks/serializers.py
```diff
@@ -300,14 +300,19 @@ def create(self, validated_data):
...
                     data=task['data'],
                     meta=task.get('meta', {}),
                     overlap=max_overlap,
                     is_labeled=len(task_annotations[i]) >= max_overlap,
                     file_upload_id=task.get('file_upload_id'),
-                    inner_id=max_inner_id + i
+                    inner_id=max_inner_id + i,
+                    total_predictions=len(task_predictions[i]),
+                    total_annotations=total_annotations,
+                    cancelled_annotations=cancelled_annotations
```
makseq, line 315: Most likely source storages are affected too.

makseq (changes requested):

label_studio/tasks/serializers.py
```diff
@@ -300,14 +300,19 @@ def create(self, validated_data):
             max_inner_id = (tasks.order_by("-inner_id")[0].inner_id + 1) if tasks else 1
 
             for i, task in enumerate(validated_tasks):
+                cancelled_annotations = len([ann for ann in task_annotations[i] if ann.get('was_cancelled', False)])
```
triklozoid, line 303: Why False? Annotation is cancelled when `was_cancelled == True`, I think.

KonstantinKorotaev: The condition is used if no was_cancelled field found in annotation

triklozoid (approved):

makseq (approved):

PR #2372, opened 2022-05-18
fix: DEV-1642: [Editor] Can't copy text from TextArea regions

yyassi-heartex:
this resolves an issue where users arent able to copy text from inputs and textareas.
I don't actually understand the intent of the code that was put in to handle copy/cut/paste (well before my time here + still unfamiliar with the all the available features) this is why i apprached this fix with an exception.

but it does read like the CurrentEntity component is over-reaching a bit, so a second look will likely needed once i have a better understanding of the intent behind this component


Travis1282 (approved):

yyassi-heartex:
closing this PR and recreating the branch to make sure, I'm branching out from `develop` instead of `master`

PR #2392, opened 2022-05-24
fix: 2295: extremely ambiguous error when attempting Labeling

dvwright:
* support import task list csv/txt like json


dvwright:
Hi @makseq not sure if this is the approach you want but this will resolve the issue [reported the user ](https://github.com/heartexlabs/label-studio/issues/2295)

PR #2393, opened 2022-05-24
feat: DEV-2132: Bounding box annotation UX improvements

yyassi-heartex:
this should get rotating bounding box working


yyassi-heartex:
recreating the branch/pr so its branching out from develop

PR #2399, opened 2022-05-25
fix: DEV-2456: Fix error when user tries to delete all tasks

KonstantinKorotaev:


label_studio/tasks/models.py
```diff
@@ -278,6 +279,20 @@ def save(self, *args, **kwargs):
...
         super().save(*args, **kwargs)
 
+    @staticmethod
+    def delete_tasks_without_signals(queryset):
+        """
+        Delete Tasks queryset with switched off signals
+        :param queryset: Tasks queryset
+        """
+        signals = [
+            (post_delete, update_is_labeled_after_removing_annotation, Annotation),
```
makseq, line 289: What if we add more signals? and forget add them here..

makseq (approved):

PR #2402, opened 2022-05-25
feat: DEV-2340: UX Annotation improvement: Always show crosshairs when hovered on image viewport

yyassi-heartex:
crosshair now enabled by default


robwijnhoven:
PR from May 25, 2022. 
Would be great if this could be merged, but someone has to dust off the code changes I am afraid.
Default crosshair would be very nice, as that is always wanted for box annotations!
Can we directly make the colors more visible? Preferable a very high-contrast color.

yyassi-heartex:
> PR from May 25, 2022. Would be great if this could be merged, but someone has to dust off the code changes I am afraid. Default crosshair would be very nice, as that is always wanted for box annotations! Can we directly make the colors more visible? Preferable a very high-contrast color.

Hi @robwijnhoven thanks for reaching out! We decided to discontinue development of this PR because its not needed. Enabling crosshairs can be done by updating the `crosshair` image tag https://labelstud.io/tags/image so I'l go ahead and close this PR

PR #2411, opened 2022-05-26
fix: DEV-2319: Local filenames are not URL encoded

dvwright:
* encode url

this is working for most characters

e.g.
```
    "GET /data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie--+--.jpeg HTTP/1.1" 404
    =>
    "GET /data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie--%2B--.jpeg HTTP/1.1" 200
```

however there is still a bug with files containing a `#` in the filename. 

for some reason that is getting stripped early on during `request` construction. I'm having a hard time finding where/when that is happening.

example: filename = `datasource/images/test/curiosity_selfie---#.jpeg`
```
[2022-05-27 17:11:54,292] [io_storages.utils::get_uri_via_regex::25] [WARNING] /data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie---#.jpeg does not match uri regex ([\"'])(?P<uri>(?P<storage>{})://[^\1=]*)\1
[27/May/2022 17:11:54] "GET /data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie--- HTTP/1.1" 404 0
```

```
<WSGIRequest: GET '/data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie---'>
```

```
select * from task
where id = 41452;
--
{
  "$undefined$": "/data/local-files/?d=Users/dwright/Documents/datasource/images/test/curiosity_selfie---#.jpeg"
}
```

### DIFF
--- a/label_studio/core/views.py
+++ b/label_studio/core/views.py
@@ -215,11 +215,7 @@ def localfiles_data(request):
         if user_has_permissions and os.path.exists(full_path):
             content_type, encoding = mimetypes.guess_type(str(full_path))
             content_type = content_type or 'application/octet-stream'
-            # full_path = "Users/dwright/Documents/datasource/images/test/LSE + + + project import 30K queues.jpeg"
-            response = RangedFileResponse(request, open(full_path, mode='rb'), content_type)
-            return response
-            # from django.http.response import FileResponse
-            # return FileResponse(request, open(full_path, mode='rb'), content_type)
+            return RangedFileResponse(request, open(full_path, mode='rb'), content_type)
         else:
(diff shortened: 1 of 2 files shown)


makseq:
I'm not sure that our users will be happy if they can't use their files with `+/\|`... characters. 
image

What do you think if we try to denormalize path somewhere here?
https://github.com/heartexlabs/label-studio/blob/DEV-2319/label_studio/core/views.py#L172

PR #2427, opened 2022-05-31
Fix: DEV-2203: Sidebar grid fix

Travis1282:
fix: DEV-2203 Sidebar grid fix


Travis1282:
The solution was already there but the override was not actually working in the css. the dreaded !important was used, which is not ideal however in this case this will only be necessary until the new sidebars are released as it is not an issue with them.

Gondragos:
The build have been delivered by other pull requests

PR #2428, opened 2022-05-31
fix: DEV-2459: tostring

juliosgarbi:
i've added .toString to convert integers to string before .replace


makseq:
this LSF build was merged in another LS PR (with LSF master branch).

PR #2452, opened 2022-06-03
Label-studio behind a proxy fails to reimport

### LINKED ISSUE #2451: Label-studio behind a proxy fails to reimport
**Describe the bug**
PR https://github.com/heartexlabs/label-studio/pull/1044 introduces an ready return for empty reimports. However, it responds with `204`.

While that may seem a good choice, it also wrong. According to the specification [1][2] `204` should not have a body.

This is not a problem when trying to access label-studio from a browser. As the documentation says, browsers have fault tolerance for this issue. The problem arises when using label-studio with postman or behind a proxy.

In my case, I am running label-studio behind a node/express/http-proxy-middleware proxy, which apparently adheres to the specification very strictly. In the case of label-studio this translates to label-studio failing to receive a response from the reimport and hanging.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204
[2] https://restfulapi.net/http-status-204-no-content/

**To Reproduce**
It is fairly easy as I said, using a node/express/http-proxy-middleware  stack , as well as using a tool like Postman.

**Expected behavior**
The expected behavior is a given for everyone I guess. Responses should reach label-studio, even if label-studio is behind a proxy.

jfragoulis:
PR #1044 [1] introduces an early return for empty reimports.
However, it responds with `204`.

While that may seem a good choice, it is also wrong. According to the
specification [2][3], `204` should not have a body.

This is not a problem when trying to access label-studio from a browser.
The problem arises when using label-studio behind a proxy.

Since the reimport's response, *needs* to reach label-stuiod, in addition
to having a body, it should be a simple `200`.

Fixes #2451

[1] https://github.com/heartexlabs/label-studio/pull/1044
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204
[3] https://restfulapi.net/http-status-204-no-content/

### DIFF
--- a/label_studio/data_import/api.py
+++ b/label_studio/data_import/api.py
@@ -311,7 +311,7 @@ class ReImportAPI(ImportAPI):
                 'file_upload_ids': [],
                 'found_formats': {},
                 'data_columns': []
-            }, status=status.HTTP_204_NO_CONTENT)
+            }, status=status.HTTP_200_OK)
 
         tasks, found_formats, data_columns = FileUpload.load_tasks_from_uploaded_files(
             project, file_upload_ids,  files_as_tasks_list=files_as_tasks_list)


farioas:
Hi @jfragoulis,

Thanks for your contribution. Somehow we missed your report and discovered this issue "a bit later":
https://github.com/heartexlabs/label-studio/pull/2751

Closing your PR. 

Thanks!

PR #2471, opened 2022-06-07
fix: DEV-2464: Move from signals to save and delete

KonstantinKorotaev:


label_studio/tasks/models.py
```diff
@@ -397,7 +426,40 @@ def update_task(self):
         self.task.save(update_fields=update_fields)
 
     def save(self, *args, **kwargs):
+        created = False if self.pk else True
```
makseq, line 429: too many lines of code here, let's split it into clear subfunctions and move them to functions.py as business layer logic

makseq (changes requested):

label_studio/tasks/models.py
KonstantinKorotaev: Fixed

PR #2486, opened 2022-06-09
fix: DEV-2303: Video regions are editable when anno history is selected

nicholasrq:


bmartel (approved):

makseq:
This branch was merged with another LS PR.

PR #2526, opened 2022-06-15
docs: DEV-2634 Remove Step 2 in "Configure access to your S3 bucket"

anithamuthu2022:


antopolskiy (changes requested):
I'd add a note that Session Token is only required in case of temporary security credentials, so that there is no confusion.

otherwise lgtm

makseq (approved):

anithamuthu2022:
> I'd add a note that Session Token is only required in case of temporary security credentials, so that there is no confusion.
> 
> otherwise lgtm

[s.antopolsky@gmail.com](mailto:s.antopolsky@gmail.com) could you test and confirm the change.

antopolskiy:
lgtm

PR #2536, opened 2022-06-17
fix: DEV-2658: Fix error for actions with ordering by joined field

KonstantinKorotaev:


label_studio/data_manager/actions/basic.py
```diff
@@ -88,7 +88,7 @@ def delete_tasks_annotations(project, queryset, **kwargs):
     annotations.delete()
     emit_webhooks_for_instance(project.organization, project, WebhookAction.ANNOTATIONS_DELETED, annotations_ids)
     start_job_async_or_sync(bulk_update_stats_project_tasks, queryset.filter(is_labeled=True))
-    start_job_async_or_sync(project.update_tasks_counters, queryset)
+    start_job_async_or_sync(project.update_tasks_counters, list(queryset))
```
makseq, line 91: what if we have 1m tasks here? they will be downloaded fully, but we can use only ids here

KonstantinKorotaev: Fixed

label_studio/projects/models.py
```diff
@@ -781,8 +782,12 @@ def update_tasks_counters(self, queryset, from_scratch=True):
         total_annotations = Count("annotations", distinct=True, filter=Q(annotations__was_cancelled=False))
         cancelled_annotations = Count("annotations", distinct=True, filter=Q(annotations__was_cancelled=True))
         total_predictions = Count("predictions", distinct=True)
-        if isinstance(queryset, list):
+        # construct QuerySet in case of list of Tasks
+        if isinstance(queryset, list) and isinstance(queryset[0], Task):
```
makseq, line 786: what if len(queryset)=0?

makseq (changes requested):

label_studio/projects/models.py
KonstantinKorotaev: It should, but added conditions to check length

makseq (approved):

PR #2555, opened 2022-06-22
fix: DEV-2687: Display source filename for S3 objects

triklozoid:


makseq (changes requested):
Filters & Ordering won't work, only the DM display will be ok. Maybe it's easier to have separate column for storages.. I understand that it's not trivial to concatenate all these storages + file_upload moreover..

makseq (approved):

PR #2558, opened 2022-06-23
feat: Experimental feature Add data field

makseq:
This PR adds the `Add Or Modify Data Field" action in the Data Manager.

**Number**
Add new data field with float value.

**String**
Add new data field with string value.

**Expressions**
* permutative sampling:
```
sample()

> sample()  # => if N_tasks=4 => 4, 1, 3, 2 
```

* uniform random from min integer to max integer:
```
random(<min_int>, <max_int>)

> random(1, 3)  # => 1, 3, 2, ..., N_tasks
```

* random sampling for string or integer choices with weights: 
```
choices(["<value1>", "<value2>", ...], [<weight1>, <weight2>, ...])

> choices(["a", "b"], [0.4, 0.6])  # => "a", "b", "b", "b", ..., N_tasks
```

* replace old substring to the new substring:
```
replace("old-string", "new-string")

> replace("a", "b")  # => abc => bbc
```

### DIFF
--- a/label_studio/data_manager/actions/experimental.py
+++ b/label_studio/data_manager/actions/experimental.py
@@ -1,5 +1,6 @@
 """This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
 """
+import random
 import logging
 import ujson as json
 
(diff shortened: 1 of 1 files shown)


hadiyouness10:
Hello, I added a new data field called random with an expression of random(min, max).
The column is added and the values are random as expected. Filtering is working too. However, when I try to sort, nothing changes

hadiyouness10:
It seems that adding images using the upload button and then creates the data field works fine. However, adding data using the cloud storage (locally) and then creating the data field doesn't allow ordering or filtering on the created field

neouranos:
> ```
> EXPERIMENTAL_FEATURES=1 label-studio
> ```

hello I can't run this command
image

PR #2572, opened 2022-06-28
Release/1.5.0

starphantom666:
Long text marking problem
If the text is too long, you have to go back to the top every time to select a label for marking and then return to marking, which is a waste of time; It is usually marked while reading. If you can select the text first and then mark, I think it is more comfortable


farioas:
Closing this PR as it's has wrong HEAD.

PR #2601, opened 2022-07-05
fix: Remove extra slash for file url

aisensiy:
This is the issue: https://github.com/heartexlabs/label-studio-frontend/issues/733

----

I am using label-studio with the `--host` parameter like this:

```
$ label-studio --host https://openbayes.com/jobs/abc/proxy/8080
```

Every thing works good until I import some images for ocr task. It seems that the generated urls in frontend have more than one `/` between the prefix and the upload path:

image

If I mannully remove the extra `/` in the 

```
/jobs/open-tutorials/jobs/proxy/8080 ---->//<---- data/upload/1/df0703be-AAMU4003561_45R1__011_0741_181125_224515_01_Manual_1.jpg
```

the url can access successfully.


makseq (approved):

aisensiy:
I do not get the reason that `gitleaks` failed...

PR #2623, opened 2022-07-07
fix: DEV-2669: align filter items and ensure visually displays selected authors correctly

bmartel:


bmartel:
This code is already present in develop. Closing.

PR #2669, opened 2022-07-14
feat: DEV-2304: Implement health check for rqworker

KonstantinKorotaev:


label_studio/core/redis.py
```diff
@@ -100,3 +102,30 @@ def start_job_async_or_sync(job, *args, **kwargs):
...
+    # check workers
+    workers = Worker.all(connection=_redis)
+    workers_count = len(workers)
+    if workers_count == 0:
+        return False, "No workers found. Start several workers."
+    if workers_count * RATIO < total_jobs:
+        error_text = f"Workers seems to be overloaded. {total_jobs} jobs in the queues."
+    # check if workers in correct state
+    if all([worker for worker in workers if worker.state in ['started', 'busy', 'idle']]):
+        return False, "No workers found. Start several workers."
```
makseq, line 130: change it to another text, it's the same as on the line 125, it's bad for debugging.

label_studio/core/views.py
```diff
@@ -78,10 +87,16 @@ def version_page(request):
...
+    redis = bool_from_request(request.GET, "redis", False)
+    if not redis:
+        logger.debug('Got /health request.')
+        return HttpResponse(json.dumps({
+            "status": "UP"
+        }))
+    else:
+        health_state = rqworker_healthcheck()
+        result = {"message": health_state[1]}
+        return JsonResponse(result, status=200 if health_state[0] else 500)
```
makseq, line 99: we need to return "status": "UP" as it was before, otherwise we will break a lot of our client's setups.

makseq (changes requested):

label_studio/core/redis.py
KonstantinKorotaev: Fixed

label_studio/core/views.py
KonstantinKorotaev: Fixed

makseq (approved):

farioas (changes requested):
This will not work in the case of SSL enabled on redis or pgsql side

Issue #2678, opened 2022-07-17
Very small maximum total size of all files uploaded during multipart (set at ~260 MB)


qanastek:
**Describe the bug**
I am pushing this issue and the pull request associated because the current maximum total size of all files uploaded during a multipart file drop from a local directory is set to only ~260 MB which is extremely low for modern datasets which used PNG, JPG, PDF or any high quality data file format.

**To Reproduce**
1. Go to your fresh label-studio install (for example, using Docker)
2. Create a project without adding data
3. Click on your project
4. Click on the "Import" button in the top right corner and drop 500 MB of JPG files
5. See error

**Expected behavior**
It will be much simpler for beginners and non IT users to have either a very high limit such as 10 GB or to be able to change this value through a dropdown menu.

**Screenshots**

![image](https://user-images.githubusercontent.com/19389475/179428078-b1d3534c-8da0-4775-a11c-183b04fda2d0.png)

**Environment (please complete the following information):**
 - OS: W10 using WSL 1
 - Label Studio Version : 1.5.0post0 (master)

**Additional context**

See the following pull-request

qanastek:
PR https://github.com/heartexlabs/label-studio/pull/2679

makseq:
You can override this var by setting environment variable DATA_UPLOAD_MAX_MEMORY_SIZE

qanastek:
As you can see [here](https://github.com/heartexlabs/label-studio/blob/8519eb1cc8e8a333794ca0dcb644173fa26f69e3/label_studio/core/settings/base.py#L415) you can change the value directly on the code or by increasing the environment variable.

PR #2679, opened 2022-07-17
Increase max size multipart to 10,48 GB

qanastek:
Issue #2678


farioas:
You can override this var by setting environment variable `DATA_UPLOAD_MAX_MEMORY_SIZE`

PR #2705, opened 2022-07-21
Improve sum expression

zachmullen:
`sum` is one of the built-in functions that can accept a raw generator expression. Also, when iterating over just dict values, use `values` instead of `items`.


sajarin:
hey @zachmullen I know it's been quite a while since you standed up this PR. `uploader.py` has changed considerably since then and so I'm closing this PR. Nonetheless, thank you for contributing here and helping to shape the project!

PR #2760, opened 2022-07-28
fix: DEV-2939: Adding new fields leads to error in label config validation

KonstantinKorotaev:


label_studio/projects/models.py
```diff
@@ -487,14 +487,6 @@ def validate_config(self, config_string, strict=False):
         if not fields_from_config:
             logger.debug(f'Data fields not found in labeling config')
             return
-        fields_from_config = {field.split('[')[0] for field in fields_from_config}  # Repeater tag support
```
niklub, line 490: can we comment with TODO: DEV-XXX reference?

KonstantinKorotaev: Added

MihajloHoma (approved):

niklub (approved):

PR #2837, opened 2022-08-17
Update columns on task patch API

makseq:


label_studio/tasks/api.py
```diff
@@ -224,6 +224,12 @@ def retrieve(self, request, *args, **kwargs):
         result['data'] = task.resolve_uri(result['data'], project)
         return Response(result)
 
+    def update(self, request, *args, **kwargs):
+        result = super().update(request, args, kwargs)
+        instance = self.get_object()
+        instance.project.summary.update_data_columns([instance])
```
triklozoid, line 230: Looks like we already call this method in post_save signal: https://github.com/heartexlabs/label-studio/blob/develop/label_studio/tasks/models.py#L655
It doesn't work? 

Also can we add a simple test about it?

PR #2840, opened 2022-08-17
fix: DEV-3212: Add validation to avoid users import local files using URL

guilhermemachado26:

LS users are able to use data import functionality to access local files on the running system, including environment variables, etc. This is achieved by using `file://` url schema (ie. `file:///etc/passwd`).

Submitting a URL like this will cause Label Studio to fetch the local file and add it to the project, which can then be accessed in the `/data/uploads` folder.

## Jira Ticket

https://heartex.atlassian.net/browse/DEV-3212

### DIFF
--- a/label_studio/data_import/uploader.py
+++ b/label_studio/data_import/uploader.py
@@ -174,6 +174,10 @@ def load_tasks(request, project):
         else:
             if settings.SSRF_PROTECTION_ENABLED and url_is_local(url):
                 raise ImportFromLocalIPError
+
+            if url.strip().startswith('file://'):
+                raise ValidationError('"url" is not valid')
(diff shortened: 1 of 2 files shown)


makseq (approved):

makseq:
> This is not the case. Now users lost ability to import files from local filesystem. I'd like to suggest restrict usage of "file://" to label-studio DATA_DIR or HOME_DIR
> 
> @makseq what do you think?

We have Local Storage functionality for file access on hard drives. It works the same way as you described, so I think it's ok to restrict file:// access here. Local Storage is more controllable and obvious way to do it.

AndrejOros (approved):

PR #2888, opened 2022-09-01
fix: DEV-3296: Add retry for bulk stats calculation

KonstantinKorotaev:


label_studio/tasks/models.py
```diff
@@ -764,8 +765,14 @@ def bulk_update_stats_project_tasks(tasks):
...
         for task in tasks:
             update_task_stats(task, save=False)
-        # start update query batches
-        bulk_update(tasks, update_fields=['is_labeled'], batch_size=settings.BATCH_SIZE)
+        try:
+            # start update query batches
+            bulk_update(tasks, update_fields=['is_labeled'], batch_size=settings.BATCH_SIZE)
+        except OperationalError as exp:
+            logger.debug("Operational error while updating task ")
+            time.sleep(settings.BATCH_JOB_RETRY_TIMEOUT)
```
triklozoid, line 773: This line will block the worker for one minute, maybe we can implement it with `queue.enqueue_in()` ? 
https://python-rq.org/docs/scheduling/

KonstantinKorotaev: Fixed, but in case there would be params for queue name and timeout - they will be lost in second job

wesleylima (approved):

PR #2899, opened 2022-09-02
feat: DEV-239: Added support to import jsonl files via API

guilhermemachado26:

These changes allow users to import tasks using jsonl files via API

## Jira Ticket

https://heartex.atlassian.net/browse/DEV-239

### DIFF
--- a/label_studio/tests/data_import/test_models.py
+++ b/label_studio/tests/data_import/test_models.py
@@ -1,0 +1,62 @@
+import pytest
+
+from django.core.files.base import ContentFile
+
+from data_import.models import FileUpload
+
+pytestmark = pytest.mark.django_db
+
+
+class TestFileUpload:
+    @pytest.fixture
+    def jsonl_file(self, configured_project):
+        return FileUpload.objects.create(
+            user=configured_project.created_by,
+            project=configured_project,
+            file=ContentFile(
+                '{"id":100,"annotations":[],"file_upload":"d123c1f3-bb1f75c8cb.json","drafts":[],"predictions":[100],"data":{},"meta":{}, "updated_by":1,"comment_authors":[]}\n'
+                '{"id":120,"annotations":[],"file_upload":"d456c1f3-bb1f75c8cb.json","drafts":[],"predictions":[120],"data":{},"meta":{}, "updated_by":2,"comment_authors":[]}\n',
+                name="data.jsonl",
+            ),
+        )
+
+    def test_read_tasks_calls_jsonl_reader(self, mocker, jsonl_file):
+        mocked_read_jsonl = mocker.patch(
+            "data_import.models.FileUpload.read_tasks_list_from_jsonl"
+        )
+
+        jsonl_file.read_tasks()
+        mocked_read_jsonl.assert_called_once()
+
+    def test_read_tasks_list_from_jsonl(self, jsonl_file):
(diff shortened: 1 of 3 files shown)


label_studio/tests/data_import/test_models.py
```diff
@@ -0,0 +1,62 @@
...
+
+from data_import.models import FileUpload
+
+pytestmark = pytest.mark.django_db
+
+
+class TestFileUpload:
+    @pytest.fixture
+    def jsonl_file(self, configured_project):
+        return FileUpload.objects.create(
```
triklozoid, line 13: I think it's better to use API for creating objects. 

Because in this case you can refactor implementation as much as you want, and you don't need to touch test on every change.

guilhermemachado26: @triklozoid  Sorry, I didn't understand. Are you suggesting to make an API request to create the FileUpload ?

triklozoid: @guilhermemachado26 

Yes, I suggest to make test on API level. 
* call import API with jsonl body
* call tasks list API to check, that import works well. 

Example of test: https://github.com/heartexlabs/label-studio/blob/fb-dev-239-api/label_studio/tests/data_manager/test_views_api.py#L10

triklozoid: @guilhermemachado26  Why do you think we need a separate test for the read_tasks_list_from_jsonl method? 
For example, if I will add the required field for the FileUpload model, this test will fail, because it uses `FileUpload.objects.create`. But this change can be completely unrelated to JsonL logic. So it will increase support costs for the project. It's ok if we can catch some bugs but what bugs we can catch, that is not covered with API level test?

guilhermemachado26: @triklozoid I do agree about stop using `FileUpload.objects.create`. I'd like to suggest [FactoryBoy](https://factoryboy.readthedocs.io/en/stable/orms.html) to create these objects, in way that we would only need to add changes in one single place.

I believe its a good practice each test only test one function. Testing methods individually keeps the code more organized and simple, once API tests have extra overhead. Also, the method might be called from other parts of the project, not only from the API flow. If we decide to update/remove the API, we will lose test coverage.

guilhermemachado26: Just to be clear, I'm completely agree to test the API to check if the whole flow works as expected and the method was called, but some methods may have complex implementation, different inputs and couple of edge cases to be tested, and maybe they should be tested individually.

triklozoid: > but some methods may have complex implementation, different inputs and couple of edge cases to be tested, and maybe they should be tested individually.

Generally, I agree, but for this particular function and tests, I see neither complex implementation or edge cases. 

If we already have it, okay, let's keep it. But for similar functions in the future, I think we don't need tests.

triklozoid (approved):

label_studio/tests/data_import/test_models.py
guilhermemachado26: Sure, thanks for your feedback

robot-ci-heartex:
Hi @guilhermemachado26!

Unfortunately you don't have membership in heartexlabs organization, your PR wasn't synced with label-studio-enterprise.

PR #2915, opened 2022-09-08
fix: DEV-1904: Fixed source storage duplicating tasks when double click Sync button

guilhermemachado26:

Users are being able to click sync button as many times as they want. 

In order to avoid running multiple sync jobs at the same time, as a first approach, we have decided to implement a method which checks if there is a sync job already in the queue for that specific project before start the sync.
## Jira Ticket

https://heartex.atlassian.net/browse/DEV-1904


makseq (approved):

makseq:
I would add simple test here where sync is called twice.

label_studio/core/redis.py
```diff
@@ -100,3 +100,16 @@ def start_job_async_or_sync(job, *args, **kwargs):
         return job
     else:
         return job(*args, **kwargs)
+
+
+def is_job_in_queue(queue, func_name, **kwargs):
+    jobs = (
```
KonstantinKorotaev, line 106: It should be checked in case job is already in progress on worker

guilhermemachado26: @KonstantinKorotaev @makseq  The job doesn't need to be necessarily in progress, right ? It could be in pending status and we still need to avoid new background syncs. My question is: Does the job remain in the queue after executed ?

makseq: @guilhermemachado26 no, the job will be erased after the execution. Yes, the job can be in queue, but not running. Sounds like we should check all jobs in queue + all running jobs.

guilhermemachado26: Sure, adding some tests to cover these cases

makseq (approved):

PR #2995, opened 2022-09-23
fix: pytz dependency conflict with celery requiring newer version.

melihozaydin:


makseq:
Could we use 
```
pytz~=2019.3
````
?
`>=` is too wide..

melihozaydin:
> @makseq Could we use
> 
> ```
> pytz~=2019.3
> ```
> 
> ? `>=` is too wide..

yes, that probably would also be sufficient.

mo-arvan:
This issue is causing a dependency hell in my work. Is there a reason this commit hasn't been merged already? I could give you a hand if it requires more work.

melihozaydin:
> This issue is causing a dependency hell in my work. Is there a reason this commit hasn't been merged already? I could give you a hand if it requires more work.

yeah the CI/CD seems to have failed. I dont know why or how tough. Logs have expired. Seems like this repo is not looked after that well. 

I also had issues with "rq" as it is not auto installed i had to manually install it. Even tough it is a subdependency.

makseq:
@mo-arvan @melihozaydin 
>  Is there a reason this commit hasn't been merged already?

the reason is here:
https://github.com/heartexlabs/label-studio/pull/2995#issuecomment-1256822033

makseq (approved):

PR #3001, opened 2022-09-26
fix: DEV-1943: After awhile, shapes disappear

yyassi-heartex:


yyassi-heartex:
no longer needed as it appears to have been resolved with another merge

Issue #3033, opened 2022-09-29
Token renewal button breaks "example projects curl" 


Ben-Epstein:
**Describe the bug**

**To Reproduce**
1. Go to local UI
2. Click on Account and Settings
3. Scroll down to 'Example fetch projects data:'
4. Click "refresh" on the token
5. Copy and paste example fetch curl into terminal

**Expected behavior**
Should get the projects, instead get error because the curl is incorrectly formatted

**Screenshots**
the closing quote comes after `Token` instead of after the actual token
<img width="613" alt="image" src="https://user-images.githubusercontent.com/22605641/193112860-f53d37b9-ed13-4ec6-9391-1fb44d73ef99.png">

**Environment (please complete the following information):**
 - OS: [e.g. iOS] MacOS
 - Label Studio Version [e.g. 0.8.0] `1.6.0`

**Additional context**
The error only happens after refreshing the token

https://user-images.githubusercontent.com/22605641/193113632-f010771f-5fd9-4c60-b58e-6f17bc7abf52.mov

Ben-Epstein:
opened a small pr here to fix https://github.com/heartexlabs/label-studio/pull/3034

PR #3034, opened 2022-09-29
fix: Token renewal button [issue #3033]

Ben-Epstein:
fixes https://github.com/heartexlabs/label-studio/issues/3033

### DIFF
--- a/label_studio/users/templates/users/user_account.html
+++ b/label_studio/users/templates/users/user_account.html
@@ -206,7 +206,7 @@
           .then(res => res.json())
           .then(res => {
             input.value = res.token;
-            example.value = `${CURL_TEMPLATE} ${res.token}'`
+            example.value = `curl -X GET ${hostname}/api/projects/ -H 'Authorization: Token ${res.token}'`
             button.classList.add('blink');
             setTimeout(() => button.classList.remove('blink'))
           });
--- a/label_studio/users/templates/users/user_account.html
+++ b/label_studio/users/templates/users/user_account.html
@@ -180,10 +180,6 @@
       {% else %}
       const hostname = 'http://localhost:8080';
       {% endif %}
-      const CURL_TEMPLATE = "curl -X GET {{ settings.HOSTNAME }}/api/projects/ -H 'Authorization: Token"
-      {% else %}
-      const CURL_TEMPLATE = "curl -X GET http://localhost:8080/api/projects/ -H 'Authorization: Token"
-      {% endif %}
 
       document.querySelectorAll('[data-copy]').forEach(button => {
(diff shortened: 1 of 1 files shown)


label_studio/users/templates/users/user_account.html
```diff
@@ -176,9 +176,9 @@
   <script>
     (() => {
       {% if settings.HOSTNAME %}
```
wesleylima, line 178: Hi @Ben-Epstein. Thanks for the fix!

What do you think think of adding another javascript const ?

```suggestion
      {% if settings.HOSTNAME %}
      const hostname = {{ settings.HOSTNAME }};
      {% else %}
      const hostname = 'http://localhost:8080';
      {% endif %}
```

Ben-Epstein: hey, sure! Works for me, and i think this is cleaner.

label_studio/users/templates/users/user_account.html
```diff
@@ -176,9 +176,13 @@
...
-      const CURL_TEMPLATE = "curl -X GET {{ settings.HOSTNAME }}/api/projects/ -H 'Authorization: Token'"
+      const hostname = {{ settings.HOSTNAME }};
       {% else %}
-      const CURL_TEMPLATE = "curl -X GET http://localhost:8080/api/projects/ -H 'Authorization: Token'"
+      const hostname = 'http://localhost:8080';
+      {% endif %}
+      const CURL_TEMPLATE = "curl -X GET {{ settings.HOSTNAME }}/api/projects/ -H 'Authorization: Token"
+      {% else %}
+      const CURL_TEMPLATE = "curl -X GET http://localhost:8080/api/projects/ -H 'Authorization: Token"
```
wesleylima, line 185: `CURL_TEMPLATE` isn't necessary anymore. These lines can be removed.

Ben-Epstein: why is the curl template not necessary?

wesleylima: `CURL_TEMPLATE` was being used to set `example.value` on line 205 before this change. Now example.value is being constructed using the `hostname` const.

makseq:
Please rename this PR to more clear, "fix 3033" is saying nothing..

Ben-Epstein:
@makseq i updated it, but the explanation of the issue is in the issue, which is linked.

label_studio/users/templates/users/user_account.html
Ben-Epstein: updated @wesleylima

wesleylima (approved):

PR #3105, opened 2022-10-18
fix: DEV-3465: Projects, displayed on user's organization page, include other organizations

makseq:


KonstantinKorotaev:
Can we create a test for this case?

PR #3114, opened 2022-10-19
fix: DEV-3397: Added session expiration settings and conditions

wesleylima:
Added session expiration settings and conditions based on last login and last activity. Also made persistent session optional on login.


label_studio/core/middleware.py
```diff
@@ -153,3 +154,21 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+
+    def process_request(self, request) -> None:
+        if not hasattr(request, "session") or request.session.is_empty():
+            return
+
+        current_time = time.time()
+        last_login = request.session['last_login'] if 'last_login' in request.session else 0
+
+        # Check if this request is too far from when the login happened
+        if (current_time - last_login) > settings.MAX_SESSION_AGE:
```
hakan458, line 170: In what case is last_login not in request.session, and therefore set to 0? In that case we will log the user out here right?

wesleylima: Correct. The effect here is that everyone will be logged out whenever this update is pushed since the session doesn't know how old the login is. Maybe that's too aggressive? But it will only happen once ever. I opted not to try to write a "now" default since that would be slightly more complex and could possibly open the session up to be extended indefinitely.

hakan458 (approved):

MihajloHoma (approved):

PR #3293, opened 2022-11-23
Feature/duplicate entry

shayantabatabaee:
This pull request is related to https://github.com/heartexlabs/label-studio/issues/3292, please kindly review it.

Thanks in advance


label_studio/tasks/serializers.py
```diff
@@ -268,16 +268,14 @@ def create(self, validated_data):
...
-
             # add tasks first
             max_overlap = self.project.maximum_annotations
 
-            # identify max inner id
             tasks = Task.objects.filter(project=self.project)
-            prev_inner_id = tasks.order_by("-inner_id")[0].inner_id if tasks else 0
-            max_inner_id = (prev_inner_id + 1) if prev_inner_id else 1
-
+            all_inner_ids = [inner_id[0] for inner_id in tasks.values_list('inner_id')]
```
makseq, line 275:
```
[inner_id[0] for inner_id in tasks.values_list('inner_id')]
```
=>
```
list(tasks.values_list('inner_id', flat=True))
```

makseq (changes requested):
Thanks for this curious PR! 

I see at least two issues:
1. Big one: changing inner_id this way you break pipelines that already use inner_id as integer. So, it's better to rename it to unique_id
2. There should be something similar in source cloud storages:
https://github.com/heartexlabs/label-studio/pull/3293/files#diff-3829314442f71e22038252a7dd1b60e33066aa55ad7462fed59a398ad7ecb820R277

shayantabatabaee:
Thanks for reviewing our changes, 

We reverted the commits that affected `inner_id` field, Also we added `unique_id` field for each task with same functionality. 
For the second issue, we have checked importing data through file and URL, and it seems to work fine. I would really appreciate that if you could explain more for the second issue if any further changes need to make.

makseq:
> Thanks for reviewing our changes,
> 
> We reverted the commits that affected `inner_id` field, Also we added `unique_id` field for each task with same functionality. For the second issue, we have checked importing data through file and URL, and it seems to work fine. I would really appreciate that if you could explain more for the second issue if any further changes need to make.

Thanks for updates!

https://github.com/heartexlabs/label-studio/blob/develop/label_studio/io_storages/base_models.py#L138 - check this regarding import storages. Try to add any source storage and sync tasks form there.

shayantabatabaee:
Thanks for your instruction, 

We added an ability to read `unique_id` from cloud storage source and also prevent from duplicate entry if task with same `unique_id` already exists in database.

shsaronian:
Hi, 
Adding a quick comment to prevent this pr from going stale and closing. There was no activity on our side as we are still awaiting further progress on this pr. 

Thank you

farioas:
Closed as stale

PR #3294, opened 2022-11-23
fix: DEV-3792: Change condition for overlap in next task

KonstantinKorotaev:


label_studio/projects/functions/next_task.py
```diff
@@ -135,7 +135,17 @@ def get_not_solved_tasks_qs(user, project, prepared_tasks, assigned_flag, queue_
 
     # if annotator is assigned for tasks, he must to solve it regardless of is_labeled=True
     if not assigned_flag:
-        not_solved_tasks = not_solved_tasks.filter(is_labeled=False)
+        # get project params for tasks that should have maximum_annotations
+        must_tasks = int(project.tasks.count() * project.overlap_cohort_percentage / 100 + 0.5)
```
makseq, line 139: required_tasks

label_studio/projects/functions/next_task.py
```diff
@@ -135,7 +135,17 @@ def get_not_solved_tasks_qs(user, project, prepared_tasks, assigned_flag, queue_
...
+        # get tasks count with maximum_annotations or greater
+        tasks_with_max_annotations = project.tasks.filter(total_annotations__gte=project.maximum_annotations).count()
+        # identify what overlap should be for next task
+        if must_tasks > tasks_with_max_annotations:
+            overlap = project.maximum_annotations
+        else:
+            overlap = 1
+        # filter tasks that should be annotated
+        not_solved_tasks = not_solved_tasks.filter(total_annotations__lt=overlap).order_by("-total_annotations")
```
makseq, line 148: seems we need index for "-total_annotations"?

makseq (changes requested):

label_studio/projects/functions/next_task.py
KonstantinKorotaev: Added

label_studio/projects/functions/next_task.py
KonstantinKorotaev: Fixed

label_studio/projects/functions/next_task.py
triklozoid: @KonstantinKorotaev  Can you check please, that postresql use this index? 
A simple way to do it: 
* restart LSE with `SILK_PROFILER=1`
* make API request
* open `/silk` url in app
* check SQL tab with explain analyze output. 

If the index is not used, let's remove it, because having an additional index makes slower insert and update operations.

KonstantinKorotaev: Index is used:

> Index Scan Backward using task_total_a_bc99f8_idx on task  (cost=3248.12..7437.04 rows=329 width=330) (actual time=3.442..3.443 rows=1 loops=1)
>         Index Cond: (total_annotations < 1)

makseq (changes requested):
Is this situation covered by test?

KonstantinKorotaev:
> Is this situation covered by test?

next_task.tavern.yml covers this case

PR #3297, opened 2022-11-23
fix: DEV-3743: Fixing aggregation for annotation results

wesleylima:


makseq (changes requested):

makseq:
Please add a test for this case, when anno result is empty

label_studio/data_manager/managers.py
```diff
@@ -492,7 +492,7 @@ def annotate_annotations_results(queryset):
         return queryset.annotate(annotations_results=Coalesce(
             GroupConcat("annotations__result"), Value(''), output_field=models.CharField()))
     else:
-        return queryset.annotate(annotations_results=ArrayAgg("annotations__result", distinct=True))
+        return queryset.annotate(annotations_results=ArrayAgg("annotations__result", distinct=True, default=Value([])))
```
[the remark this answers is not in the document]
wesleylima: This worked properly in sqlite before. Just re-confirmed now.

label_studio/data_manager/aggregates.py
```diff
@@ -0,0 +1,81 @@
...
+
+from django.db.models.lookups import Exact
+from django.db.models import Func, Value
+from django.contrib.postgres import lookups
+from django.contrib.postgres.aggregates import ArrayAgg as OriginalArrayAgg
+from django.contrib.postgres.fields import ArrayField
+from django.contrib.postgres.fields.array import ArrayLenTransform as OriginalArrayLenTransform, ArrayInLookup as OriginalArrayInLookup
+
+
+class ArrayRHSMixin:
```
makseq, line 16: I would solve it using another `if` statement here: https://github.com/heartexlabs/label-studio/blob/develop/label_studio/data_manager/managers.py#L195

Will it correctly work it with `[]` (empty list), when annotation.result = [] (instead of annotation.result = None) ?

label_studio/tests/tasks_api.tavern.yml
```diff
@@ -24,4 +24,161 @@ stages:
     method: GET
     url: '{django_live_url}/api/tasks/{task_pk}?project={project_pk}'
   response:
-    status_code: 200
\ No newline at end of file
+    status_code: 200
+
```
makseq, line 28: For data manager filters we have a special folder:
image

makseq (changes requested):

makseq:
@wesleylima Are you sure you reverted this branch to the correct commit? Please check this comment: https://github.com/heartexlabs/label-studio/pull/3297/files#r1052506144

makseq (changes requested):
Are you sure you reverted this branch to the correct commit? 
https://github.com/heartexlabs/label-studio/pull/3297/files#r1052506144

label_studio/data_manager/aggregates.py
wesleylima: I don't want to make too many changes to this file. It's just a "fast forward" of aggregates.py to a newer django version without requiring the latest django version. I think it completely addresses the bug we're trying to solve as is.

makseq: I still have two concerns here:
1. performance, how will it be affected? 
2. from user perspective `annotation.result = []` is empty annotation (you can try to get this by clicking submit without any region creation). However, this approach will process [] as not empty annotation.

makseq (changes requested):

label_studio/data_manager/managers.py
```diff
@@ -195,7 +194,13 @@ def add_result_filter(field_name, _filter, filter_expressions, project):
     elif _filter.operator == Operator.NOT_CONTAINS:
         filter_expressions.append(~Q(subquery))
         return 'continue'
-
+    elif _filter.operator == Operator.EMPTY and field_name == 'annotations_results':
+        if _filter.value == 'true':
```
makseq: Let's edit this a little: 
```  
cast_value(_filter)  # it works inplace
if _filter.value: 
  ...
```

makseq (changes requested):

label_studio/data_manager/managers.py
wesleylima: yes this is cleaner. Unfortunately the UI is sending us the "annotation_results is empty" filter typed as a String so `cast_value` won't cast it into a Boolean
```
{
  "filter": "filter:tasks:annotations_results",
  "operator": "empty",
  "value": "true",
  "type": "String"
}
```

makseq (approved):

makseq (approved):

PR #3318, opened 2022-11-28
feat: DEV-3259: Updated_by field on annotation level for "Update", "Accept" and "Reject"

hakan458:


label_studio/tasks/migrations/0032_annotation_updated_by_fill.py
```diff
@@ -0,0 +1,53 @@
...
+from django.conf import settings
+from django.db import migrations, models
+from django.db.models import F
+
+from projects.models import Project
+from tasks.models import Annotation
+
+
+def _fill_annotations_updated_by(project):
+    Annotation.objects.filter(project=project).update(updated_by=F('completed_by'))
```
makseq, line 14: I guess the most correct value for updated by is stored in AnnotationHistory, so probably we can use something like:

```
from annotation_history.models import AnnotationHistory as AH

subquery = AnnotationHistory.objects.filter(annotation_id=OuterRef('pk'), action_type=AH.ActionType.UPDATED).order_by('id')[0])

Annotation.objects
  .filter(project=project).
  .annotate(tmp=subquery)
  .update(updated_by=F('tmp__created_by'))
```
or if postgres doesn't support this, we can rewrite it to simple form with bulk_update.

label_studio/tasks/migrations/0032_annotation_updated_by_fill.py
```diff
@@ -0,0 +1,53 @@
...
+    projects = Project.objects.all()
+    for project in projects:
+        migration = AsyncMigrationStatus.objects.create(
+                project=project,
+                name='0032_annotation_updated_by_fill',
+                status=AsyncMigrationStatus.STATUS_STARTED,
+        )
+
+        start_job_async_or_sync(_fill_annotations_updated_by, project)
+        migration.status = AsyncMigrationStatus.STATUS_FINISHED
```
makseq, line 38: this status can't be changed here, it should be done inside of _fill_annotations_updated_by after all calculations

makseq (changes requested):

makseq (approved):

triklozoid (approved):

PR #3377, opened 2022-12-06
fix: DEV-3911: Move persistent storages to OS

makseq:


label_studio/core/settings/base.py
```diff
@@ -556,3 +556,46 @@ def collect_versions_dummy(**kwargs):
...
+    AWS_S3_ENDPOINT_URL = get_env('MINIO_STORAGE_ENDPOINT')
+    AWS_QUERYSTRING_AUTH = False
+    # make domain for FileUpload.file
+    AWS_S3_SECURE_URLS = False
+    AWS_S3_URL_PROTOCOL = 'http:' if HOSTNAME.startswith('http://') else 'https:'
+    AWS_S3_CUSTOM_DOMAIN = HOSTNAME.replace('http://', '').replace('https://', '') + '/data'
+
+if get_env('STORAGE_TYPE') == "s3":
+    CLOUD_FILE_STORAGE_ENABLED = True
+    DEFAULT_FILE_STORAGE = 'htx.storage.CustomS3Boto3Storage'
```
triklozoid, line 575: `htx.storage.CustomS3Boto3Storage` does not exist in this repository

makseq: thanks, I am moving it there

label_studio/data_import/api.py
```diff
@@ -465,3 +470,43 @@ def get(self, *args, **kwargs):
...
         else:
             return Response(status=status.HTTP_404_NOT_FOUND)
+
+
+class DownloadStorageData(APIView):
+    """ Check auth for nginx auth_request
+    """
+    swagger_schema = None
+    http_method_names = ['get']
+    permission_classes = (IsAuthenticated, )
```
niklub, line 480: should RBAC be presented here or how it works?

triklozoid: @niklub  No, this call is available for all authenticated users, we have permissions check inside `get` code: 
```
if file_upload is not None and file_upload.has_permission(request.user):
```

PR #3406, opened 2022-12-11
Feature/export_filter

shayantabatabaee:
In this pull request, We added the `from_creation_date` and `to_creation_date` to the service `api/projects/{projectId}/export`. This fields filtered tasks based on date.  Also we added an ability that affected Export button in the DataManager. So if any filter is set the Export button would only export filtered data.
Please kindly review, and let me know if any change is needed.

Best regards


shayantabatabaee:
This PR is related to https://github.com/heartexlabs/label-studio/issues/3458. Please kindly review it.

makseq:
Sorry but filters are supported in the export snapshots:
https://labelstud.io/guide/export.html#Export-snapshots-using-the-API
You can use it via SDK or API. Probably we will add GUI for this, right now the GUI available in enterprise only.

PR #3408, opened 2022-12-12
docs: Audio Player feature documentation

anithamuthu2022:
[DOC-9]

[DOC-9]: https://heartex.atlassian.net/browse/DOC-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


makseq (changes requested):
@anithamuthu2022 we should write how to switch to the old audio engine in the very beginning of the article. @bmartel please help here

bmartel:
> @anithamuthu2022 we should write how to switch to the old audio engine in the very beginning of the article. @bmartel please help here

@anithamuthu2022 It should be stated both that this is a new alternative to the existing audio engine, and how to go back to a previous version in case there are issues for the user's workflow. 

For LabelStudio Opensource users who wish to return to a previous audio engine version, they need to set an environment variable on their operating instance:
```shell
ff_front_dev_2715_audio_3_280722_short=false
```

anithamuthu2022:
@bmartel @makseq I've addressed your feedback for review/approval. Thank you!

anithamuthu2022:
@makseq @nicholasrq @juliosgarbi @farioas this PR is ready for review/approval.

PR #3466, opened 2022-12-21
fix: DEV-4074: Fix task matching query does not exist status code to 404

KonstantinKorotaev:


label_studio/tasks/api.py
```diff
@@ -192,7 +192,11 @@ def get_queryset(self):
             kwargs = {'all_fields': True}
         project = self.request.query_params.get('project') or self.request.data.get('project')
         if not project:
-            project = Task.objects.get(id=self.request.parser_context['kwargs'].get('pk')).project.id
+            try:
```
niklub, line 195: I don't feel it is the right way to fix it - try simply returning 404 instead of 500

niklub (changes requested):

label_studio/tasks/api.py
KonstantinKorotaev: Fixed, moved to get_object_or_404

label_studio/tasks/api.py
```diff
@@ -182,6 +182,11 @@ def get(self, request, pk):
         return Response(data)
 
     def get_queryset(self):
+        task_id = self.request.parser_context['kwargs'].get('pk')
+        task = generics.get_object_or_404(
+            Task.objects.filter(id=task_id),
```
niklub: why do we need this queryset here? why not `generics.get_object_or_404(Task, pk=task_id)` ?

triklozoid (approved):

MihajloHoma (approved):

PR #3482, opened 2022-12-23
fix: DEV-4090: Prevent annotations duplicating

triklozoid:


KonstantinKorotaev (approved):

hlomzik:
There is a branch name clash here and in #3491 ("dev" capitalization), so I merged them all into #3491 as it has more correct naming and it's connected with other repos.

Issue #3337, opened 2022-11-30
Error when creating filter for storage filename


nazikus:
**Describe the bug**
When creating a new view filter based on storage filename, exception pops up.

**To Reproduce**
1. Go to Project view
2. Click on 'Filters', select "Storage filename" "contains"
3. Exception pop ups

**Expected behavior**
Expect to filter tasks by storage filename filter.

**Screenshots**
![image](https://user-images.githubusercontent.com/2558905/204758817-e59f82bd-73af-4692-a5ba-9babe303da3d.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04, docker image tag 1.6.0 
 - Label Studio Version 1.6.0rc5

**Additional context**

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/sql/query.py", line 1955, in add_fields
    join_info = self.setup_joins(name.split(LOOKUP_SEP), opts, alias, allow_many=allow_m2m)
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/sql/query.py", line 1648, in setup_joins
    path, final_field, targets, rest = self.names_to_path(
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/sql/query.py", line 1562, in names_to_path
    raise FieldError("Cannot resolve keyword '%s' into field. "
django.core.exceptions.FieldError: Cannot resolve keyword 'storage_filename' into field. Choices are: annotations, cancelled_annotations, comment_authors, comment_count, created_at, data, drafts, file_upload, file_upload_id, id, inner_id, io_storages_azureblobimportstoragelink, io_storages_gcsimportstoragelink, io_storages_localfilesimportstoragelink, io_storages_redisimportstoragelink, io_storages_s3importstoragelink, is_labeled, last_comment_updated_at, locks, meta, overlap, predictions, project, project_id, total_annotations, total_predictions, unresolved_comment_count, updated_at, updated_by, updated_by_id

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/label-studio/label_studio/data_manager/api.py", line 202, in get
[...]

PR #3494, opened 2022-12-27
fix: DEV-3337: add storage_filename to filters (contains and exact) 

chudomozg:
Above all, thank you for label-studio. I would like to point out that I am pushing to opensource project first time. So please, don't be too harsh if i do something wrong.

There is issue [#3337](https://github.com/heartexlabs/label-studio/issues/3337) about the inability to filter by storage_name. I added ability to filtering by storage_filename. I've used [django-queryable-properties](https://github.com/W1ldPo1nter/django-queryable-properties) library. So, now we can use any properties for filters. I have only implemented a few filters: icontains and exact. Because there are some trouble with __isnull filter and regex.

### DIFF
--- a/label_studio/tasks/models.py
+++ b/label_studio/tasks/models.py
@@ -21,6 +21,7 @@ from django.utils.timesince import timesince
 from django.utils.timezone import now
 from django.dispatch import receiver, Signal
 from django.core.files.storage import default_storage
+from queryable_properties.properties import queryable_property
 from rest_framework.exceptions import ValidationError
 
 from core.feature_flags import flag_set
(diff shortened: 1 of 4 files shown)


label_studio/tasks/models.py
```diff
@@ -45,9 +45,10 @@ class Task(TaskMixin, models.Model):
     """ Business tasks from project
     """
     id = models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID', db_index=True)
-    data = JSONField('data', null=False, help_text='User imported or uploaded data for a task. Data is formatted according to '
-                                                   'the project label config. You can find examples of data for your project '
-                                                   'on the Import page in the Label Studio Data Manager UI.')
+    data = JSONField('data', null=False,
```
makseq, line 48: please don't change lines that are not related to your PR

makseq (changes requested):

chudomozg:
I've found some problem in my code when I was testing it. I will fix it and open new merge request.

PR #3519, opened 2023-01-06
fix: DEV-3771: storage_data now responds to HEAD request without error

wesleylima:


hakan458 (approved):

label_studio/data_import/api.py
```diff
@@ -468,7 +468,7 @@ class DownloadStorageData(APIView):
     """ Check auth for nginx auth_request
     """
     swagger_schema = None
-    http_method_names = ['get']
+    http_method_names = ['get', 'head']
```
makseq, line 472: don't we need to implement def head() or something like this in DownloadStorageData view? 
Probably I don't know how it works exactly and how will django generate head api calls... but in GET response we have this headers: 

```
 response['X-Accel-Redirect'] = redirect
response['Content-Disposition'] = 'attachment; filename="{}"'.format(filepath)
```

will they be rendered in head automatically?

makseq (changes requested):

label_studio/data_import/api.py
makseq: BTW, I am not sure this implementaion of DownloadStorageData is correct, because it doesn't use RangedFileResponse (just search in our codes mentions of this class and you will see how it should be implemented). It's not your fault, sure, however if we switch to RangedFileResponse usage, GET request will contain more headers like 'byte-range' and my question 'how does django generate HEAD responses' becomes more interesting..

triklozoid: @makseq we don't return file body from this API handler, so we don't need RangedFileResponse
I checked, for me partial loading works fine.

makseq (approved):

PR #3549, opened 2023-01-11
fix: DEV-4185: Switch level of next task log messages

triklozoid:


KonstantinKorotaev (approved):

label_studio/tasks/models.py
```diff
@@ -129,6 +129,8 @@ def get_locked_by(cls, user, project=None, tasks=None):
 
     def has_lock(self, user=None):
         """Check whether current task has been locked by some user"""
+        from projects.functions.next_task import get_next_task_logging_level
```
wesleylima, line 132: Is there an advantage to importing `get_next_task_logging_level` in this function block and again on [173](https://github.com/heartexlabs/label-studio/pull/3549/files#diff-278a5eaed0e213875745abafc04f21d5564f53bbcf2160a6469d1ea0e3526e24R173) instead of at the top of the file?

triklozoid: It's a workaround for circular import problem. No other advantages.

hakan458 (approved):

PR #3553, opened 2023-01-12
fix: LSDV-4681: Relax and update pydantic version

makseq:


makseq:
https://github.com/heartexlabs/label-studio-sdk/issues/98

PR #3560, opened 2023-01-14
docs: DOC-109: Add instructions and images for the Frontend Settings tab

anithamuthu2022:


makseq (changes requested):
image

I believe we should mention all these settings using text, because right now they are not indexed. Also we should use bold text for them or h5 headers, because otherwise they will be indexed poorly

anithamuthu2022:
> image
> 
> I believe we should mention all these settings using text, because right now they are not indexed. Also we should use bold text for them or h5 headers, because otherwise they will be indexed poorly

@makseq this task is completed for your review.

PR #3575, opened 2023-01-19
fix: DEV-4228: Critical overlap issue with task locks and skipped annotations

makseq:


label_studio/tasks/models.py
```diff
@@ -129,13 +129,23 @@ def get_locked_by(cls, user, project=None, tasks=None):
     def has_lock(self, user=None):
         """Check whether current task has been locked by some user"""
         from projects.functions.next_task import get_next_task_logging_level
+        SkipQueue = self.project.SkipQueue
+
+        if self.project.skip_queue == SkipQueue.REQUEUE_FOR_ME:
```
niklub, line 135: this block should be enhanced with logging

label_studio/projects/functions/next_task.py
```diff
@@ -272,7 +273,8 @@ def get_next_task(user, prepared_tasks, project, dm_queue, assigned_flag=None):
                 count = next_task.annotations.filter(was_cancelled=False).count()
                 task_overlap_reached = count >= next_task.overlap
                 global_overlap_reached = count >= project.maximum_annotations
-                if next_task.is_labeled or task_overlap_reached or global_overlap_reached:
+                locks = next_task.locks.count() > project.maximum_annotations - next_task.annotations.count()
```
niklub, line 279: this whole block introduces some additional queries - let's move it under the feature flag `fflag_fix_back_dev_4185_next_task_additional_logging_long` and enable only on demand

niklub (changes requested):
We need pytest coverage here

label_studio/projects/functions/next_task.py
```diff
@@ -23,7 +23,7 @@ def get_next_task_logging_level():
 def _get_random_unlocked(task_query, user, upper_limit=None):
     for task in task_query.order_by('?').only('id')[:settings.RANDOM_NEXT_TASK_SAMPLE_SIZE]:
         try:
-            task = Task.objects.select_for_update(skip_locked=True).get(pk=task.id)
```
triklozoid: Why do we need this change?

PR #3642, opened 2023-02-02
feat: LSDV-3936: Task history is stored on the backend

triklozoid:


label_studio/projects/functions/stream_history.py
```diff
@@ -0,0 +1,27 @@
...
+            else:
+                task_ids = set([h['taskId'] for h in history.data])
+                if next_task.id not in task_ids:
+                    history.data.append(new_history_data)
+                if len(task_ids) + 1 > settings.LABEL_STREAM_HISTORY_LIMIT:
+                    history.data = history.data[-settings.LABEL_STREAM_HISTORY_LIMIT:]
+            history.save()
+
+
+def fill_history_annotation(user, task, annotation):
```
makseq, line 27: do we need fill_history_draft? similar functionality, but for drafts

triklozoid: @makseq can one annotator have several drafts for the same task? If not, so we don't need to select some particular one.

makseq: if he is an admin, he can - via quickview

triklozoid: @makseq Now user actions in quickview don't affect label stream history, if we want to change it, I think we need to collect product requirements first. I'm solving only the problem "history disappears after page refresh" described in the task.

makseq: I guess the created drafts via quickview can create an issue when you move through history and don't specify which id LSF should use to load. E.g.:
1 user goes to label stream and postpones draft A
2 user creates draft B for task 1 in quickview
3 user goes back in task history and sees draft B instead of A

PR #3691, opened 2023-02-13
feat: LSDV-4583-5 Add valueList validation

triklozoid:


label_studio/core/utils/schema/label_config_schema.json
```diff
@@ -17,6 +17,13 @@
         {"type": "boolean"},
         {"type": "number"}
       ]},
+    "@valueList": {
+      "anyOf": [
+      {
+        "type": "array",
```
hlomzik: It should be string as well. It doesn't contain arrays (and actually can't as a property?), but only points to task data, the same as `value`

hlomzik (changes requested):

label_studio/core/utils/schema/label_config_schema.json
triklozoid: Fixed.

hlomzik (approved):

discardedhunk (approved):

PR #3695, opened 2023-02-14
fix: LSDV-3074: Ignore feature flag file in git

makseq:
If you want to include feature_flags.json, use 
```
git add -f feature_flags.json
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,5 +77,5 @@ docker-compose.override.yml
 .bash_history
 
 # if you want to include feature flags - use 
-# git add -f feature_flags.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -74,4 +74,8 @@ docker-compose.override.yml
 .config/*
 .python_history
 .viminfo
-.bash_history
\ No newline at end of file
+.bash_history
+
+# if you want to include feature flags - use 
+# git add -f feature_flags.json
```
farioas: `git add -f label_studio/feature_flags.json`

PR #3725, opened 2023-02-22
fix: LSDV-3854-1: Add extension check

triklozoid:


label_studio/core/settings/base.py
```diff
@@ -400,6 +400,31 @@
 UPLOAD_DIR = 'upload'
 AVATAR_PATH = 'avatars'
 
+SUPPORTED_EXTENSIONS = set(
```
farioas, line 403: There are no video formats in this list.
Here's Chrome supported formats: https://sites.google.com/a/chromium.org/dev/audio-video

triklozoid: Added video extensions

makseq (changes requested):
Let's add ff

label_studio/data_import/uploader.py
```diff
@@ -58,6 +58,12 @@ def check_tasks_max_file_size(value):
         raise ValidationError(f'Maximum total size of all files is {settings.TASKS_MAX_FILE_SIZE} bytes, '
                               f'current size is {total} bytes')
 
+def check_extensions(files):
+    for filename, file_obj in files.items():
+        _, ext = os.path.splitext(file_obj.name)
```
makseq, line 63: If file doesnt have extension, it fails.

label_studio/tests/data_import.tavern.yml
```diff
@@ -733,3 +733,39 @@ stages:
...
+    files:
+      file: tests/test_suites/samples/test.wrong_ext
+    headers:
+      content-type: multipart/form-data
+    method: POST
+    url: '{django_live_url}/api/projects/{project_pk}/import?commit_to_project=false'
+  response:
+    json:
+      validation_errors:
+        non_field_errors:
```
makseq, line 768: Let's add file without extension in test too

makseq (changes requested):

label_studio/tests/data_import.tavern.yml
triklozoid: Added.

label_studio/data_import/uploader.py
triklozoid: Can you provide an example? 
For me is works ok: 
```
In [1]: import os                                                                         

In [2]: filename = 'test'                                                                 

In [3]: os.path.splitext(filename)                                                        
Out[3]: ('test', '')
```

makseq (approved):

PR #3787, opened 2023-03-07
feat: DEV-4096: Multi-user labelling and allow users to delete their own annotations

PC91:


label_studio/data_manager/actions/basic.py
```diff
@@ -70,6 +70,40 @@ def delete_tasks(project, queryset, **kwargs):
     return {'processed_items': count, 'reload': reload,
             'detail': 'Deleted ' + str(count) + ' tasks'}
 
+def delete_my_tasks_annotations(project, queryset, **kwargs):
+    """ Delete annotations of connected user by tasks ids
+
+    :param project: project instance
+    :param queryset: filtered tasks db queryset
+    """
+    request = kwargs['request']
```
makseq: It's better to avoid code duplication, if you want to reuse this code just make it as external function above and use in these two places.

makseq (changes requested):

label_studio/data_manager/actions/basic.py
PC91: Thank you @makseq! I changed the code

PR #3836, opened 2023-03-14
fix: LSDV-4536: Fix numbers ordering in json

triklozoid:


label_studio/tests/data_manager/filters/float_tasks.json
```diff
@@ -0,0 +1,5 @@
+[
+ {"a": "456", "b": 789, "c": 4.56, "d": "string"},
+ {"a": "123", "b": 123, "c": 12.3, "d": "string"},
+ {"a": "789", "b": 456, "c": -7.89, "d": "string"}
+]
```
makseq, line 5: Let add test where data field is dict and list. And user selected number in column type. It should not fail with 500

makseq (changes requested):

wesleylima (approved):

AndrejOros (approved):

makseq (approved):

PR #3909, opened 2023-03-27
fix: LSDV-4862: Clarify Feature Flags usage in our products

makseq:


label_studio/core/views.py
```diff
@@ -226,4 +226,14 @@ def feature_flags(request):
     user = request.user
     if not user.is_authenticated:
         return HttpResponseForbidden()
-    return HttpResponse(json.dumps(all_flags(request.user), indent=4), status=200)
+
+    flags = all_flags(request.user)
+    flags['SYSTEM'] = {
```
hlomzik: I'd stick with format already used in this dict: `$default`, `$cloud`, etc... Or maybe just move values in this format into `$system` key.

hlomzik (changes requested):

hlomzik (approved):

PR #3943, opened 2023-03-31
fix: LSDV-4700: Add remove_user from org method

triklozoid:


label_studio/organizations/models.py
```diff
@@ -105,6 +105,12 @@ def add_user(self, user):
             om.save()
 
             return om    
+
+    def remove_user(self, user):
+        OrganizationMember.objects.filter(user=user, organization=self).delete()
```
makseq, line 110: will LseOrganizationMember be removed?

makseq (changes requested):

label_studio/organizations/models.py
triklozoid: Yes, it's one_to_one with cascade.

makseq (approved):

vladimirheartex (approved):

PR #3993, opened 2023-04-06
fix: LSDV-4821: Check labels in draft for config validation

triklozoid:


label_studio/tasks/models.py
```diff
@@ -577,6 +577,13 @@ def has_permission(self, user):
         user.project = self.task.project  # link for activity log
         return self.task.project.has_permission(user)
 
+    def save(self, *args, **kwargs):
```
wesleylima, line 592: Maybe this is a wider refactor since we'd have to do this for non-draft labels... but maybe we can make `update_created_labels_drafts` async?

wesleylima (approved):

label_studio/tasks/models.py
triklozoid: It's a good idea, but we need some infrastructure for it, we have two main problems now: 
* updates from 1 project can fill the queue and slow down the whole system
* race conditions and problems with data consistency

vladimirheartex (approved):

PR #4059, opened 2023-04-18
feat: LSDV-4974: Add status and debug info to storages

makseq:


label_studio/io_storages/base_models.py
```diff
@@ -1,53 +1,235 @@
...
+        self.meta['time_last_ping'] = str(self.time_in_progress)
+
+        self.save(update_fields=['status', 'meta'])
+
+    def info_set_completed(self, last_sync_count, **kwargs):
+        self.status = self.Status.COMPLETED
+        self.last_sync = timezone.now()
+        self.last_sync_count = last_sync_count
+
+        time_completed = datetime.now()
```
triklozoid: Why we don't use `timezone.now()` for all cases?

label_studio/io_storages/base_models.py
```diff
@@ -221,17 +410,28 @@ def sync(self):
...
-                                    meta=meta)
-                self.last_sync_job = job.id
-                self.save()
-                # job_id = sync_background.delay()  # TODO: @niklub: check this fix
-                logger.info(f'Storage sync background job {job.id} for storage {self} has been started')
+            if (
+                    not is_job_in_queue(queue, "import_sync_background", meta=meta) and
+                    not is_job_on_worker(job_id=self.last_sync_job, queue_name='low')
+            ):
+                sync_job = queue.enqueue(
```
triklozoid, line 430: Can we do it with our general `start_job_async_or_sync` function? I like the idea of having this `if redis_connected` check only in one place in the project.

triklozoid (approved):

hakan458 (approved):

label_studio/io_storages/base_models.py
makseq: i would use datetime.now() everywhere, because server time shouldn't depend on time zone. So, it's better to change last_sync to datetime rather than time_completed...

label_studio/io_storages/base_models.py
makseq: it was done not by me, but I think it's because of 
```
 if (
                    not is_job_in_queue(queue, "import_sync_background", meta=meta) and
                    not is_job_on_worker(job_id=self.last_sync_job, queue_name='low')
            ):
```
in start_job_async_or_sync we don't have this check..

label_studio/io_storages/base_models.py
triklozoid: For one server you can do this, but for geo-distributed server configuration it's not the best approach, because you can have simultaneously running servers in different timezones, and if you take just local time, your time_completed calculation will be wrong.

AndrejOros (approved):

PR #4106, opened 2023-04-27
feat: LSDV-4972: Optimize the requests for presigning urls

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [X] Backend (API)
- [X] Frontend

The introduction of the presigned url proxy approach allows for consistent, secure and transparent handling of all task data so the client can safely use the urls embedded in the JSON response without it expiring. This works great for most of the requests made from the application, except for image requests. Image requests are not able to properly cache the response and therefore will always re-request the presigned url endpoint which burdens the server.

#### What does this fix?
Here we are intercepting the requests made to presign, with a service worker, and cache only the redirect response (not the final response body!) so that the browser can effectively use its own heuristics on the resources presigned url, caching the actual response either in disk/memory cache. This reduces the amounts of requests that hit the server by an order of magnitude, while ensuring the requests are still valid and not being made to an expired presigned url. Because of the way in which Media elements request data, we are skipping Range requests here as there is no need to cache these redirects since it will be making the subsequent requests to the presigned url directly, and not the internal proxy redirect url, and has handling already for recovering from and refreshing presigned urls through the proxy.

None.

Improves performance by removing a pile of redundant requests and making the natural browser cache effective once more.

This is more secure as the only aspect of which is cached is the presigned url which the request would be redirected to, after it is checked fully against the user on the server. This means the url which is user facing is always valid, and never stale, but remains secure at all times.

- Cache the signing in a server side cache. This is non optimal as the cache hits for this application will almost never be cross-user, and would be harder to implement and maintain. Also the effectiveness of this cache is far reduced, as we would still have to check the task/user and project storage anyways for permissions. This is all not necessary when caching the redirect url client-side.

- `fflag_fix_all_lsdv_4711_cors_errors_accessing_task_data_short`

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Presigned Urls.


label_studio/core/static/js/sw.js
```diff
@@ -0,0 +1,169 @@
...
+  event.respondWith(
+    // Cache first approach, look for the cached response and check if it's still valid
+    caches
+      .match(event.request)
+      .then(async (cachedResponse) => {
+        if (cachedResponse) {
+          // Check if the expiration timestamp has passed
+          const { expirationTimestamp } = parseCacheHeaders(cachedResponse);
+
+          const now = Date.now();
```
hlomzik: not used later, so maybe just move it to check?

hlomzik (approved):

label_studio/core/static/js/sw.js
bmartel: Ya that would be better, it was initially reused but since many iterations later it looks to be standing alone once more 👍

vladimirheartex (approved):

PR #4189, opened 2023-05-15
fix: LSDV-4902: GCS pre-signed urls are broken in LS since v1.7.2 and in LSE

makseq:


label_studio/io_storages/gcs/form_layout.yml
```diff
@@ -15,7 +15,7 @@ gcs_credentials: &gcs_credentials
   - type: password
     name: google_application_credentials
     label: Google Application Credentials
-    placeholder: Copy and paste here the content of credentials.json
+    placeholder: Paste here the content of credentials.json OR leave blank to use ADC
```
wesleylima: This might be clearer phrasing:
"Paste the contents of credentials.json here OR leave blank to use ADC"

wesleylima (approved):

MihajloHoma (approved):

PR #4243, opened 2023-05-25
fix: LSDV-5163: Storage errors AttributeError can't set attribute 'storage' when pre-signed proxy urls are turned on

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [X] Backend (API)
- [ ] Frontend

There were some errors being reported for some usages when the pre-signed proxy urls were enabled, and the point of which this appears to have been failing was on the setting of the storage property of the task. So instead of doing that, resolve without setting the property, and wrap the resolving uri function to catch any additional unexpected errors possibly thrown from the storage client itself, log the error and handle it gracefully to avoid 500's.

#### What does this fix?
An error which became present when enabled `AttributeError can't set attribute 'storage' `, which threw 500 errors for around 300 requests in about 1 minute time, before disabling.

No.

No.

None.

`fflag_fix_all_lsdv_4711_cors_errors_accessing_task_data_short`

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [X] integration
- [ ] unit

Import Storages, Presigned Storage Data


niklub (approved):

bmartel:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-lsdv-5163 (477a9cc0b01240d485c279aec1b4b2d035e08f93)
>
> [Workflow run](https://github.com/heartexlabs/label-studio/actions/runs/5082719626)

makseq (approved):

label_studio/data_import/api.py
```diff
@@ -604,19 +604,23 @@ def get(self, request, *args, **kwargs):
...
 
-        if resolved is None or resolved['url'] is None:
+        if resolved is None or resolved.get('url') is None:
             return Response(status=status.HTTP_404_NOT_FOUND)
 
-        url = resolved['url']
-        maxAge = 0
-        if resolved['presign_ttl']:
-            maxAge = resolved['presign_ttl'] * 60
+        url = resolved.get('url')
```
triklozoid: I'm not sure we need `.get` here. We already had this check before, but if we will remove it in the future, I think it's a bit clearer exception, than redirect to None, that we'll get in this case.

triklozoid (approved):

label_studio/data_import/api.py
bmartel: true, I'll put that back to align better with the control flows. Thanks!

PR #4372, opened 2023-06-13
fix: Use class name in validation message

triklozoid:


label_studio/tasks/validation.py
```diff
@@ -120,7 +120,7 @@ def check_allowed(task):
     @staticmethod
     def raise_if_wrong_class(task, key, class_def):
         if key in task and not isinstance(task[key], class_def):
-            raise ValidationError('Task[{key}] must be {class_def}'.format(key=key, class_def=class_def))
+            raise ValidationError('Task[{key}] must be {class_def.__name__}'.format(key=key, class_def=class_def))
```
makseq: what if class_def = (list, dict) ? I think it will print 'tuple', that's incorrect

makseq (changes requested):

label_studio/tasks/validation.py
triklozoid: Fixed.

makseq (approved):

PR #4394, opened 2023-06-16
fix: LSDV-5289: Prevent db deadlocks

triklozoid:


makseq (approved):

label_studio/projects/models.py
```diff
@@ -412,7 +412,8 @@ def _rearrange_overlap_cohort(self):
         logger.info(f"Required tasks {must_tasks} and left required tasks {left_must_tasks}")
         if left_must_tasks > 0:
             # if there are unfinished tasks update tasks with count(annotations) >= overlap
-            tasks_with_max_annotations.update(overlap=max_annotations, is_labeled=True)
+            ids = tasks_with_max_annotations.values_list('id', flat=True)
+            all_project_tasks.filter(id__in=ids).update(overlap=max_annotations, is_labeled=True)
```
makseq, line 416: why all project tasks?

makseq (changes requested):

label_studio/projects/models.py
triklozoid: It's the same subset of tasks, I just make the first step with loading ids and make an update with a simple condition, filter by list of ids.

makseq (approved):

wesleylima (approved):

PR #4467, opened 2023-07-03
fix: LSDV-5289: Decrease lockiness of task update code

jombooth:

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

Lengthy deadlocks (1+hr) from this code have caused some tasks on rqworker to hang for more than an hour. We theorize this is because the existing implementation updates all task counts in a loop wrapped in `transaction.atomic()`, then updates the same tasks' `is_labeled` fields in another bulk update loop, also wrapped in a different `transaction.atomic()`.

In addition to each loop wrapped in `atomic` resulting in a lock on each row that it updates that lasts for the duration of the entire loop, the atomicity guarantee we really want (and which the present implementation does not provide) is that if a task's counts are updated, that task's `is_labeled` field is updated; a failure in the latter step should result in a rollback of the former.

This new implementation should hold locks for much less time (since, after each batch of rows has had its counts updated and `is_labeled` values updated, it can be fully released from all locks), and also give us the atomicity guarantee that we're seeking.

Yes. This is expected to reduce database deadlocks originating from rq_worker updating tasks' counts and `is_labeled` fields.

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -847,17 +847,25 @@ class Project(ProjectMixin, models.Model):
             bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
         return len(objs)
 
-    def _update_tasks_counters_and_is_labeled(self, queryset, from_scratch=True):
+    def _update_tasks_counters_and_is_labeled(self, task_ids, from_scratch=True):
         """
-        Update tasks counters and is_labeled in a single operation
-        :param queryset: Tasks to update queryset
+        Update tasks counters and is_labeled in batches of size settings.BATCH_SIZE.
+        :param task_ids: List of task ids to be updated
         :param from_scratch: Skip calculated tasks
         :return: Count of updated tasks
         """
-        queryset = make_queryset_from_iterable(queryset)
-        objs = self._update_tasks_counters(queryset, from_scratch)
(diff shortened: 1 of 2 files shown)


label_studio/projects/models.py
```diff
@@ -847,17 +847,25 @@ def _update_tasks_counters(self, queryset, from_scratch=True):
...
         :return: Count of updated tasks
         """
-        queryset = make_queryset_from_iterable(queryset)
-        objs = self._update_tasks_counters(queryset, from_scratch)
-        bulk_update_stats_project_tasks(queryset, self)
-        return objs
+        num_tasks_updated = 0
+        page_idx = 0
+
+        while (task_ids_slice := task_ids[page_idx * settings.BATCH_SIZE:(page_idx + 1) * settings.BATCH_SIZE]):
```
jombooth, line 860: Flagging that this usage of an assignment expression (aka walrus operator) requires Python 3.8. Python 3.7 recently reached EOL; do we still need to support it?

jombooth: per @farioas we are fine here, because support for 3.7 was dropped several weeks ago.

niklub (approved):

triklozoid (approved):

PR #4483, opened 2023-07-07
fix: LSDV-5348: More robust and uniform SSRF defenses

jombooth:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

While reviewing a CodeQL finding (https://github.com/heartexlabs/label-studio/security/code-scanning/586) which turned out to be a false positive, a few shortcomings in the SSRF defense code were identified:

- The previous `url_is_local` implementation would throw an exception in the UI when a URL value like `asdf` was passed. Note that it was only called when SSRF_PROTECTION_ENABLED is True.

image

After this change, the result is

image

- A separate check, outside of the `url_is_local` function, tested to see whether the URL began with "file://" in the uploader. This check would have been possible to bypass by using "FILE://" as the scheme, but luckily would have led to the above (unintentional) exception from `url_is_local` if SSRF protection was enabled via `settings.SSRF_PROTECTION_ENABLED`. Additionally, `requests.get`, which we have been using to fetch data since `fflag_fix_back_lsdv_4568_import_csv_links_03032023_short` was enabled, throws an exception for a `file://` scheme by default.
- The previous URL validation approach was inconsistent between the uploader code and the ML Backend Serializer, which had no separate check for any non-http[s] schemes. Now, we always allowlist only "http" and "https" schemes when accepting URLs from clients.
- `urllib.parse.urlparse` was previously used in the implementation of `url_is_local`; `requests` by contrast uses `urrlib3` to parse URLs. Per https://claroty.com/wp-content/uploads/2022/01/Exploiting-URL-Parsing-Confusion.pdf, to minimize risk of exploits stemming from parser inconsistencies, it's best to minimize the number of URL parsers used in any application.
- Testing for SSRF defenses was previously anemic; this PR adds extensive unit tests for the validator function.

#### What does this fix?
See screenshots above; this definitely fixes a low-level exception reaching the UI if SSRF_PROTECTION_ENABLED, and may also help protect LS against other exploits.

`urllib3` was exact pinned to 1.26.16, which was the version `pip show` revealed was already installed in the LS app container.
This is because `urllib3.util.parse_url` is now imported in core/utils/io.py

This change should make our SSRF defenses more robust, and adds testing.

`fflag_fix_back_lsdv_4568_import_csv_links_03032023_short` was removed by this change. AFAIK this feature flag is enabled everywhere.

_(check only one)_
- [ ] No

A feature flag has been removed.

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit

URL data uploader, ML Backend API

### DIFF
--- a/label_studio/feature_flags.json
+++ b/label_studio/feature_flags.json
@@ -3336,33 +3336,6 @@
       "version": 2,
       "deleted": false
     },
-    "fflag_fix_back_lsdv_4568_import_csv_links_03032023_short": {
-      "key": "fflag_fix_back_lsdv_4568_import_csv_links_03032023_short",
-      "on": true,
-      "prerequisites": [],
-      "targets": [],
-      "contextTargets": [],
-      "rules": [],
-      "fallthrough": {
-        "variation": 0
-      },
-      "offVariation": 1,
-      "variations": [
-        true,
-        false
-      ],
-      "clientSideAvailability": {
-        "usingMobileKey": false,
-        "usingEnvironmentId": false
-      },
-      "clientSide": false,
-      "salt": "4d6e875af4b3453a827a94db5eb8a88d",
-      "trackEvents": false,
-      "trackEventsFallthrough": false,
-      "debugEventsUntilDate": null,
-      "version": 3,
-      "deleted": false
-    },
     "fflag_fix_back_lsdv_4648_annotator_filter_29052023_short": {
       "key": "fflag_fix_back_lsdv_4648_annotator_filter_29052023_short",
       "on": false,
--- a/label_studio/tests/test_core.py
+++ b/label_studio/tests/test_core.py
@@ -4,6 +4,10 @@ import pytest
 import types
 from core.utils.common import int_from_request
 from core.utils.params import bool_from_request
+from core.utils.io import validate_upload_url
+from core.utils.exceptions import InvalidUploadUrlError
+from core.utils.exceptions import LabelStudioAPIException
+
 from rest_framework.exceptions import ValidationError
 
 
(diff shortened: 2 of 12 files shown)


jombooth:
I see there's some import-related unit test failures; will fix this tomorrow.

question for reviewers: would changing this default value to True be an option?

https://github.com/heartexlabs/label-studio/blob/1b9963f8a3b39ec4efb08d9b665fb78ccba86737/label_studio/core/settings/base.py#L398

wesleylima (approved):

label_studio/tests/test_core.py
```diff
@@ -134,3 +138,51 @@ def test_start_browser():
...
+    ('https://172.31.255.255', True, InvalidUploadUrlError),
+    # Start and end of 192.168.0.0/16
+    ('https://192.168.0.0', True, InvalidUploadUrlError),
+    ('https://192.168.255.255', True, InvalidUploadUrlError),
+    # Valid external IPs
+    ('https://4.4.4.4', True, None),
+    ('https://8.8.8.8', True, None),
+    ('http://8.8.8.8', False, None),
+    # Valid external websites
+    ('https://example.org', True, None),
```
jombooth, line 169: Flagging for reviewers that the way this test currently works, these hosts will actually be resolved to a real IP address in test. I think an optimal solution would be mocking the implementation of `socket.gethostbyname` in this test, but I could also see an argument for leaving as is.

label_studio/feature_flags.json
```diff
@@ -3336,33 +3336,6 @@
       "version": 2,
       "deleted": false
     },
-    "fflag_fix_back_lsdv_4568_import_csv_links_03032023_short": {
```
dredivaris, line 3339: I'm not sure about this but I was told the contents of feature_flags.json is overridden automatically.  So I don't know if this code removal will have any effect, might want to double check

label_studio/tests/utils.py
```diff
@@ -54,14 +54,15 @@
...
-    with requests_mock.Mocker(real_http=True) as m:
-        url='https://data.heartextest.net'
+    with mock.patch('data_import.uploader.validate_upload_url'):
+        with requests_mock.Mocker(real_http=True) as m:
+            url='https://data.heartextest.net'
 
-        with open('./tests/test_suites/samples/test_1.csv', 'rb') as f:
-            matcher = re.compile('data.heartextest.net/test_1.csv')
+            with open('./tests/test_suites/samples/test_1.csv', 'rb') as f:
+                matcher = re.compile('data.heartextest.net/test_1.csv')
```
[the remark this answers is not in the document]
dredivaris: yeah this is should probably be `'data\.heartextest\.net/test_1\.csv'`

dredivaris (approved):
LGTM good work!

label_studio/feature_flags.json
jombooth: The git blame makes it look like most updates to this file are automatic, but the flag isn't used after this diff and I did verify that the FF is enabled everywhere, so I think it'll be okay to delete this even if it's not strictly necessary.

label_studio/tests/test_core.py
jombooth: Going to merge as is to avoid further expanding the scope of this diff, but I'll keep an eye on whether we get flakes due to (for example) DNS issues.

label_studio/feature_flags.json
dredivaris: Yeah I don't think removing this is high risk given that its enabled everywhere and may be overwritten anyways.  I just wanted to note that I was not sure this would have any effect on this file.

PR #4511, opened 2023-07-12
fix: LSDV-5289: add FF to toggle long atomic transactions for perf/deadlocks reasons

jombooth:
\### PR fulfills these requirements

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

Bulk update calls in transactions are likely untenable for performance/deadlock reasons, and it's not clear we were getting much utility from the transactions in any case. Per https://www.postgresql.org/docs/15/explicit-locking.html#LOCKING-DEADLOCKS "it is a bad idea for applications to hold transactions open for long periods of time", and for example, `update_task_stats` will make one or more sql queries for each task under certain conditions, while every row updated by `_update_tasks_counters` is still held by a row-level lock.

#### What does this fix?
_(if this is a bug fix)_

Uses a feature flag (`fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short` - which defaults to on) to govern whether certain bulk updates will run inside transactions.

Also simplify a query and fix a broken test.

Once `fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short` is set to false, under some conditions, bulk update loops will throw an exception and the changes will not be rolled back, as they previously were.

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Database performance

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -874,7 +874,7 @@ class Project(ProjectMixin, models.Model):
                 queryset = make_queryset_from_iterable(task_ids_slice)
                 num_tasks_updated += self._update_tasks_counters(queryset, from_scratch)
                 bulk_update_stats_project_tasks(queryset, self)
-                page_idx += 1
+            page_idx += 1
         return num_tasks_updated
 
     def _update_tasks_counters_and_task_states(self, queryset, maximum_annotations_changed,
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -845,8 +845,8 @@ class Project(ProjectMixin, models.Model):
             objs.append(task)
         with conditional_atomic(
             predicate=flag_set,
-            predicate_args=['fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short']
-            predicate_kwargs={'user': self.organization.created_by}
+            predicate_args=['fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short'],
+            predicate_kwargs={'user': self.organization.created_by},
         ):
             bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
         return len(objs)
@@ -865,8 +865,8 @@ class Project(ProjectMixin, models.Model):
         while (task_ids_slice := task_ids[page_idx * settings.BATCH_SIZE:(page_idx + 1) * settings.BATCH_SIZE]):
             with conditional_atomic(
                 predicate=flag_set,
-                predicate_args=['fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short']
-                predicate_kwargs={'user': organization_created_by}
+                predicate_args=['fflag_fix_back_lsdv_5289_run_bulk_updates_in_transactions_short'],
+                predicate_kwargs={'user': organization_created_by},
             ):
                 # If counters are updated, is_labeled must be updated as well. Hence, if either fails, we
                 # will roll back. NB: as part of LSDV-5289, we are considering eliminating this transaction
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -14,7 +14,13 @@ from annoying.fields import AutoOneToOneField
 from data_manager.managers import TaskQuerySet
 from projects.functions.utils import make_queryset_from_iterable
 from tasks.models import Task, Prediction, Annotation, AnnotationDraft, Q_task_finished_annotations, bulk_update_stats_project_tasks
-from core.utils.common import create_hash, get_attr_or_item, load_func, merge_labels_counters
+from core.utils.common import (
+    conditional_atomic,
+    create_hash,
+    get_attr_or_item,
+    load_func,
+    merge_labels_counters,
+)
 from core.utils.exceptions import LabelStudioValidationErrorSentryIgnored
 from core.label_config import (
     validate_label_config,
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -843,8 +843,7 @@ class Project(ProjectMixin, models.Model):
             task.total_predictions = task.new_total_predictions
             objs.append(task)
 
-        with transaction.atomic():
-            bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
+        bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
         return len(objs)
 
     def _update_tasks_counters_and_is_labeled(self, task_ids, from_scratch=True):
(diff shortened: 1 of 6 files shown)


label_studio/tests/test_block_objects.py
```diff
@@ -22,6 +20,6 @@ def test_export(
     assert Task.objects.filter(is_labeled=True).count() == 0
 
 
-def worker_change_stats(time_seconds, tasks):
+def worker_change_stats(tasks):
```
jombooth, line 26: Previously this thread was passed the wrong arguments, and would never actually start. When it does, because of sqlite's locking behavior, sometimes it flakes out: https://github.com/heartexlabs/label-studio/actions/runs/5537143705/jobs/10105651923?pr=4511

I'm not sure that testing this code with two threads making the same call is optimal, in any case; in addition, the sleep() calls make this test run very slowly.

label_studio/projects/models.py
```diff
@@ -843,8 +843,7 @@ def _update_tasks_counters(self, queryset, from_scratch=True):
             task.total_predictions = task.new_total_predictions
             objs.append(task)
 
-        with transaction.atomic():
-            bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
+        bulk_update(objs, update_fields=['total_annotations', 'cancelled_annotations', 'total_predictions'], batch_size=settings.BATCH_SIZE)
```
dredivaris: I'm not suggesting we do this as I don't fully understand the implications, but I noticed we'e running the bulk update in batches in `bulk_update`.  I wonder if rollback on transaction failure _is_ is sufficiently important, maybe a compromise would be to run each batch as an atomic operation and that unit could then be rolled back if need be 🤔

label_studio/tasks/models.py
```diff
@@ -876,29 +876,28 @@ def bulk_update_stats_project_tasks(tasks, project=None):
...
+        # update objects without saving if we can't use overlap
+        for task in tasks:
+            update_task_stats(task, save=False)
+        try:
+            # start update query batches
+            bulk_update(tasks, update_fields=['is_labeled'], batch_size=settings.BATCH_SIZE)
+        except OperationalError as exp:
+            logger.error("Operational error while updating tasks: {exc}", exc_info=True)
+            # try to update query batches one more time
+            start_job_async_or_sync(bulk_update, tasks, in_seconds=settings.BATCH_JOB_RETRY_TIMEOUT, update_fields=['is_labeled'], batch_size=settings.BATCH_SIZE)
```
dredivaris: Given that we're removing the transaction, it could be helpful to wrap some or all of this in a generalized exception handler that logs any failures so we can keep abreast of when/where this fails (if ever).  This would help us keep track of consequences of this removal.  We can leave as is and could still receive logs of exceptions but having a singular log message for the block (with corresponding exception) would make tracking this a lot faster.

dredivaris (approved):
LGTM but I added a few optional suggestions

label_studio/tasks/models.py
```diff
@@ -876,29 +876,28 @@ def bulk_update_stats_project_tasks(tasks, project=None):
...
+
+    use_overlap = project._can_use_overlap()
+    maximum_annotations = project.maximum_annotations
+    # update filters if we can use overlap
+    if use_overlap:
+        # finished tasks
+        finished_tasks = tasks.filter(Q(total_annotations__gte=maximum_annotations) |
+                                        Q(total_annotations__gte=1, overlap=1))
+        finished_tasks_ids = finished_tasks.values_list('id', flat=True)
+        tasks.update(is_labeled=Q(id__in=finished_tasks_ids))
```
jombooth: Query that this emits (I wasn't sure that a Q object could be used this way, but it seems that it can):

```
DEBUG    django.db.backends:utils.py:123 (0.000) UPDATE "task" SET "is_labeled" = ("task"."id" IN (SELECT U0."id" FROM "task" U0 WHERE (U0."project_id" = 1 AND (U0."total_annotations" >= 1 OR (U0."overlap" = 1 AND U0."total_annotations" >= 1))))) WHERE "task"."project_id" = 1; args=(1, 1, 1, 1, 1)
```

label_studio/tasks/models.py
jombooth: I had a somewhat related thought about how this is implemented, which is that if a particular batch fails consistently with OperationalErrors for some reason, we'll schedule retries infinitely as there is no retry budget set. I agree that a more generic approach would be well-advised, with logging as you suggest

dredivaris: Right; If this block of code is idempotent, seems we can just retry until it completes

label_studio/projects/models.py
jombooth: My understanding is that one of the primary goals of the transaction approach was to ensure that if the tasks' counters get an update, so should the `is_enabled` fields on the tasks. This is why I took a similar approach in writing https://github.com/heartexlabs/label-studio/pull/4511/files#diff-6ff9748c9ae68f8cccd65f4ee20df06ed0a8b9c32b58da39db3001a7178b1dabL861-L866 - but it seems like this code is still too lock-prone in practice. Under some conditions `bulk_update_stats_project_tasks` will make 2000+ sql queries https://github.com/heartexlabs/label-studio/pull/4511/files#diff-278a5eaed0e213875745abafc04f21d5564f53bbcf2160a6469d1ea0e3526e24R892-R893, while holding a lock on all the rows affected by `_update_tasks_counters`

jombooth: ...but then, there is the possibility in the code as written that instead an OperationalError exception doesn't propagate up to where the atomic context manager can see it, and therefore, despite an error, we still don't roll back the is_enabled changes: https://github.com/heartexlabs/label-studio/pull/4511/files#diff-278a5eaed0e213875745abafc04f21d5564f53bbcf2160a6469d1ea0e3526e24R897

dredivaris: Is this currently synchronous in the UI? Wonder if we can move this to be handled by rq workers where potential retries wouldn't be an issue and it could be done without atomic as it is being done here.  We could then mark `is_enabled` when it is complete.  Just a thought; I haven't looked at the entirety of related code so speaking partially from ignorance.

label_studio/tasks/models.py
jombooth: Agreed but with the caveat that I think we should have a fixed retry budget (ie, we will retry this job no more than 5 times) - if for some reason the query is consistently throwing an exception, we don't want an infinite loop of retries that never succeed

dredivaris: Yes and probably an exponential retry as well (pretty standard)

label_studio/tests/test_block_objects.py
jombooth: Still not really sure what the goal of the thread in this test was, but I set it up such that it should be reliable on both sqlite and other DBs

jombooth:
This PR is ready for a second pass of review. I've added feature flag functionality to it such that we can test the impact of removing these transactions in a more gradual and reversible manner.

label_studio/projects/models.py
jombooth: As far as I can tell, no, this code doesn't get called synchronously anywhere; it's only executed on workers as-is.

triklozoid (approved):

PR #4582, opened 2023-07-31
fix: LSDV-5221: Making sure that subquery for annotation only returns annotation matching its respective task

wesleylima:
… 

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/data_manager/managers.py
+++ b/label_studio/data_manager/managers.py
@@ -192,6 +192,7 @@ def add_result_filter(field_name, _filter, filter_expressions, project):
             Annotation.objects
                 .annotate(json_str=RawSQL('cast(result as text)', ''))
                 .filter(Q(project=project) & Q(json_str__contains=_filter.value))
+                .filter(task=OuterRef('pk'))
                 .values_list('task', flat=True)
         )
     # Predictions: they don't have `project` yet
@@ -436,7 +437,8 @@ def apply_filters(queryset, filters, project, request):
             cast_value(_filter)
             filter_expressions.append(Q(**{field_name: _filter.value}))
 
-    logger.debug(f'Apply filter: {filter_expressions}')
+    # Do not uncomment this. Stringifying filter_expressions evaluates the (sub)queryset.
+    # logger.debug(f'Apply filter: {filter_expressions}')
     if filters.conjunction == ConjunctionEnum.OR:
         result_filter = Q()
         for filter_expression in filter_expressions:
--- a/label_studio/data_manager/managers.py
+++ b/label_studio/data_manager/managers.py
@@ -437,8 +437,12 @@ def apply_filters(queryset, filters, project, request):
             cast_value(_filter)
             filter_expressions.append(Q(**{field_name: _filter.value}))
 
-    # Do not uncomment this. Stringifying filter_expressions evaluates the (sub)queryset.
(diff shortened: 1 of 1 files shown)


label_studio/data_manager/managers.py
```diff
@@ -436,7 +437,8 @@ def apply_filters(queryset, filters, project, request):
             cast_value(_filter)
             filter_expressions.append(Q(**{field_name: _filter.value}))
 
-    logger.debug(f'Apply filter: {filter_expressions}')
+    # Do not uncomment this. Stringifying filter_expressions evaluates the (sub)queryset.
```
dredivaris: can we just delete this?  Why are we leaving this logging statement in here?

label_studio/data_manager/managers.py
```diff
@@ -192,6 +192,7 @@ def add_result_filter(field_name, _filter, filter_expressions, project):
             Annotation.objects
                 .annotate(json_str=RawSQL('cast(result as text)', ''))
                 .filter(Q(project=project) & Q(json_str__contains=_filter.value))
```
dredivaris, line 194: Seems that the only reason you are annotating to cast the JSONField into json_str is to run contains on it.  AFAIK Django supports __contains filter lookup on JSONFields (though it is more targeted than the way contains is used here).  Can you clarify why you're doing it like this?  I assume its because we want to match on either keys or values?

dredivaris (commented):
Just had a few questions

label_studio/data_manager/managers.py
wesleylima: I wanted to leave an explanation in code as to why this was removed and shouldn't be re-added

dredivaris: hmm.. normally I would say remove it but in this case its a bit of a warning, so yeah there's value in leaving it in there

label_studio/data_manager/managers.py
wesleylima: I didn't modify this line and it seems like it's been this way for a while. Regardless, I think there's a twofold reason why this works best this way.
1. Performance: The requirements for this filter are such that actually parsing the JSON is unnecessary. Just doing a string search is likely faster
2. Variable schema: I think the right side of the  `__contains` argument in a JSONField needs to be an object matching the shape of what's in that JSON field. In annotation.results, the shape of this JSON document and the location of where the actual "label" that we're searching for will vary widely depending on the label config

dredivaris: 🤦  oops I was digging into the details of the query and completely overlooked that wasn't your change

dredivaris (approved):
LGTM

label_studio/data_manager/managers.py
bmartel: I would maybe make the comment along the lines of:

```
"""WARNING: Stringifying filter_expressions will evaluate the (sub)queryset.
    Do not use a log in the following manner:
    logger.debug(f'Apply filter: {filter_expressions}')
"""
```

bmartel (approved):

MihajloHoma (approved):

PR #4629, opened 2023-08-14
perf: LSDV-4695: Add Project relation to Prediction model to improve overall performance

dredivaris:

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

Prediction performance in general has not been optimal due to the necessity of querying through the task model whenever needing to filter on project (Prediction -> Task -> Project). 

#### What does this fix?
This PR is the first of a set of 2 PRs.  This PR adds migrations to add the a project foreign key field to Prediction as well as a data migration to populate the new field with the correct project for existing Predictions.  I've also added changes across the code on Prediction saves / bulk saves etc to ensure any new Predictions added after the migration will store the project properly.

Once the first PR is live, the second PR will introduce changes to queries which use the new project relation rather than querying through task -> project.

The PRs were separated in this way because until all Predictions have been updated with the new project fk, the queries could not be optimized to use the fk.

None - and this is by design.  I've introduced code changes here on all Prediction save's/bulk saves to save project_id on create.  This is necessary because the next PR will introduce query updates that depend on this Prediction.project_id linkage and if its missing on any Predictions, the query will break.  I'll use a feature flag on the following PR to test the query changes however.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Predictions


label_studio/tasks/migrations/0042_auto_20230810_2304.py
```diff
@@ -0,0 +1,37 @@
...
+
+def set_project_from_task(apps, schema_editor):
+    Prediction = apps.get_model('tasks', 'Prediction')
+
+    connection.queries_log.clear()
+
+    predictions_with_task = Prediction.objects.all().select_related('task')
+
+    # Break the update into batches using Paginator
+    paginator = Paginator(predictions_with_task, 1000)
```
yyassi-heartex: is there a particular reason we're passing in 1000 here? can we define it as a variable so the code is a little more readable? is there an optimal size for batching?

dredivaris: Yes I can put this in a variable to avoid magic numbers.  Good question regarding optimum batch size 🤔 , in this case 1000 was an arbitrary batch size though I think thats common.  While I imagine there is an optimum, it probably varies on what sort of migration you're doing and even on db resources.  Finding an optimum would probably require testing the migration with different batch sizes, which seems overkill given this will only be run once.  The 2 reasons for batching here are 1. to avoid memory issues with loading all Predictions in memory at once and 2. to avoid locking all the Predictions while a long query is run on them.  For that goal I think 1000 is an adequate number (if I recall correctly we have ~4 million predictions in prod)

yyassi-heartex: cool, thank you for the wonderful explanation - makes sense

yyassi-heartex (approved):

label_studio/data_import/api.py
```diff
@@ -327,7 +327,11 @@ class ImportPredictionsAPI(generics.CreateAPIView):
     def create(self, request, *args, **kwargs):
         # check project permissions
         project = self.get_object()
-        tasks_ids = set(Task.objects.filter(project=project).values_list('id', flat=True))
+
+        tasks_ids_and_project_ids = set(Task.objects.filter(project=project).values_list('id', 'project_id'))
```
bmartel: In the cases of projects which already have 10-20k tasks or more (some I've seen have 100k+ tasks), this will start to blow up the data import process and task creation process. 

I noticed we were already doing this, I guess I wonder if this isn't already a problem we face. Also to this point, won't the project_id be redundant and wasteful to make this grab the additional field which we already have?

```suggestion
        tasks_ids = set(Task.objects.filter(project=project).values_list('id', flat=True))
```

label_studio/data_import/api.py
```diff
@@ -327,7 +327,11 @@ class ImportPredictionsAPI(generics.CreateAPIView):
...
+        tasks_ids = {task_id_and_project_id[0] for task_id_and_project_id in tasks_ids_and_project_ids}
+        task_to_project_mapping = {task_id: project_id for task_id, project_id in tasks_ids_and_project_ids}
```
bmartel:
```suggestion
```

We reasonably could also remove this and just grab the direct project id since we already have it.

label_studio/data_import/api.py
```diff
@@ -337,6 +341,7 @@ def create(self, request, *args, **kwargs):
                     f'from project {project} tasks')
             predictions.append(Prediction(
                 task_id=item['task'],
+                project_id=task_to_project_mapping[item['task']],
```
bmartel:
```suggestion
                project_id=project.id,
```

label_studio/data_import/api.py
```diff
@@ -327,7 +327,11 @@ class ImportPredictionsAPI(generics.CreateAPIView):
     def create(self, request, *args, **kwargs):
         # check project permissions
         project = self.get_object()
```
bmartel, line 329: we already have the project object here

label_studio/data_import/api.py
dredivaris: Seriously, looking at this with a fresh set of eyes, I have no idea why I did it this way.  Looks like I'm filtering tasks on project and redundantly including projects on all of those tasks.  This can be greatly simplified.

label_studio/data_import/api.py
dredivaris: 💯

bmartel (approved):

PR #4706, opened 2023-08-28
docs: DIA-403: Create user docs for Data Discovery

caitlinwheeless:
Data discovery docs for PP. These will not be available in the side bar, will need to view at links:

dataset_search
dataset_manage
dataset_create


docs/source/guide/dataset_create.md
```diff
@@ -0,0 +1,148 @@
...
+tier: enterprise
+order: 0
+order_enterprise: 405
+meta_title: Create a Dataset to use with data discovery in Label Studio Enterprise
+meta_description: How to create a Dataset in Label Studio Enterprise using Google Cloud, Azure, or AWS.
+hide_sidebar: true
+---
+
+!!! note
+    At this time, we only support building datasets from unstructured data, meaning that the data must be in individual files rather than a structured format such as CSV or JSON.
```
pakelley: can we say "a bucket of unstructured data"? I think it could be confusing to imply that our requirement is "unstructured-ness", rather than "the data format we currently support is a filesystem"

docs/source/guide/dataset_create.md
```diff
@@ -0,0 +1,148 @@
...
+
+!!! note
+    To create a new Dataset, your [user role](manage_users#Roles-in-Label-Studio-Enterprise) must have Owner or Administrator permissions. 
+
+## Before you begin
+
+Datasets are retrieved from your cloud storage environment. As such, you will need to provide the appropriate access key to pull data from your cloud environment.
+
+
+## Datasets using Google Cloud Storage
```
pakelley, line 24: I think it likely makes sense to have redundant docs in this case (since Datasets works pretty differently from Projects in some cases), but just to be sure: have you looked at the existing Label Studio docs on setting this stuff up, and are we purposefully duplicating some of that here?

docs/source/guide/dataset_manage.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Dataset settings
+
+From the Datasets page, click the overflow menu next to dataset and select **Settings**.  
+
+![Overflow menu next to a dataset](/guide/images/data_discovery/dataset_settings.png)
+
+From here you can do the following:
+
+- Edit the dataset name and description.
```
pakelley, line 22: Mind linking to the section where all that gets set up? I figure when users are reading this, it's likely they've forgotten how they set everything up

docs/source/guide/dataset_create.md
caitlinwheeless: I think we are repeating information elsewhere - and a future task for me would be to make sure I'm using content reuse when possible. But I'd rather have all the information up front and here where they need it rather than sending them off to other pages in different sections (in this case, the pages on importing data from cloud storage).

pakelley (approved):

caitlinwheeless:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into docs/data-discovery (15c9fb805feab773bd7320f5861aa9439e49bcaa)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6017812305)

PR #4730, opened 2023-09-01
feat: LSDV-5506: Add n+1 query detection to dev environment and pytests

dredivaris:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

This n+1 profiler will add inefficient query detection to pytests as well optionally, the LS when running locally in debug mode.  It will allow us to gain visibility into potential queries that can be improved and includes n+1, lazy and eager load detection.

Note, about 65 n+1 profiler errors were detected out of 509 tests.

nplusone library was added to test requirements

I've implemented a fixture which allows the profiler to be run all tests except those marked with the `pytest.mark.skip_nplusone` marker.  This allows us to ignore tests which fail the profile for the time being but mark them for later improvement.  

There is another option to set `NPLUSONE_RAISE = True` in the Django settings file.  This will cause the n+1 profiler to automatically raise exceptions where improvable queries are found essentially obviating the fixture.  

There are a handful additional queries which are detected with each approach so its unclear that there is a better option, however, the fixture method gives us more granular control of where n+1 detection is enabled so I'm following that approach. 

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [x] unit

Tests and CI


label_studio/core/settings/base.py
```diff
@@ -222,6 +224,11 @@
     'core.current_request.ThreadLocalMiddleware',
 ]
 
+# adds n+1 and eager loading query profiling to logs
+if DEBUG:
```
dredivaris, line 228: This allows n+1 profile to run when the app is run locally so n+1 queries logs are displayed while running the app, not just during tests.  We might consider putting this behind an additional parameter. But unlike LSE, label studio doesn't have any issues with RQ so this can be run in a normal app run.

dredivaris:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-lsdv-5506-1 (2f052aa54a9fbdeef0f39d6bff3d1e44998c1cf7)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6175229066)

label_studio/tests/conftest.py
```diff
@@ -769,3 +770,13 @@ def tick_clock(_, seconds: int = 1) -> None:
     now += timedelta(seconds=seconds)
     freezer = freeze_time(now)
     freezer.start()
+
+
+@pytest.fixture(autouse=True)
+def _raise_nplusone(request):
```
jombooth, line 776: Why is this called `_raise_nplusone`? I do see that there's an nplusone setting NPLUSONE_RAISE that configures whether nplusone will raise an exception when it detects an n+1 query: https://github.com/jmcarp/nplusone#customizing-notifications

Do tests start failing if that setting is enabled?

dredivaris: `_raise_nplusone` raises nplusone issues unless they are marked to skip by `skip_nplusone`

jombooth: Another potential concern: https://github.com/jmcarp/nplusone#generic says that when using the Profiler one has to import the relevant extension for one's ORM. I _think_ in this case that would mean adding `import nplusone.ext.django`.

My major underlying concern is that there aren't that many tests marked `skip_nplusone`, I would have expected many more failures.

dredivaris: Interesting that it didn't fail without the extension, and it seems to detecting nplusone issues.  Its possible that when adding the extension it will catch more issues.  Wish the docs specified the reason for it.  Anyway, I'll test with the extension and see if we gain any added benefits from it.

PR #4731, opened 2023-09-01
fix: LEAP-1: Start migration to monorepo

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

The basis of this solution surrounds the collapsing of our multi-repo frontend code to a mono-repos in LabelStudio.

Orchestration and tooling will be handled by [NX.dev](http://nx.dev/), which allows us to maintain our code as a mono-repo architecture in a much more focused manner and with less boilerplate, configuration juggling and maintenance burden.

#### What does this fix?
We are moving ReactApp into monorepo app `labelstudio`
Adding the build and test commands on package.json
Updating the build to target the new static folder

Here we are migrating the ReactApp to the monorepo structure, and we are starting use NX to run and build the dist files 

we started using [NX](https://nx.dev/) to build and run the project.

As it's just migrating the ReactApp to the monorepo structure I guess it doesn't change the performance

no

All the possible approaches have been discussed and as a team we decided that using NX would be the best one

There are no FF applied in this project, it was discussed with the team as well and it was defined that as it is a migration the best way to proceed in case of problem is roll back to the last working state.

_(check only one)_
- [x] No

It's not covered by feature flags, but as I explain in the last topic, it was discussed with the team and we decided that as it's a migration the best way to proceed in case of problem is roll back to the last working state.

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


yyassi-heartex (commented):
looking good so far! but we do need the makefile to get updated to use `/web` instead of `/label_studio/frontend`

label_studio/frontend/.eslintignore
```diff
@@ -3,4 +3,4 @@ webpack.config.js
 get-build.js
 jest-puppeteer.config.js
 *.html
-release.js
+release.js
```
hlomzik: I think it's better to leave an empty string at the end of the file and we may actually have it in linting rules, they just don't apply to all file types maybe. Please, return it back here and in other places, this will reduce the diff and denoise blame

web/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
```
hlomzik, line 8: yeah, you even have it here! =)

web/.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+      }
+    },
+    {
+      "files": ["*.ts", "*.tsx"],
+      "extends": ["plugin:@nx/typescript"],
+      "rules": {}
+    },
+    {
+      "files": ["*.js", "*.jsx"],
+      "extends": ["plugin:@nx/javascript"],
```
hlomzik: are we changing our linting again? %)
if so, there will be bunch of new excess changes again

hlomzik:
Please, add more text into PR description, better to add before the whole template.
As I see currently this code is just duplicated from "react app" and both versions will sit in repo? Where are we going to remove current app? If you do it in this PR it may actually decrease the diff drastically and only real changes will be visible.

web/.eslintrc.json
juliosgarbi: I removed all the global lint config, now each project will keep having their only lint config, but now in this new structure we can start thinking to use the same lint config to all projects and use just the global config file!

yyassi-heartex (approved):

web/.eslintrc.json
nick-skriabin: we have a separate repo with all the rules, might finally use it globally

nick-skriabin (approved):

PR #4745, opened 2023-09-06
docs: LSDV-5553: Add styling for info/tip admonition

caitlinwheeless:
We currently have several note types - note (gray), warning (orange), attention (red). Added styling for an info note that can be used for tips or ideas:

```
!!! info tip
     You can do XYZ to reduce your time spent doing ABC!
```


docs/themes/v2/source/css/styles.css
```diff
@@ -1085,6 +1086,13 @@ main {
   --color-background: var(--color-yellow-5);
 }
 
+.content-markdown .admonition.info {
+  --color-border: #DD7DFF;
```
huguestennier:
```suggestion
  --color-border: var(--color-purple-90);
```

docs/themes/v2/source/css/styles.css
```diff
@@ -1085,6 +1086,13 @@ main {
...
+  --color-title: var(--color-white);
+  --color-title-background: #DD7DFF;
```
huguestennier:
```suggestion
  --color-title-background: var(--color-purple-90);
```

docs/themes/v2/source/css/styles.css
```diff
@@ -1085,6 +1087,13 @@ main {
...
+  --color-background: #FDF6FF;
```
huguestennier: Could we reuse the Purple (Plum) 0 here? The shade is slightly different `#FBF2FC` but it should work

```suggestion
  --color-background: var(--color-purple-0);
```

caitlinwheeless: Sounds good

huguestennier (approved):

PR #4778, opened 2023-09-13
fix: LEAP-19: Add GCS bucket cache

triklozoid:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/io_storages/gcs/utils.py
```diff
@@ -30,6 +31,21 @@ class ConvertBlobTo(Enum):
         JSON_DICT = 3
         BASE64 = 4
 
+    @classmethod
+    @lru_cache(maxsize=1)
+    def get_bucket(
+        cls,
+        google_project_id: str = None,
+        google_application_credentials: Union[str, dict] = None,
+        bucket_name: str = None,
```
jombooth: All of these types should be `Optional[X]` where `X` is what they currently are, if we really want them to default to None.

label_studio/io_storages/gcs/utils.py
```diff
@@ -30,6 +31,21 @@ class ConvertBlobTo(Enum):
         JSON_DICT = 3
         BASE64 = 4
 
+    @classmethod
+    @lru_cache(maxsize=1)
```
jombooth, line 36: It strikes me that leaving the gcs bucket around with an infinite TTL might lead to unintended behavior - suppose for example that the bucket is deleted in GCS, this function would continue to return the bucket, even though an actual call to `get_client` would no longer return it. An alternative approach would be to use a request-scoped cache for the bucket, which I think would provide an equivalent speedup in the example from the LEAP-19 ticket (since it's a single request trace). See for example: https://pypi.org/project/django-request-cache/ which I think could be useful in other settings as well.

label_studio/io_storages/gcs/utils.py
triklozoid: Fixed.

label_studio/io_storages/gcs/utils.py
triklozoid: I don't want to add 2 third party middlewares for this. 

Cache has size of 1 element, so `unintended behavior` will be quite rare, and even with request based cache it is still possible, bucket can be deleted after we initialized object. 

I can add ttl to this cache, do you think it's safe enough?

jombooth: A TTL would indeed get us most of the way there. I've been thinking for a while that we could benefit from having this package in the build if we don't already: https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache

triklozoid: Added ttl.

label_studio/io_storages/gcs/utils.py
```diff
@@ -38,6 +39,7 @@ def get_bucket(
         google_project_id: Optional[str] = None,
         google_application_credentials: Optional[Union[str, dict]] = None,
         bucket_name: Optional[str] = None,
+        ttl_hash=None,
```
jombooth: Please add a type for this (probably Optional[int])

label_studio/core/utils/common.py
```diff
@@ -744,3 +744,8 @@ def wrapper(*args, **kwargs):
 
 def empty(*args, **kwargs):
     pass
+
+
+def get_ttl_hash(seconds=60):
```
jombooth: Please add types for both `seconds` and the return value

triklozoid: Added.

label_studio/io_storages/gcs/utils.py
triklozoid: Added.

label_studio/io_storages/gcs/utils.py
```diff
@@ -30,6 +32,22 @@ class ConvertBlobTo(Enum):
...
         BASE64 = 4
 
+    @classmethod
+    @lru_cache(maxsize=1)
+    def get_bucket(
+        cls,
+        google_project_id: Optional[str] = None,
+        google_application_credentials: Optional[Union[str, dict]] = None,
+        bucket_name: Optional[str] = None,
+        ttl_hash: Optional[int] = None,
```
jombooth: Please make this a required argument (ie, type `int` and without a default value), users shouldn't call this in a way that would cause buckets to be cached indefinitely

jombooth (approved):
LGTM modulo this small change: https://github.com/HumanSignal/label-studio/pull/4778/files#r1355927021

label_studio/io_storages/gcs/utils.py
triklozoid: fixed

MihajloHoma (approved):

PR #4806, opened 2023-09-20
feat: OPTIC-115: Add is_deleted field to User model

dredivaris:

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

User


label_studio/users/migrations/0007_user_is_deleted.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('users', '0006_user_allow_newsletters'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='is_deleted',
+            field=models.BooleanField(default=False, help_text='Designates whether to treat this user as deleted. Select this instead of deleting accounts.', verbose_name='deleted'),
```
bmartel: Is this supposed to say "Select" or "Unselect" like the field?

label_studio/users/models.py
bmartel, line 1: Are we not adding an index to the `deleted` column?

label_studio/users/migrations/0007_user_is_deleted.py
dredivaris: I meant select as I wrote it as selecting a boolean field would suggest marking it true for deleted.  But maybe select is not the best word here

label_studio/users/models.py
dredivaris: Thanks for catching this!  Yeah, that was the plan, it totally slipped my mind.

label_studio/users/migrations/0007_user_is_deleted.py
dredivaris: Ahh yes the field is wrong then, they should both say select

bmartel (approved):

bmartel:
Code looks fine, can you take a look at why the tests are all failing @dredivaris ?

dredivaris:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-optic-115 (75b4f5f402fb710015a1270db350665d4f174bcb)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6305534954)

label_studio/users/models.py
```diff
@@ -95,7 +95,13 @@ class User(UserMixin, AbstractBaseUser, PermissionsMixin, UserLastActivityMixin)
     is_active = models.BooleanField(
         _('active'),
         default=True,
-        help_text=_('Designates whether to treat this user as active. ' 'Unselect this instead of deleting accounts.'),
+        help_text=_('Designates whether to treat this user as active. Unselect this instead of deleting accounts.'),
+    )
+    is_deleted = models.BooleanField(
+        _('deleted'),
+        default=False,
+        db_index=True,
```
wesleylima, line 103: I'm not sure indexing this field like this will give us any performance gains

Maybe it's worth investigating the usecase more, but I think we would be okay without an index.

wesleylima (approved):

label_studio/users/models.py
dredivaris: Yeah we might be ok without an index but given the user table will filter on this field by default on every user query (is_deleted=False), I think it can't hurt and can only help even if the gain isn't huge - even small gains will add up on less db load. 

To your point, most user queries query on an individual user (generally by id or email) so I'm guessing in those cases, the index wont make a difference.  Also we expect most users will not be in deleted state so this field will be weighted False which is primarily what we're querying on which also would suggest it would not effect performance as much.

I think I was leaning towards using an index simply from the mindset that it doesn't hurt us much to add it (boolean indexes are relatively lightweight) and since its a field we'll always be querying on it feels safer add it.  Also to protect from unexpected performance issues from new user queries (for example if we add a query for is_deleted=True).

If you feel strongly about this I can remove it but for now I'll leave it in.

dredivaris:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-optic-115 (169e00474e1ef9d53fe17f1d8069952efa47f257)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6354257389)

PR #4832, opened 2023-09-26
feat: OPTIC-118: Query updates to support user soft deletion

dredivaris:
This PR adds the following:
- adds a default is_deleted filter in the UserManager which filters out soft deleted users by default (use `with_deleted()` to bypass this filter).  This addition caused some issues with migrations so a parameter `disable_is_deleted_filter` was added to the manager to allow it to be disabled during migrations.  Adding this parameter however had the effect of creating a new migration (I think Django tracks changes to managers in migrations though this doesn't have any actual effect on the database.  I've checked and this migration is only created when I set the `disable_is_deleted_filter=True` in initial user migration [0001_squashed_0009_auto_20210219_1237.py](https://github.com/HumanSignal/label-studio/pull/4832/files#diff-41b3a1e27082ee6c77df7e40fb9d0af80e381bf55d713ef72acd03ca6d79a4b4)] 
- adds an is_deleted column to the admin page as well as an is_deleted search and list filter field
- adds user obfuscation to `get_initials` so deleted users return DU
- adds user obfuscation to the base user serializer so deleted users have the following fields obfuscated: email, first_name, last_name, username

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Users and display of user information


dredivaris:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-optic-118 (a44fdc4866ad011a3f0a7a2ff00de3ce8b0e8716)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6353982988)

label_studio/users/mixins.py
```diff
@@ -7,3 +10,20 @@ def has_permission(self, user):
...
+    @cached_property
+    def get_all_field_names(self):
+        return [field.name for field in self.model._meta.fields]
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if 'is_deleted' in self.get_all_field_names:
+            qs = qs.filter(is_deleted=False)
+        else:
+            qs = qs.filter(user__is_deleted=False)
```
jombooth: Why is `get_all_field_names` needed? Will there be a user model that doesn't have the `is_deleted` field defined on it?

dredivaris: Originally I had anticipated other models with relations to user would benefit from a manager so it was written to handle being added to other models as well.  Since then I determined that wouldn't really be necessarily.  I'll probably be removing this mixin entirely and baking it in to the current user manager.  However, looks the check to self.get_all_field_names is still needed, tests fail without it because this manager is also used in tests and in migrations so during test migration setup, the users model is called before is_deleted has been added.  I wonder if there's a more elegant way to do this 🤔

dredivaris: Seems easier to just catch the exception and that way we don't need `get_all_field_names` on the filter of `is_deleted` so if its not there we just ignore the exception and carry on.

label_studio/users/models.py
```diff
@@ -41,6 +46,15 @@ def _create_user(self, email, password, **extra_fields):
 
         return user
 
+    def get_queryset(self):
+        qs = super().get_queryset()
```
dredivaris: I debated putting this behind the feature flag as well but as it modifies the manager which required changes to the migrations themselves I'm leaning towards leaving it without the flag - this wont have an effect until users start being soft deleted.

label_studio/users/models.py
```diff
@@ -26,6 +27,10 @@
 class UserManager(BaseUserManager):
     use_in_migrations = True
 
+    def __init__(self, disable_is_deleted_filter=False, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.disable_is_deleted_filter = disable_is_deleted_filter
```
jombooth: please define and type this field on the UserManager class, like `use_in_migrations` above (which should really be `use_in_migrations: bool = True`)

label_studio/users/models.py
```diff
@@ -26,6 +27,10 @@
 class UserManager(BaseUserManager):
     use_in_migrations = True
 
+    def __init__(self, disable_is_deleted_filter=False, *args, **kwargs):
```
jombooth: Double negatives are a very common source of bugs, could we try to stick to positive phrasing and define this as `filter_soft_deleted: bool = True`?

label_studio/users/models.py
```diff
@@ -41,6 +46,15 @@ def _create_user(self, email, password, **extra_fields):
...
         return user
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if not self.disable_is_deleted_filter:
+            qs = qs.filter(is_deleted=False)
+        return qs
+
+    def with_deleted(self):
+        return super().get_queryset()
```
jombooth: Perhaps `with_soft_deleted`?

label_studio/users/models.py
```diff
@@ -190,6 +204,11 @@ def reset_token(self):
 
     def get_initials(self):
         initials = '?'
+
+        if flag_set('fflag_feat_all_optic_114_soft_delete_for_churned_employees', user='auto'):
```
jombooth: shouldn't we pass `self` as user? otherwise this won't work if used in redis queues for example

label_studio/users/serializers.py
```diff
@@ -30,6 +31,14 @@ def to_representation(self, instance):
         if uid not in self.context[key]:
             self.context[key][uid] = super().to_representation(instance)
 
+        if flag_set('fflag_feat_all_optic_114_soft_delete_for_churned_employees', user='auto'):
```
jombooth: similarly here I i think we can pass `instance` as `user`

label_studio/users/serializers.py
```diff
@@ -30,6 +31,14 @@ def to_representation(self, instance):
...
+            if instance.is_deleted:
+                for field in ['username', 'first_name', 'last_name', 'email']:
+                    if field == 'last_name':
+                        self.context[key][uid][field] = 'USER'
+                    else:
+                        self.context[key][uid][field] = 'DELETED'
```
jombooth: nit: could be

```
self.context[key][uid][field] = 'USER' if field == 'last_name' else 'DELETED'
```

jombooth (commented):
Just a few more things to address

jombooth:
Here's a general strategy for checking if a migration affects the db: https://docs.djangoproject.com/en/4.2/ref/django-admin/#sqlmigrate

label_studio/users/models.py
jombooth: On further reflection, I would actually support implementing `with_deleted` as a different manager on the model class instead of a method within this manager. This would unlock writing queries like `Users.with_deleted.<method>` instead of `Users.objects.with_deleted().<method>`

label_studio/users/models.py
bmartel: That was my inclination as well, the part that will naturally make this actionable will be feature flagged and until actually utilized will have no effect. The other argument would be if something were to break as a result of this change, but again this is under test so that should not be an issue.

label_studio/users/models.py
jombooth: Actually with the second manager refactoring, we can get rid of this param entirely, in favor of the class field. Instead we should subclass UserManager as WithDeletedUserManager, with the only change being that the `filter_soft_deleted` class field is set to True, then update the migration to use users.models.WithDeletedUserManager here https://github.com/HumanSignal/label-studio/pull/4832/files#diff-41b3a1e27082ee6c77df7e40fb9d0af80e381bf55d713ef72acd03ca6d79a4b4R71

jombooth (approved):
:heart_eyes:

bmartel (approved):

PR #4870, opened 2023-10-05
feat: LEAP-53: Move editor into the monorepo

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

This is the last piece of the migration, here we are putting all together, DM, Editor and ReactApp, we are using NX to compile everything.
On this update we had to add different webpack configs to each project, so we can have unique css prefixes for each project, also it keeps being compiled as it was, following one of the premisses on this project that is not change anything, just migrate from where we had to the new structure.

Basically nothing changes on the behavior on the user perspective, but all the migration will make things easier to develop and to organize what is LSO and what is LSE

The current behavior thing on the user point of view is exactly the same, what is changing is the structure, we are removing DM and LSF from LSE and adding on LSO.

none

On the user perspective nothing, but probably we will feel some improvement to build the project as NX has a great cache system.

no

There are different tools that we could have used here, but it was broadly discussed and defined that the best way we should follow is using NX

No feature flags defined, we decided as we are doing a migration that changes totally the structure, in case of problem we should roll back the commit to the last stable point.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


web/package.json
```diff
@@ -3,24 +3,37 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
-    "dev:watch": "nx run labelstudio:build:development --watch",
+    "ls:watch": "nx run labelstudio:build:development --watch",
+    "lsf:watch": "nx run editor:build:development --watch",
+    "dm:watch": "nx run datamanager:build:development --watch",
```
yyassi-heartex, line 8: anyway we can get a `fe:watch` which gets all three watchers going?

yyassi-heartex (approved):
looks good in the sense that nothing jumps out at me as bad - i have a minor request but its optional

bmartel (approved):

PR #4876, opened 2023-10-06
feat: OPTIC-116: User soft-deletion API

dredivaris:
This PR introduces a soft deletion API.  The core `api.UserAPI` already includes a deletion endpoint, so this API uses a separate `UserSoftDeleteView` which appends `soft-delete/` to the deletion endpoint to differentiate it from the standard deletion endpoint.

Permissions here are a bit tricky as role permissions are required in LSE but need to be accommodated in the endpoint defined in LS, to handle this I've added a `USER_PERM` setting which specifies which permission class to use.   LS currently allows only owners to delete users.

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/core/api_permissions.py
```diff
@@ -4,3 +4,10 @@
 class HasObjectPermission(BasePermission):
     def has_object_permission(self, request, view, obj):
         return obj.has_permission(request.user)
+
+
+class HasOwnerPermission(BasePermission):
```
bmartel, line 9: Are soft deletes exclusive to Owner, or would this be something that an Admin would also be capable of?

dredivaris: This permission handler is native to LS where there are no roles.  In LSE you're right I'm using LSE's permission handler that handles admins as well (but I'll double check to be sure)

bmartel (approved):

dredivaris:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-optic-116 (47de35b56afe32a44c416f40fe8167a2ba738c97)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6436027957)

label_studio/tests/users.tavern.yml
```diff
@@ -83,3 +83,99 @@ stages:
...
+    method: GET
+    url: '{django_live_url}/api/users/{user_pk}'
+  response:
+    status_code: 200
+    save:
+      json:
+        org_pk: active_organization
+- type: ref
+  id: get_invite_url
+- type: ref
```
jombooth: formatting: please put a blank line between each stage, this makes the tests much easier to read

label_studio/users/models.py
```diff
@@ -165,7 +165,7 @@ def active_organization_contributed_project_number(self):
 
     @property
     def own_organization(self):
-        return Organization.objects.get(created_by=self)
+        return Organization.objects.filter(created_by=self).first()
```
jombooth: Couple notes here:

- in general let's use `fast_first` by default, we probably don't need the ordering behavior of `first`
- this changes the type of own_organization; before its return type was `Organization`, and now it's `Optional[Organization]`. Could we add that return type annotation, and make sure there are no callers that aren't prepared to handle a `None`?

label_studio/users/models.py
```diff
@@ -199,6 +199,12 @@ def reset_token(self):
             token.delete()
         return Token.objects.create(user=self)
 
+    def soft_delete(self):
```
jombooth: Here and below, let's add type annotations to all new functions.

label_studio/users/models.py
```diff
@@ -199,6 +199,12 @@ def reset_token(self):
...
+        self.is_deleted = True
```
jombooth, line 203: This doesn't seem to call `self.save()`. If that's intentional, we should explain the rationale in a comment; also I'd be in favor of a docstring explaining why this deletes Token objects

label_studio/users/models.py
```diff
@@ -199,6 +199,12 @@ def reset_token(self):
...
+        token = Token.objects.filter(user=self)
+        if token.exists():
+            token.delete()
```
jombooth: this can be `Token.objects.filter(user=self).delete()`, one fewer query

label_studio/users/views.py
```diff
@@ -149,3 +158,28 @@ def user_account(request):
         'users/user_account.html',
         {'settings': settings, 'user': user, 'user_profile_form': form, 'token': token},
     )
+
+
+class UserSoftDeleteView(generics.RetrieveDestroyAPIView):
+    permission_required = ViewClassPermission(
+        DELETE=all_permissions.organizations_change,
+    )
+    queryset = User.with_deleted.all()
```
jombooth: This would allow soft-deleting a soft-deleted user a second time - any reason for that? I think the other manager might be a better fit here

label_studio/users/views.py
```diff
@@ -149,3 +158,28 @@ def user_account(request):
...
+    serializer_class = UserSerializer
+    permission_classes = (IsAuthenticated, HasObjectPermission)
+
+    def get_object(self):
+        pk = self.kwargs[self.lookup_field]
+        user = self.queryset.filter(active_organization=self.request.user.active_organization).get(pk=pk)
```
jombooth: Probably worth a comment that this line effectively stops users from being able to delete users in other organizations (I take it we can rely on `request.user.active_organization` for this?)

label_studio/tests/users.tavern.yml
```diff
@@ -83,3 +83,99 @@ stages:
...
+- id: logout
+  type: ref
+- name: get_my_user_with_token_fails_as_user_is_now_deleted
+  request:
+    headers:
+      authorization: "Token {user_token}"
+    url: "{django_live_url}/api/current-user/whoami"
+    method: GET
+  response:
+    status_code: 401
```
jombooth, line 213: The reason this test passes is that the tokens were deleted, even though `is_deleted` wasn't actually set on the user as far as I can tell (see https://github.com/HumanSignal/label-studio/pull/4876/files#diff-18d3ee828d3b90980ee44c34eae7c6aa18eede8d3024d74149de34ebd7e77beaR203). Is there a more reliable way we could identify whether the user was soft deleted in test, eg a get request for the user by id?

jombooth (changes requested):
Requesting changes because it looks like the soft delete endpoint isn't actually updating the User#is_deleted field as it should be

label_studio/users/models.py
dredivaris: Actually there are no other callers of `own_organization` prior to my usage, looks like it wasn't being used at all.

label_studio/users/views.py
dredivaris: No I cant see a reason for that, bth not sure why I used that there.

label_studio/users/models.py
dredivaris: good catch, might as well improve the piece of code I copied it from as well (the method above)😆

label_studio/users/models.py
dredivaris: Old habits die HARD!  Can't wait till we have mypy in there to save you the trouble

dredivaris: ahh I forgot about the issues with mypy/lse its a shame its so tricky to get it working on here.

label_studio/users/models.py
dredivaris: Odd I thought I had added a save 🤔 this was recent refactor to move delete code into the user class I probably accidentally deleted it. What is concerning is the tests should have caught this.  Lemme take a closer look at whats going on with them.

label_studio/users/views.py
dredivaris: makes sense.  Yes I'm pretty certain thats the field we can rely on, but we can check with @bmartel  as well to be safe.

label_studio/tests/users.tavern.yml
dredivaris: Yep, I noticed this as well, I'm now adding methods to check this to the tests:
1. try deleting the user twice, now after the first deletion they should no longer be found by the endpoint and second deletion will fail
2. Try logging in with the user after they have been deleted, it should fail with a 200 (ironically) as 302 indicates successful login and redirection to projects page.

label_studio/tests/users.tavern.yml
```diff
@@ -83,3 +90,124 @@ stages:
...
+    method: DELETE
+  response:
+    status_code: 204
+
+- name: soft_delete_user_fails_second_deletion_attempt
+  request:
+    url: "{django_live_url}/users/{user_pk}/soft-delete"
+    method: DELETE
+  response:
+    status_code: 500
```
jombooth: We should avoid intentionally returning 500s, especially in a case where we know what went wrong - can we make this 404 instead?

label_studio/users/models.py
```diff
@@ -164,8 +165,8 @@ def active_organization_contributed_project_number(self):
         return annotations.values_list('project').distinct().count()
 
     @property
-    def own_organization(self):
-        return Organization.objects.get(created_by=self)
+    def own_organization(self) -> Organization:
```
jombooth: This type should be `Optional[Organization]` because `fast_first` can return a `None` (as with `first`). If there's no possibility of `None` values it's not necessary to change the implementation.

label_studio/users/views.py
jombooth: to this comment: https://github.com/HumanSignal/label-studio/pull/4876/files#r1353458664, I think the 500 is likely coming from this `get` call. One option would be to use `fast_first` on this queryset instead and handle the None case by raising an exception.

label_studio/users/models.py
dredivaris: 🤔 yeah I think you're right, only owners have an `own_organization` so for others it will return None (previous to my changes it was using `get` so it would just raise an exception 😬

jombooth (approved):
Approval modulo fixing the 500 when deleting an already-soft-deleted user.

label_studio/users/models.py
jombooth: I gotta resuscitate that `mypy` PR soon, mypy would 100% catch this :grin:

label_studio/tests/users.tavern.yml
dredivaris: Makes sense

PR #4907, opened 2023-10-16
docs: DOC-122: Data discovery with AWS storage

caitlinwheeless:
This PR:

- Adds information about creating a dataset from AWS S3 buckets
- Changes where the data discovery images are located to align with the standard image directory
- Small change to the CSS so that all images have the same border formatting
- Updates the image links on other dataset pages to reflect the new image folder location


docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+
+3. Complete the following fields: 
+
+    <div class="noheader rowheader">
+
+    | | |
+    | --- | --- |
+    | Bucket Name | Enter the name of the AWS S3 bucket. |
+    | Bucket Prefix | Enter the folder name within the bucket that you would like to use.  For example, `data-set-1` or `data-set-1/subfolder-2`.  |
+    | File Name Filter | Use glob format to filter which file types to sync. For example, to sync all JPG files, enter `*jpg`. To sync all JPG and PNG files, enter `**/**+(jpg\|png)`.<br><br>At this time, we support the following file types: .jpg, .jpeg, .png, .txt, .text |
```
pakelley, line 189: Maybe I'm missing something obvious, but why does jpg/png require a more complex prefix than just jpg? I'd expect it to just be `*(jpg|png)`?

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | Region Name | By default, the region is `us-east-1`. If your bucket is located in a different region, enter it here.  |
```
pakelley: I'd like for this wording to communicate (sort of explicitly) that if you /are/ in `us-east-1` that you should just keep the default.
I'm thinking something like:
* If your bucket is located in a different region, enter it here, otherwise keep the default
* If your bucket is located in a different region, overwrite this field with your desired region.

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | S3 Endpoint | Enter an S3 endpoint if you want to override the URL created by S3 to access your bucket. |
+    | Access Key ID | Enter the ID for the access key you created in AWS. Ensure this access key has read permissions for the S3 bucket you are targeting (see "Create an AWS access key" above). |
```
pakelley: Is there a way to link to that section? It seems like we can link to underlying sections like [Create a policy for the user](https://deploy-preview-4907--heartex-docs.netlify.app/guide/dataset_create.html#Create-a-policy-for-the-user) (which even auto-expands the section), but manually modifying the anchor in the url to [Create an AWS access key](https://deploy-preview-4907--heartex-docs.netlify.app/guide/dataset_create.html#Create-an-AWS-access-key) doesn't seem to work.

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | Secret Access Key | Enter the secret portion of the access key you created earlier. |
```
pakelley, line 187: I know we discussed this, so apologies if I'm forgetting a discussion point, but if there's an obvious page to link to in the AWS docs for this I think it'd be helpful. You and I were both confused about this, so I'm guessing we'll help some users out if we can link to this for them.

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | Session Token | If you are using a session token as part of your authorization (for example, MFA), enter it here. |
```
pakelley: I know we discussed this, so apologies if I'm forgetting a discussion point, but if there's an obvious page to link to in the AWS docs for this I think it'd be helpful. You and I were both confused about this, so I'm guessing we'll help some users out if we can link to this for them.

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | Treat every bucket object as a source file | **Enabled** - Each bucket object will be imported as a separate piece of data (unstructured data). This creates a URL for each bucket object.<br>You should leave this option enabled if your bucket contains BLOB storage files such as JPG, MP3, or similar file types. <br><br>**Disabled** - Disable this option if you are using structured data, such as JSON or CSV files.<br><br>**NOTE:** At this time, we only support unstructured data. |
```
pakelley: Ok, just want to massage this a bit, I'm thinking we could word this as something like:
```md
**Enabled** - Each object in the bucket will be imported as a separate record in the dataset.<br>You should leave this option enabled if you are importing a bucket of unstructured data files such as JPG, MP3, or TXT.
```
I think the note about creating a URL is a technical detail that isn't relevant, but correct me if I'm missing something there.
Also our use of unstructured/BLOB has been pretty loose, but I think this is a more accurate wording (obviously subjective though, and I trust you if you think I'm making it less clear).
Lastly, I want to point out text files explicitly here. The "or similar file types" seems redundant with "files such as" but I'll trust your opinion here as well if you like it there 🙂

pakelley: To square the "disabled" with my phrasing of "enabled", I'm thinking something like:
```md
**Disabled** - Disable this option if you are importing structured data, such as JSON or CSV files.
```

pakelley: Lastly, because it's so painful for us to have this option without the user being able to use it, I just want to call out that we're working on structured data 😅 
I'm thinking something like:
```md
**NOTE:** At this time, we only support importing a bucket of unstructured data, but structured data is coming soon.
```
Granted this is a bit clumsy, but it's honest and the team is actively working on this support

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+    | Use pre-signed URLs | If your tasks contain `s3://…` links, they must be pre-signed in order to be displayed in the browser. |
+    | Pre-signed URL counter | Adjust the counter for how many minutes the pre-signed URLs are valid. |
+
+    </div>
+
+4. Click **Check Connection** to verify your credentials. If your connection is valid, click **Next**. 
+
+    ![Check Dataset connection](/images/data_discovery/dataset_check_connection_aws.png)
+
+5. Provide a name for your dataset column and select a type. The data type you select instructs Label Studio on how to index your data in a vector database, so that it can be searched using an AI-powered semantic search.
```
pakelley: I want to smooth out some technical details, lmk if you think this is clearer:
> Provide a name for your dataset column and select a data type. This data type tells Label Studio how to store your data in a way that can be searched using an AI-powered "semantic" search.

docs/source/guide/dataset_create.md
```diff
@@ -20,26 +20,217 @@ hide_sidebar: true
...
+            ]
+
+    ```
+
+
+###### Create a user 
+
+This user can be tied to a specific person or a group. 
+
+1. From the AWS Management Console, use the search bar or navigation menu to locate the **IAM** service.
```
pakelley, line 123: It probably makes sense to have our own steps here that are specific to the process for LS, but is there an obvious page in AWS's docs to link to here? So when our steps here go stale a user can still fall back to the AWS docs.

docs/source/guide/dataset_create.md
caitlinwheeless: I think that's just the way glob works? `*(jpg|png)` wouldn't work -- see https://globster.xyz/. FWIW, I was trying to give a somewhat simpler example vs. what we give in the placeholder, which is `**/**+(jp?(e)g|png)`

caitlinwheeless: Updated with your suggestion.

docs/source/guide/dataset_create.md
caitlinwheeless: Updated with:

By default, the region is `us-east-1`. If your bucket is located in a different region, overwrite the default and enter your region here. Otherwise, keep the default

docs/source/guide/dataset_create.md
caitlinwheeless: Ah - you're right! I tried to do that and didn't see an anchor in the expand-y thing title, but yeah - we can just link to the anchor within it. Added.

docs/source/guide/dataset_create.md
caitlinwheeless: Added links to AWS docs

docs/source/guide/dataset_create.md
caitlinwheeless: Yeah, all good edits. I was just copying and pasting from existing docs, and should have read closer. Only change I made from yours is to replace MP3 with PNG, since that is a supported file type. Also updated this field description under the Google portion.

docs/source/guide/dataset_create.md
caitlinwheeless: Yeah, I like to make it easier for beginner users so they don't have to flip back and forth between two sets of docs. Also the AWS docs can be overwhelming. But good point about it going stale -- added a link list at the top to AWS docs.

docs/source/guide/dataset_create.md
caitlinwheeless: Rewrote as "Provide a name for your dataset column and select a data type. The data type that you select tells Label Studio how to store your data in a way that can be searched using an AI-powered semantic search."

Also update under the GCS section.

pakelley (approved):

PR #4927, opened 2023-10-18
feat: OPTIC-181: refactoring of soft-delete from Users to OrganizationMembers

dredivaris:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/core/api_permissions.py
```diff
@@ -6,8 +6,9 @@ def has_object_permission(self, request, view, obj):
         return obj.has_permission(request.user)
 
 
-class HasOwnerPermission(BasePermission):
+class MemberHasOwnerPermission(BasePermission):
     def has_object_permission(self, request, view, obj):
-        if not request.user.own_organization or obj.active_organization != request.user.active_organization:
+        if request.method == 'DELETE' and not request.user.own_organization:
```
jombooth: :wave: just took a glance at this - I think we shouldn't use http methods in permissions implementations. I suspect one could still perform deletes by making a PATCH request with `deleted_at` to the `OrganizationMembershipAPI` for example.

Makefile
```diff
@@ -1,6 +1,6 @@
 # Run Django dev server with Sqlite
 run-dev:
-	DJANGO_DB=sqlite LOG_DIR=tmp DEBUG=true LOG_LEVEL=DEBUG DJANGO_SETTINGS_MODULE=core.settings.label_studio python label_studio/manage.py runserver
+	DJANGO_DB=sqlite LOG_DIR=tmp DEBUG=true LOG_LEVEL=DEBUG fflag_fix_all_optic_79_task_count_is_wrong_short=false DJANGO_SETTINGS_MODULE=core.settings.label_studio python label_studio/manage.py runserver
```
bmartel:
```suggestion
	DJANGO_DB=sqlite LOG_DIR=tmp DEBUG=true LOG_LEVEL=DEBUG DJANGO_SETTINGS_MODULE=core.settings.label_studio python label_studio/manage.py runserver
```
Seems this was accidentally committed.

label_studio/organizations/models.py
```diff
@@ -94,7 +121,7 @@ def has_project_member(self, user):
         return self.projects.filter(members__user=user).exists()
 
     def has_permission(self, user):
-        if self in user.organizations.all():
+        if self in user.organizations.filter(organizationmember__deleted_at__isnull=True):
```
wesleylima: I see that `if self in user.organizations...` is how this was being handled before, but I think this is suboptimal since it'll pull in all organizations the user is a member of when when we could just have the db handle the querying (I don't _think_ django magics a good query out this)

Something like `return OrganizationMember.objects.filter(user=user, organization=self, deleted_at__isnull=True).exists()` should be better.

label_studio/users/mixins.py
```diff
@@ -4,6 +4,6 @@ def is_annotator(self):
         return False
 
     def has_permission(self, user):
-        if user.active_organization in self.organizations.all():
+        if user.active_organization in self.organizations.filter(organizationmember__deleted_at__isnull=True):
```
wesleylima: Same idea here. Should be a query on OrganizationMember that has `active_organization` as a filter instead of `if ... in ....`

label_studio/users/serializers.py
```diff
@@ -18,7 +19,26 @@ def get_avatar(self, user):
...
+    def _is_deleted(self, instance):
+        if organization_members := self.context.get('organization_members', None):
+            organization_member_for_user = next(
+                (
+                    organization_member
+                    for organization_member in organization_members
+                    if organization_member.user_id == instance.id
+                ),
+                None,
+            )
```
wesleylima, line 37: So you have an `organization_members` iterable that you're filtering to find the `organization_member` object that for  user `instance` ?

I think I understand it, but I had to stare it for quite a while to get it. A comment to explain it might be helpful? Even better if it was it's own function or method so that you don't keep iterating through `organization_members` even after you get a `organization_member.user_id == instance.id` match.

label_studio/users/serializers.py
```diff
@@ -18,7 +19,26 @@ def get_avatar(self, user):
         return user.avatar_url
 
     def get_initials(self, user):
-        return user.get_initials()
+        if flag_set('fflag_feat_all_optic_114_soft_delete_for_churned_employees', user=user):
+            return user.get_initials(self._is_deleted(user))
+        else:
+            return user.get_initials()
+
+    def _is_deleted(self, instance):
```
wesleylima, line 27: could `instance` be `user` or are there cases where this is called and it isn't a user object?

label_studio/users/serializers.py
dredivaris: Yes I'll revisit and either add a comment or try to modify the code so its more clarifying (we can do this with a list with a break though I'm not sure if I prefer that).  This code is using a generator so we should only iterate once (via `next`)

wesleylima: Ah cool that makes sense. It'll quit the iteration when the match is found :+1:

label_studio/users/serializers.py
dredivaris: Its my understanding that `get_` methods in a DRF serializer receive an instance of the the model declared in their Meta class.  Here's an example from DRF docs:
```
class UserSerializer(serializers.ModelSerializer):
    days_since_joined = serializers.SerializerMethodField()

    class Meta:
        model = User
        fields = '__all__'

    def get_days_since_joined(self, obj):
        return (now() - obj.date_joined).days
```

I'll rename all those params so they're consistent though.

wesleylima (approved):

label_studio/organizations/models.py
```diff
@@ -65,6 +66,8 @@ def soft_delete(self):
             ).first()
             self.user.save(update_fields=['active_organization'])
 
+        TaskLock.release_locks_for_user(self.user)
```
wesleylima: could use the related names:
`self.user.task_locks.delete()`

dredivaris: Yeah I like that better

bmartel (approved):

wesleylima (approved):

bmartel:
/git merge develop
> Merge remote-tracking branch 'origin/develop' into fb-optic-181 (5a6254526916579d1cc82bfdac2792d873f32359)
>
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/6745613230)

PR #4965, opened 2023-10-25
docs: Support parents in sidebar menu

huguestennier:
Add support for parents in the sidebar with `parent` and `parent_enterprise` (if different from `parent`) frontmatter value


caitlinwheeless (approved):
Very exciting!

PR #5051, opened 2023-11-16
feat: LEAP-239: Move ls-frontend-test to monorepo and make cypress tests work

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

The reason for this ls-frontend-test repo is to have cypress helpers in all frontend repos and as we are merging them all into one repo it can also go there, which will simplify everything. Right now it doesn’t have Follow Merge and we have to update it manually.

#### What does this fix?
We bring ls-frontend-test repo as a lib into monorepo and make the cypress tests working as a monorepo e2e test
We also added some scripts to package as:
`lsf:integration` to run integration tests
`lsf:integration:watch` to run integration tests in UI mode
`lsf:test` -> `lsf:e2e` rename the test for better clarity on which test we are executing

no

no

no

_(check only one)_
- [x] No

_(check all that apply)_
- [x] e2e
- [x] integration
- [ ] unit


web/libs/frontend-test/src/cypress/support/commands.ts
```diff
@@ -0,0 +1,119 @@
...
+    cy.get(hiddenSelector).invoke('css', 'visibility', '');
+  }
+  log.end();
+  return obj;
+});
+
+Cypress.Commands.add('compareScreenshot', {
+  prevSubject: ['element'],
+}, (subject, name, compare, screenshotCompareOptions: Partial<Loggable & Timeoutable & CompareScreenshotOptions & Tresholdable> = {}) => {
+  const { treshold = 0.1, withHidden = [], ...screenshotOptions } = screenshotCompareOptions;
```
yyassi-heartex, line 65: I think we have a small type-i here `treshold`
and I see 0.1 is earlier in the file as `failureThreshold` can you define it as as a const value so it can be reused since it looks like we're using this value in more than 1 spot

yyassi-heartex (approved):
its looking good, i just had a note about a type-o (which i saw in multiple spots and a small tweak) but its likely these were existing minor issues that could be addressed another time - this ticket is more about moving things around and getting things ready for monorepo so ✅

PR #5095, opened 2023-11-27
docs:  DOC-159: Reorganizing pages in Getting Started section

caitlinwheeless:
Restructuring docs, starting with Getting Started section:

- Moved Quick Start to separate page
- Moved Terminology section to a separate page
- Moved location of Troubleshooting page under Labeling
- Moved "Starting Label Studio" to the more technical Install section 
- Moved security-focused info from Overview page to the page about Security
- Removed duplicate section about LSE vs. LSO features
- Updated links where necessary

### DIFF
--- a/docs/source/guide/FAQ.md
+++ b/docs/source/guide/FAQ.md
@@ -1,5 +1,5 @@
 ---
-title: Troubleshoot Labeling Issues
+title: Troubleshoot labeling issues
 short: Troubleshooting
 type: guide
 tier: all
--- a/docs/source/guide/FAQ.md
+++ b/docs/source/guide/FAQ.md
@@ -6,7 +6,7 @@ tier: all
 order: 220
 order_enterprise: 120
 section: "Labeling"
-meta_title: Troubleshoot Label Studio
+meta_title: Troubleshoot labeling issues
 meta_description: Troubleshoot common issuesTroubleshoot machine learning with Label Studio configuration and performance so that you can return to your machine learning and data science projects.
 ---
 
--- a/docs/source/guide/FAQ.md
+++ b/docs/source/guide/FAQ.md
@@ -1,11 +1,11 @@
 ---
-title: Troubleshoot Label Studio
+title: Troubleshoot Labeling Issues
 short: Troubleshooting
 type: guide
 tier: all
-order: 15
-order_enterprise: 10
-section: "Get started"
+order: 220
+order_enterprise: 120
+section: "Labeling"
 meta_title: Troubleshoot Label Studio
 meta_description: Troubleshoot common issuesTroubleshoot machine learning with Label Studio configuration and performance so that you can return to your machine learning and data science projects.
 ---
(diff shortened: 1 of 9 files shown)


docs/source/guide/get_started.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Label Studio
+title: Label Studio overview
```
hogepodge, line 2: Are we using Title Case for titles?

docs/source/guide/quick_start.md
```diff
@@ -0,0 +1,42 @@
...
+tier: opensource
+type: guide
+order: 7
+order_enterprise: 0
+meta_title: Quick Start
+meta_description: description
+section: "Get started"
+date: 2023-11-27 13:34:32
+---
+
```
hogepodge, line 13: Whitespace/newline consistency?

docs/source/guide/FAQ.md
```diff
@@ -1,11 +1,11 @@
 ---
-title: Troubleshoot Label Studio
+title: Troubleshoot Labeling Issues
```
hogepodge: Is this the right title for us to be using? These all aren't just labeling issues, but can be issues with Label Studio.

hogepodge (changes requested):

docs/source/guide/FAQ.md
caitlinwheeless: Ah - you're right. It should be sentence case. Habit of mine. I'll fix this (I think there are a few others too.)

docs/source/guide/get_started.md
caitlinwheeless: Sentence case. This one is right, the other one is wrong.

docs/source/guide/quick_start.md
caitlinwheeless: Fixed

docs/source/guide/FAQ.md
caitlinwheeless: I wanted to move this out of Getting Started, because it seems too technical for a gentle introduction to the product. From what I can tell, the sections most out of place are:

**Blank page when loading a project** -- this should be in a "project troubleshooting" page, but we we don't have one. We can keep it here until I create one. 
**HTML label offsets are in the wrong places** -- this would be better in "Import/Export troubleshooting." Again, I can create that in a separate PR. I already found another issue to add to that one.

amymrou: I'm concerned about consistent terminology. For example, meta_title reads "Troubleshoot Label Studio". Also share @hogepodge 's concern - the scope of troubleshooting in this document goes beyond labeling. If this is one step in a broader refactor of the documentation, I think this is ok.

caitlinwheeless: @amymrou - Good catch on the meta title. I'll fix that. 

My next PR is going to be surfacing our troubleshooting info in the docs. We currently have 7 troubleshooting sections. Five of them are subsections sprinkled throughout the docs. Two are their own pages. I want to consolidate in some places and make them all top-level pages. 

From this page, there are two issues that are not labeling related. They'll "live" here briefly, but I'll move them out in my next PR where I'll be focusing on troubleshooting.

hogepodge (approved):
Thanks for the changes. LGTM.

PR #5137, opened 2023-12-04
chore: LEAP-176: set up CSP in Label Studio

jombooth:
Add a reasonably strict CSP policy to Label Studio Open Source, defaulting to report only mode, everywhere except the /data/ API, where the only CSP directive is "sandbox" and true CSP (not report only mode) will be used.

Adds several environment variables:

- ENABLE_CSP (default True)
- LS_CSP_REPORT_ONLY (default True)
- LS_CSP_REPORT_URI (default None)

We should test this policy on our deployments + check for reports. In a future version we can change the REPORT_ONLY default to False.


label_studio/templates/base.html
```diff
@@ -59,13 +59,7 @@
 <div class="app-wrapper"></div>
 
 <template id="main-content">
-  <main class="main" style="background: transparent">
-
-      <div class="ui floating dropdown theme basic" style="float: right;">
-        {% block top-buttons %}
-        {% endblock %}
-      </div>
-    </div>
```
jombooth, line 68: as discussed with @hlomzik this wasn't used and is broken.

label_studio/core/settings/base.py
```diff
@@ -674,3 +674,43 @@ def collect_versions_dummy(**kwargs):
...
+        'https://*.google.com',
+    )
+    CSP_CONNECT_SRC = (
+        "'self'",
+        "'report-sample'",
+        'https://*.google-analytics.com',
+        'https://*.analytics.google.com',
+        'https://analytics.google.com',
+        'https://*.googletagmanager.com',
+        'https://*.g.double' + 'click.net',  # hacky way of suppressing codespell complaint
```
bmartel, line 708: Codespell was complaining about this line 😅? I'm a bit concerned if codespell is running on words in code files, not certain the validity of most words and their usage there. Source code is very much a meta language with regards to spelling and grammar 😬.

label_studio/core/settings/base.py
```diff
@@ -674,3 +674,43 @@ def collect_versions_dummy(**kwargs):
...
+        'https://*.ingest.sentry.io',
+    )
+    # Note that this will be overridden to true CSP for views that use the override_report_only_csp decorator
```
bmartel: I had to re-read this a few times. Does this mean that the header will be rewritten to Report CSP if true, or the LS_CSP_REPORT_ONLY=True will be overridden to False making it use CSP? I believe it is the latter just clarifying.

bmartel:
```suggestion
    # Note that LS_CSP_REPORT_ONLY will be overridden and use actual CSP for views that use the override_report_only_csp decorator
```

bmartel: The use of the word true was making it hard to quickly grok this note. Suggesting the above as a possible amendment.

bmartel (approved):

label_studio/core/settings/base.py
jombooth: It wanted 'double-click' instead of 'doubleclick' and afaik there is no way to suppress it with a comment :upside_down_face: :sweat_smile:

label_studio/core/settings/base.py
jombooth: https://github.com/HumanSignal/label-studio/pull/5137/commits/f9d06093e608af53939579e17fd474354de52723 I had actually made a change for this locally already but forgot to push it :grin: agreed!

PR #5157, opened 2023-12-08
fix: LEAP-449: remove Project model from migration path

jombooth:
When `redis_job_for_calculation` runs during migrations such as data_manager's 0008_manual_counters_update, it constructs Project objects, which sometimes fails because project fields haven't been created yet. This makes it impossible to migrate a Label Studio 1.5.0 db to 1.10.0, as @farioas discovered.

This PR avoids the need to construct Project instances, or select all Project fields, during this migration. This change makes it possible to upgrade from 1.5 to 1.10 straightforwardly.

Additionally, project.update_tasks_counters being async while project._update_tasks_counters was sync was pretty confusing. This PR eliminates that.

### DIFF
--- a/label_studio/tasks/functions.py
+++ b/label_studio/tasks/functions.py
@@ -3,13 +3,17 @@ import logging
 import os
 import sys
 
+from core.bulk_update_utils import bulk_update
 from core.models import AsyncMigrationStatus
 from core.redis import start_job_async_or_sync
 from core.utils.common import batch
 from data_export.mixins import ExportMixin
 from data_export.models import DataExport
 from data_export.serializers import ExportDataSerializer
+from data_manager.managers import TaskQuerySet
 from django.conf import settings
+from django.db import transaction
+from django.db.models import Count, Q
 from organizations.models import Organization
 from projects.models import Project
 from tasks.models import Annotation, Prediction, Task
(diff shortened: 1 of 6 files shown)


label_studio/tasks/functions.py
```diff
@@ -57,25 +61,34 @@ def redis_job_for_calculation(org_id, from_scratch, migration_name='0018_manual_
...
         )
+        project_tasks = Task.objects.filter(project_id=project_dict['id'])
         logger.debug(
-            f'Start processing stats project <{project.title}> ({project.id}) '
-            f'with task count {project.tasks.count()} and updated_at {project.updated_at}'
+            f'Start processing stats project <{project_dict["title"]}> ({project_dict["id"]}) '
+            f'with task count {project_tasks.count()} and updated_at {project_dict["updated_at"]}'
         )
 
-        task_count = project.update_tasks_counters(project.tasks.all(), from_scratch=from_scratch)
```
jombooth, line 72: I think this would previously have set `task_count` to `None`, since `project#update_tasks_counters` launched a Redis job and returned None. This is running in a task anyway, so probably fine to switch this call to a synchronous one.

bmartel (approved):

PR #5178, opened 2023-12-11
feat: DIA-771: Add validate methods for glob patterns if it matches at least something 

KonstantinKorotaev:
- [X] Docs have been added/updated (for bug fixes/features)

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [X] Backend (API)
- [ ] Frontend

Add validate methods for glob patterns if it matches at least something 

#### What does this fix?
_(if this is a bug fix)_
Add validate methods for glob patterns if it matches at least something 

Add validate methods for glob patterns if it matches at least something 

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Datasets


yyassi-heartex (approved):

label_studio/core/settings/base.py
```diff
@@ -714,3 +714,7 @@ def collect_versions_dummy(**kwargs):
     CSP_INCLUDE_NONCE_IN = ['script-src', 'default-src']
 
     MIDDLEWARE.append('core.middleware.HumanSignalCspMiddleware')
+
+CLOUD_MAX_CHECKED_OBJECTS = get_env('CLOUD_MAX_CHECKED_OBJECTS', 1000)
```
pakelley: As we discussed in SU, I don't think limiting our search to the first 1k records is a wide enough use case here, unfortunately. I haven't had the chance to look into the alternatives, but I think there are many cases where users need to be able to filter out many files from their bucket using the pattern they provide.

pakelley (changes requested):

label_studio/core/settings/base.py
KonstantinKorotaev: Maybe we can add a warning instead of validation error?
We can show such message: "First 1000 object do not match glob pattern. Narrow your glob pattern or proceed if glob pattern is correct"

label_studio/io_storages/azure_blob/utils.py
```diff
@@ -42,3 +45,36 @@ def get_blob_metadata(cls, url: str, container: str, account_name: str = None, a
...
+        regex = re.compile(str(pattern))
+        # match pattern against all keys in the container
+        for index, key in enumerate(generator):
+            # skip directories
+            if key.endswith('/'):
+                logger.debug(key + ' is skipped because it is a folder')
+                continue
+            if regex and regex.match(key):
+                logger.debug(key + ' matches file pattern')
+                return ""
```
pakelley: On one hand returning a string keeps thing simple, which is great, but on the other it seems weird to signal "yes, this is valid" using an empty string.
On the other other hand, I guess this just gets used by django under the hood; is this a pretty standard django pattern?

label_studio/core/settings/base.py
```diff
@@ -714,3 +714,6 @@ def collect_versions_dummy(**kwargs):
     CSP_INCLUDE_NONCE_IN = ['script-src', 'default-src']
 
     MIDDLEWARE.append('core.middleware.HumanSignalCspMiddleware')
+
+CLOUD_PAGE_CHECKED_OBJECTS = get_env('CLOUD_PAGE_CHECKED_OBJECTS', 10000)
```
pakelley: Since we're in a global config file, mind naming these something more specific? E.g. `DATA_DISCOVERY_NUM_OBJECTS_TO_CHECK_BEFORE_READ`? (I know this is pretty verbose, I'm up for some middle ground or other ideas, just want it to be relatively clear what these are for)

label_studio/io_storages/azure_blob/utils.py
hakan458: It allows us to keep the API generic while having the ability to return a custom error message including some params from the function itself

pakelley (approved):

PR #5277, opened 2024-01-10
ci: PLATE-795: Use pyproject.toml as a source version

nikitabelonogov:


label_studio/__init__.py
```diff
@@ -6,8 +6,7 @@
 package_name = 'label-studio'
 
 # Package version
-__version__ = '1.10.1dev'
-
+__version__ = importlib.metadata.metadata(package_name).get('version')
```
jombooth, line 9: I think this will set `__version__` to `1.10.2dev0`:

```
>>> package_name='label-studio'
>>> importlib.metadata.metadata(package_name).get('version')
'1.10.2.dev0'
```

which is slightly different to what is specified in pyproject.toml: https://github.com/HumanSignal/label-studio/blob/develop/pyproject.toml#L107

IIRC the frontend also reads pyproject.toml, and will do so using a toml reading package, and will obtain the version string `1.10.2dev`.

To fix this inconsistency, can we also change the value in pyproject.toml to be `1.10.2dev0`?

jombooth (approved):
LGTM, thank you!

PR #5313, opened 2024-01-19
fix: LEAP-634: Actualize label config validation

Gondragos:
Config validation didn't take into account many tags and their attributes. This PR might fix this. 

It also fixes nested `Choice`  validation.

It also will fix validation messages. (Usually there is no path at all and sometimes it doesn't tell about the problem itself)

### Explain the changes you've made
1. Changing `exc.path` into `exc.absolute_path` fixes displaying validation paths from `Validation failed on :`  to `Validation failed on View/Labels/Label: `.
2. Changing
```json
{
  "anyOf": [
    {
      "type": "array",
      "items": {"$ref": "#/definitions/ref1"}
    }, 
    {
      "$ref": "#/definitions/ref1"
    }
  ]
}  
```
into
```json
{
  "anyOf": [
    {
      "$ref": "#/definitions/ref1"
    },
    {
      "type": "array",
      "items": {"$ref": "#/definitions/ref1"}
    } 
  ]
}  
```
fixes displaying cause of validation fails from `OrderedDict() is not of type 'array'` to something more meaningful like `'value' is a required property`

3. The line
```python
# config = _fix_choices(config)
```
was hidden under the comment due to the problem, which caused this workaround, was solved by json schema itself.

Fixing displayed paths has leaded to the situation when
```xml
<View>
  <Text name="text" value="$text" />
  <Choices name="choice" toName="choice">
    <Choice />
    <Choice value="any" />
  </Choices>
</View>
```
shows `View/Choices/Choice/0`, but
```xml
<View>
  <Text name="text" value="$text" />
  <Choices name="choice" toName="choice">
    <Choice />
  </Choices>
</View>
``` 
shows  `View/Choices/Choice`. And previously it was just `Choice/0` in this case.

To make it more consistent we can use removed right now function `_fix_choices` but convert all tags into arrays and in the same time get rid of `MaybeMultiple*` rules in schema. But so far it doesn't look to me that it's necessary, 'cause doesn't really affects user's navigation.

- [ ] Performance
- [ ] Security
- [ ] UX

- [ ] No

It depends on the validity of the validation rules.

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)

`Labeling Interface`


label_studio/core/label_config.py
```diff
@@ -84,7 +84,7 @@ def parse_config_to_json(config_string):
     if xml is None:
         raise etree.ParseError('xml is empty or incorrect')
     config = xmljson.badgerfish.data(xml)
-    config = _fix_choices(config)
+    #  config = _fix_choices(config)
```
jombooth: Is this intentionally left in? Why don't we want to `_fix_choices` anymore?

jombooth: nevermind, I see this was mentioned in the description - my feeling is that if we really don't need this we can delete the line (?)

jombooth (commented):
I don't know that much about this label config validation code, and don't feel extremely confident approving without doing a sync review to understand its impact better. Is there anyone else more familiar with this who could take a look?

I will say we probably don't want to check in a commented-out function call; one way forward would be to remove the line and add a comment like 

```
# previously we called `_fix_choices` here, but no longer need to because...
```

Gondragos:
> I will say we probably don't want to check in a commented-out function call; one way forward would be to remove the line and add a comment like
> 
> ```
> # previously we called `_fix_choices` here, but no longer need to because...
> ```

@jombooth Thanks for that! It looks much better.

jombooth (approved):
LGTM, if someone more knowledgeable can do a second look that'd be great but I think we should get this over to QA, and the changes look reasonable!

triklozoid (approved):

robot-ci-heartex:

robot-ci-heartex:

robot-ci-heartex:

robot-ci-heartex:

robot-ci-heartex:

robot-ci-heartex:

PR #5316, opened 2024-01-19
fix: LEAP-396: More exhaustative IP validation for SSRF defenses, plus user configurability

jombooth:
Ban all IPs within reserved blocks, for both IPv4 and IPv6, by default: https://en.wikipedia.org/wiki/Reserved_IP_addresses

Additionally, allow users the ability to ban additional blocks (with `USER_ADDITIONAL_BANNED_SUBNETS`), or to specify their full list of banned IP blocks themselves (see `USE_DEFAULT_BANNED_SUBNETS`, default `True`), as their network setup may require.

Also, make the error message more helpful when SSRF protection blocks an upload:
image

### DIFF
--- a/label_studio/tests/data_import/test_uploader.py
+++ b/label_studio/tests/data_import/test_uploader.py
@@ -72,6 +72,53 @@ class TestUploader:
                 load_tasks(request, project)
             assert 'URL resolves to a reserved network address (block: 127.0.0.0/8)' in str(e.value)
 
+        def test_user_specified_block(self, project, settings):
+            settings.SSRF_PROTECTION_ENABLED = True
+            settings.USER_ADDITIONAL_BANNED_SUBNETS = ['1.2.3.4']
+            request = MockedRequest(url='http://validurl.com')
+
+            # Mock the necessary parts of the response object
+            mock_response = Mock()
+            mock_response.raw._connection.sock.getpeername.return_value = ('1.2.3.4', 8080)
+
+            # Patch the requests.get call in the data_import.uploader module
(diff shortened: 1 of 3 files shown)


label_studio/tests/data_import/test_uploader.py
```diff
@@ -70,7 +70,54 @@ def test_local_url_after_redirect(self, project, settings):
                 ValidationError
             ) as e:
                 load_tasks(request, project)
-            assert 'The provided URL was not valid.' in str(e.value)
+            assert 'URL resolves to a reserved network address (block: 127.0.0.0/8)' in str(e.value)
+
+        def test_user_specified_block(self, project, settings):
+            settings.SSRF_PROTECTION_ENABLED = True
```
bmartel, line 76: Do we have a test anywhere that outlines what the behaviour of this code is when: `settings.SSRF_PROTECTION_ENABLED = False`?

bmartel (approved):

wesleylima (approved):

label_studio/tests/data_import/test_uploader.py
jombooth: It's not covered in depth, but the TL;DR is that we don't check for hosts that point to reserved addresses if SSRF protection is off. We do, however, reject urls with unsafe schemes like `file://`, as shown in this test: https://github.com/HumanSignal/label-studio/pull/5316/files#diff-140c3d60bd366710ad0470bbc1907a34d4f3065d1e8647dc33636d0268100bf1L38-L47

PR #5317, opened 2024-01-19
feat: OPTIC-139: [M1] Workflow Automation: Task Assignment

bmartel:
Reinstates task assignment.


label_studio/data_manager/managers.py
```diff
@@ -498,8 +513,56 @@ def __init__(self, expression, distinct=False, output_field=None, **extra):
 def annotate_completed_at(queryset):
     from tasks.models import Annotation
 
+    LseProject = load_func(settings.LSE_PROJECT)
+    get_tasks_agreement_queryset = load_func(settings.GET_TASKS_AGREEMENT_QUERYSET)
+
```
luarmr, line 527: Let's do an early return here. Easy to read:
```
    if not (get_tasks_agreement_queryset and LseProject and queryset.exists()):
        return queryset
        
    project_id = queryset.first().project_id
    if not LseProject.objects.filter(project_id=project_id).exists():
        return queryset
```
Well adding the annotate. That also may be an opportunity to improve

luarmr: Btw not sure I understand this?
`LseProject.objects.filter(project_id=queryset[0].project_id).exists()`
Should not be enought with `queryset.exists())` os is a chance that this project_id doesn't exit?

label_studio/data_manager/managers.py
```diff
@@ -498,8 +513,56 @@ def __init__(self, expression, distinct=False, output_field=None, **extra):
...
+        and queryset
+        and LseProject.objects.filter(project_id=queryset[0].project_id).exists()
+    ):
+        queryset = get_tasks_agreement_queryset(queryset)
+
+        # Subquery to get the agreement_threshold for each project
+        agreement_threshold_subquery = Subquery(
+            LseProject.objects.filter(project_id=OuterRef('project_id')).values('agreement_threshold')[:1],
+            output_field=FloatField(),
+        )
```
luarmr: OuterRef('project_id') is need it or is always the same project id. 
Also coming from this
`LseProject.objects.filter(project_id=queryset[0].project_id).exists(`
if we check only the first i guess all are the same.
Sorry from throw comment without read the complete file

luarmr: I'll wait to confirm this part with. 
If it is only 1 project this will be way more easy since remove the need of subquery so I expected to be wrong.

label_studio/projects/functions/__init__.py
```diff
@@ -22,19 +22,13 @@ def annotate_finished_task_number(queryset):
 
 def annotate_total_predictions_number(queryset):
     if flag_set(
-        'fflag_fix_back_lsdv_4719_improve_performance_of_project_annotations',
```
wesleylima, line 25: This is just removing the `fflag_fix_back_lsdv_4719_improve_performance_of_project_annotations` feature flag

label_studio/projects/functions/__init__.py
```diff
@@ -22,19 +22,13 @@ def annotate_finished_task_number(queryset):
...
+        'fflag_perf_back_lsdv_4695_update_prediction_query_to_use_direct_project_relation',
```
bmartel, line 25: Is the fflag_fix_back_lsdv_4719_improve_performance_of_project_annotations still used at all, or is this intended to be removed?

label_studio/projects/functions/next_task.py
```diff
@@ -141,13 +143,36 @@ def get_not_solved_tasks_qs(user, project, prepared_tasks, assigned_flag, queue_
...
-    # if annotator is assigned for tasks, he must to solve it regardless of is_labeled=True
+    prioritized_on_agreement = False
+    # if annotator is assigned for tasks, he must solve it regardless of is_labeled=True
     if not assigned_flag:
-        not_solved_tasks = not_solved_tasks.filter(is_labeled=False)
+        # include tasks that have been completed if their agreement is not at threshold if threshold setting is set
+        lse_project = getattr(project, 'lse_project', None)
+        if (
+            lse_project
+            and flag_set('fflag_feat_optic_161_project_settings_for_low_agreement_threshold_score_short')
```
bmartel:
```suggestion
            and flag_set('fflag_feat_optic_161_project_settings_for_low_agreement_threshold_score_short', user='auto')
```
I think this is in scope here no?

label_studio/data_manager/managers.py
```diff
@@ -496,12 +510,60 @@ def __init__(self, expression, distinct=False, output_field=None, **extra):
 
 
 def annotate_completed_at(queryset):
+    if flag_set('fflag_feat_optic_161_project_settings_for_low_agreement_threshold_score_short'):
+        return annotated_completed_at_considering_agreement_threshold(queryset)
```
luarmr, line 514: basicly this is hidding from the rest of users, correct?

luarmr (approved):

wesleylima (approved):

label_studio/data_manager/managers.py
wesleylima: Yes. It leaves the default behavior/queries in place if the user doesn't have the feature flag on

label_studio/projects/functions/__init__.py
wesleylima: `fflag_fix_back_lsdv_4719_improve_performance_of_project_annotations` has been serving True to all for 7 months. Safe to remove.

label_studio/data_manager/managers.py
```diff
@@ -496,12 +510,60 @@ def __init__(self, expression, distinct=False, output_field=None, **extra):
...
+
+    newest_annotation = Annotation.objects.filter(task=OuterRef('pk')).order_by('-id')[:1]
+    if not LseProject:
+        # Not LSE so there will not be agreement_threshold-based task completeness
+        return queryset.annotate(
+            completed_at=Case(When(is_labeled=True, then=Subquery(newest_annotation.values('created_at'))))
+        )
+
+    lse_project = LseProject.objects.filter(project_id=project_id).first()
+    agreement_threshold = lse_project.agreement_threshold
```
bmartel: is this possibly attempting to call against None?

maybe guard it just in case and drop the need to check lse_project first in the not below
```suggestion
    agreement_threshold = lse_project.agreement_threshold if lse_project else None
```

label_studio/projects/functions/next_task.py
wesleylima: doesn't hurt to try!

label_studio/data_manager/managers.py
```diff
@@ -496,12 +510,60 @@ def __init__(self, expression, distinct=False, output_field=None, **extra):
...
+    if not (get_tasks_agreement_queryset and lse_project and agreement_threshold):
```
bmartel:
```suggestion
    if not (get_tasks_agreement_queryset and agreement_threshold):
```

label_studio/projects/functions/next_task.py
```diff
@@ -141,13 +143,36 @@ def get_not_solved_tasks_qs(user, project, prepared_tasks, assigned_flag, queue_
...
+            and lse_project.agreement_threshold is not None
+            and get_tasks_agreement_queryset
+            and user.is_annotator
+        ):
+            not_solved_tasks = (
+                get_tasks_agreement_queryset(not_solved_tasks)
+                # include tasks that are not labeled or are labeled but fall below the agreement threshold
+                .filter(
+                    Q(_agreement__lt=lse_project.agreement_threshold, is_labeled=True) | Q(is_labeled=False)
```
bmartel, line 162: Not a question for right now, but this line struck me as a bit odd. When would we be considering  any task as labeled that is also under the agreement_threshold? Is that even possible? If we didn't require that check and could use the agreement_threshold ubiquitously that would surely streamline this call further.

wesleylima: There's a configurable max annotations. If it reaches that max number of annotations even through it falls below the threshold, it's considered annotated

label_studio/tasks/models.py
```diff
@@ -247,6 +249,25 @@ def has_lock(self, user=None):
...
+    def overlap_with_agreement_threshold(self, num, num_locks):
+        try:
+            from stats.models import get_task_agreement
+        except (ModuleNotFoundError, ImportError):
+            pass
+
+        # Limit to one extra annotator at a time when the task is under the threshold and meets the overlap criteria,
+        # regardless of the max_additional_annotators_assignable setting. This ensures recalculating agreement after
+        # each annotation and prevents concurrent annotations from dropping the agreement below the threshold.
+        if hasattr(self.project, 'lse_project') and self.project.lse_project.agreement_threshold is not None:
```
bmartel:
```suggestion
        if hasattr(self.project, 'lse_project') and self.project.lse_project is not None and self.project.lse_project.agreement_threshold is not None:
```

label_studio/tasks/models.py
```diff
@@ -247,6 +249,25 @@ def has_lock(self, user=None):
     def num_locks(self):
         return self.locks.filter(expire_at__gt=now()).count()
 
+    def overlap_with_agreement_threshold(self, num, num_locks):
+        try:
+            from stats.models import get_task_agreement
+        except (ModuleNotFoundError, ImportError):
+            pass
```
bmartel: This should be moved within the if check below. No point in adding a failure condition for 50% of the use cases.

wesleylima (approved):

PR #5324, opened 2024-01-22
docs: DOC-167: Add highlight description for on-prem release notes

caitlinwheeless:
To improve the release notes page usability, added a brief highlight description for each release

- [ ] Community docs
- [X] Enterprise docs

### DIFF
--- a/docs/source/guide/release_notes/onprem/2.4.3.md
+++ b/docs/source/guide/release_notes/onprem/2.4.3.md
@@ -4,6 +4,8 @@ hide_sidebar: true
 
 ## Label Studio Enterprise 2.4.3
 
+<div class="onprem-highlight">New <code>splitchannels</code> option on audio configs, keyboard shortcuts for the Data Manager, restore locked annotations</div>
+
 ### Enhancements
 - Support simultaneous render of multi-channel audio with added splitchannels="true" option on Audio config (larger memory requirement)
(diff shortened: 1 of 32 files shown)


docs/source/guide/release_notes/onprem/2.4.3.md
```diff
@@ -4,6 +4,8 @@ hide_sidebar: true
 
 ## Label Studio Enterprise 2.4.3
 
+<div class="onprem-highlight">New <code>splitchannels</code> option on audio configs, keyboard shortcuts for the Data Manager, restore locked annotations</div>
```
jombooth, line 7: potentially silly question - should we use backticks (\`) instead of code tags around `splitchannels`, to avoid line breaks? The GH preview suggests we'll get potentially unwanted linebreaks from using the code tags

jombooth (approved):
LG!

docs/source/guide/release_notes/onprem/2.4.3.md
caitlinwheeless: I tried, but they didn't render correctly in the HTML. I think it's because they're in a `<div>` element and not a `<p> ` element? It was either leave them unstylized or specify the HTML.

jombooth: SGTM, fine to leave as is!

PR #4443, opened 2023-06-26
docs: LSDV-5343: Docs for visibleWhen; remove FF_DEV_1372

robot-ci-heartex:
Hi @hlomzik!

This PR was created in response to a PRs in upstream repo:
- https://github.com/heartexlabs/label-studio-frontend/pull/1483
- https://github.com/HumanSignal/label-studio-frontend/pull/1483

PR #5349, opened 2024-01-23
chore: LEAP-516,LEAP-517: upgrade `follow-redirects` package to address CVE-2023-26159

jombooth:
Upgrades follow-redirects in web/, docs/ and web/libs/editor/tests/e2e/. Methodology used was from here: https://github.com/yarnpkg/yarn/issues/4986#issuecomment-1500862435


jombooth:
@hlomzik @Gondragos - Clicking around on the Netlify docs previews above, links seem to be working okay. Is there a different way I should be testing the docs build?

Gondragos:
@jombooth To be honest I can't say I'm an expert in this docs part. But I would say that docs were not generated this time. It looks like it needs some changes sources to do that or something like that.
As I understand here is an example of how it should look like: https://github.com/HumanSignal/label-studio/pull/4443
There is the `Deploy Preview` row in a table which I think might be an indication of an existing of a new build.

jombooth:
@Gondragos I think it does successfully build, then get updated to say "cancelled" for some reason. See for example: https://app.netlify.com/sites/heartex-docs/deploys/65b194384af5560008ac732c

Gondragos:
@jombooth Actually, I think you're right. From what I see there, it makes sense. 
I just want @hlomzik to confirm that as I believe he could have some more context.

hlomzik (approved):

PR #5358, opened 2024-01-24
feat: DIA-868: Component library + Chip Input

nick-skriabin:
Introducing the UI component library with its first citizen - the ChipInput component.

UI library contains:
- UI (React) component written in TypeScript and styled with SCSS
- Assets library -- icons in SVG format
- Components tests -- Cypress
- Storybook with interactive access to the components and/or components composition

This work was initiated by DIA-868 to support so-called "classes input" (see designs in the ticket). The basis for this component is "MultiEmailInput" from LSE's invitation flow. The component was refactored from JS+Stylus into TS+SCSS, and generalized to be universal.

The work on the component library was initiated due to inability to test components that are not the part of the UI yet.

Note: Name "Classes component" comes from the upcoming PW feature, but the general name for the underlying component is "ChipInput".

- [x] Tests for the changes have been added/updated

### What feature flags were used to cover this change?
None

Implement the whole interface and include a bunch of tickets in a single PR.

- [ ] Performance
- [ ] Security
- [x] UX - new generalized chip component to support future work
- [x] DX - introduction of the component library must improve DX by avoiding code duplication, scope creep. It also improves code quality as every component must be covered with tests.

- [x] No

- [ ] e2e (codecept)
- [x] integration (cypress)
- [ ] unit (jest)

Monorepo, PW


web/libs/storybook/src/components/ChipInput.scss
```diff
@@ -0,0 +1,96 @@
+@import "../assets/styles/base.scss";
+
+$danger_color: #ff0000;
```
yyassi-heartex: might be a prudent to add the danger color to base.scss to make it available to other components as more components get added here

web/libs/storybook/tests/support/e2e.ts
```diff
@@ -0,0 +1,24 @@
+import { CURRENT_FLAGS } from '../feature-flags';
+import '@humansignal/frontend-test/cypress/support/e2e'
+
+Cypress.on('uncaught:exception', (err, runnable) => {
```
hlomzik:
```suggestion
Cypress.on('uncaught:exception', () => {
```

web/libs/storybook/tests/support/e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+  //it is to prevent outliner failing tests with uncaught error
+  return false;
+});
+
+beforeEach(() => {
+  cy.on('uncaught:exception', err => {
+    return !err.message.includes('ResizeObserver loop completed with undelivered notifications.');
```
hlomzik: Most probably we have some other errors, I remember @Gondragos added similar ignore procedure for them. And potentially we could have such error, so better to use array of errors to ignore.

web/libs/storybook/.storybook/main.ts
```diff
@@ -0,0 +1,34 @@
...
+    options: {},
+  },
+
+  viteFinal: async (config) =>
+    mergeConfig(config, {
+      plugins: [
+        nxViteTsPaths(),
+        svgr({
+          esbuildOptions: {},
+          svgrOptions: { plugins: ['@svgr/plugin-jsx'], ref: true },
```
hlomzik, line 24: Not sure how does this work exactly, but in all our other configs we also use `removeViewBox: false` for proper work of some icons.

web/libs/storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,36 @@
...
+    --gray-8: #9d9d9d;
+    --gray-9: #8c8c8c;
+
+    --theme-error: var(--red-6);
+    --theme-text: #171819;
+    --theme-white: white;
+    --theme-accent: var(--blue-6);
+    --theme-bg: white; /*var(--gray-2)*/
+
+    --pirmary-on-containte: #052861;
```
hlomzik: Definitely typo in "primary" and I'm not sure I understand the last word

hlomzik: and in other places as well

web/libs/storybook/README.md
hlomzik, line 1: Would be great to use this README to describe main principles on how to work with storybook here. For example should we really put components directly here, which ones and why?

hlomzik (changes requested):

web/libs/storybook/tests/support/e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+    console.log('Setting feature flags', CURRENT_FLAGS);
+    Object.assign(win, {
+      DISABLE_DEFAULT_LSF_INIT: true,
+      APP_SETTINGS: {
+        ...(win.APP_SETTINGS ?? {}),
+        feature_flags: CURRENT_FLAGS,
+      },
+    });
+  });
+  cy.log('Feature flags set');
```
hlomzik: we have both cy.log and console.log here. do we need them?

web/libs/storybook/.storybook/main.ts
nick-skriabin: just checked, it doesn't remove it, so we're good

web/libs/storybook/tests/support/e2e.ts
nick-skriabin: ty, reorganized the setup here. I also removed whole e2e section as it's the same thing in cypress. so we only have components tests there now

web/libs/storybook/README.md
nick-skriabin: still working on it. there'l be readme + separate doc about the components lib

web/libs/storybook/tests/support/e2e.ts
nick-skriabin: removed it for now completely as we don't have component-level feature flags yet

web/libs/storybook/src/components/ChipInput.scss
nick-skriabin: replaced with an existing css var

web/libs/storybook/.storybook/main.ts
hlomzik: ah, yes, NX does this automatically, that's why it was good in monorepo and failed during back-porting!

web/libs/storybook/.storybook/preview-head.html
hlomzik: other places fixed but this exact line is still untouched :D

hlomzik (approved):

web/libs/storybook/src/assets/icons/index.ts
```diff
@@ -0,0 +1,3 @@
+/// <reference types="vite-plugin-svgr/client" />
+
```
yyassi-heartex, line 2: this is a good opportunity to clean things up
```suggestion
```

nick-skriabin: without it for some reason TS complains about types (can't import things with `?react`). gotta figure out a different way. but this is recommended by the svgr themself

yyassi-heartex: interesting, then lets disregard that - thanks for following up!

yyassi-heartex (approved):

MihajloHoma (approved):

PR #5394, opened 2024-01-30
chore: Bump label-studio-converter version to 0.0.58

robot-ci-heartex:
Hi @robot-ci-heartex!

This PR was created in response to a manual trigger of the release workflow here: https://github.com/HumanSignal/label-studio/actions/runs/7714911871.

PR #5396, opened 2024-01-30
chore: LEAP-608: bump label-studio-converter version

jombooth:
Supersedes #5394. Adds a new version of LSC that will support JSON-MIN exports from projects that use Repeater


jombooth:
closed due to follow merge failure

PR #5432, opened 2024-02-06
feat: LEAP-648: Reset and recalculate project summary

makseq:
Adds "Reset Cache" button to the Danger Zone page within project settings. Tested using pytest rather than tavern for fine-grained control over the contents of ProjectSummary json fields

image


label_studio/projects/functions/utils.py
```diff
@@ -26,3 +26,17 @@ def make_queryset_from_iterable(tasks_list):
...
+    if project.annotations.exists():
+        summary.update_created_annotations_and_labels(project.annotations.all())
+    else:
+        summary.save(update_fields=['created_labels'])
+    summary.created_labels_drafts = {}
+    drafts = AnnotationDraft.objects.filter(task__project=project)
+    if drafts.exists():
+        summary.update_created_labels_drafts(drafts)
+    else:
+        summary.save(update_fields=['created_labels_drafts'])
```
jombooth: Do we really need this branching? Seems to me that if there are no drafts, `update_created_labels_drafts` should save an empty dict of drafts anyway, right?

web/apps/labelstudio/src/pages/Settings/DangerZone.js
```diff
@@ -59,11 +65,20 @@ export const DangerZone = () => {
     type: 'predictions',
     disabled: true, //&& !project.total_predictions_number,
     label: `Delete ${project.total_predictions_number} Predictions`,
+  }, {
+    type: 'label_cache',
+    help:
+      `If you are unable to modify the labeling configuration due to validation errors ` +
+      `concerning existing labels, but you are confident that the labels don't exist. You can ` +
+      `use this action to reset the cache and try again.`,
+    label: `Reset Label Cache`,
```
hlomzik:
```suggestion
      'If you are unable to modify the labeling configuration due to validation errors ' +
      'concerning existing labels, but you are confident that the labels don't exist. You can ' +
      'use this action to reset the cache and try again.',
    label: 'Reset Label Cache',
```

web/apps/labelstudio/src/pages/Settings/DangerZone.js
```diff
@@ -28,6 +28,12 @@ export const DangerZone = () => {
           // console.log('delete tasks');
         } else if(type === 'predictions') {
           // console.log('delete predictions');
+        } else if(type === 'label_cache') {
```
hlomzik:
```suggestion
        } else if (type === 'label_cache') {
```

label_studio/projects/functions/utils.py
```diff
@@ -26,3 +26,17 @@ def make_queryset_from_iterable(tasks_list):
                 raise ValueError(f'Unknown object type: {str(task)}')
         queryset = Task.objects.filter(id__in=ids)
     return queryset
+
+
+def recalculate_created_labels_from_scratch(project, summary):
+    summary.created_labels = {}
```
jombooth: should we also clear out `summary.created_annotations` here? If so, we should probably rename the function. If not, I'm concerned that `update_created_annotations_and_labels` may further mess up `summary.created_annotations`

label_studio/projects/functions/utils.py
```diff
@@ -26,3 +26,17 @@ def make_queryset_from_iterable(tasks_list):
...
+    if project.annotations.exists():
+        summary.update_created_annotations_and_labels(project.annotations.all())
+    else:
+        summary.save(update_fields=['created_labels'])
```
jombooth: ditto feedback below re: the branching, though it's more complicated in this case because there are two fields potentially affected by the first codepath.

label_studio/projects/api.py
```diff
@@ -397,6 +399,31 @@ def get(self, *args, **kwargs):
         return super(ProjectSummaryAPI, self).get(*args, **kwargs)
 
 
+class ProjectSummaryResetCreatedLabelsAPI(GetParentObjectMixin, generics.CreateAPIView):
```
jombooth: Please add a unit test for this API. Also, is the `IsAuthenticated` permission not needed here? I do see that `GetParentObjectMixin` will check object permissions, but I'm not sure if that means we don't need to check authentication. Either way, this could be checked in a test.

label_studio/projects/models.py
```diff
@@ -1200,6 +1200,15 @@ def update_created_annotations_and_labels(self, annotations):
         self.save(update_fields=['created_annotations', 'created_labels'])
 
     def remove_created_annotations_and_labels(self, annotations):
+        # we are going to remove all drafts, make full refresh
+        if self.project.annotations.count() == annotations.count():
```
jombooth: Why do these need to be equal?

label_studio/projects/functions/utils.py
makseq: I just wanted to make it maximally clear, and who knows - maybe update_created_labels_drafts will be changed after sometime, however this reset function will have 100% confidence that it was reset to {}.

label_studio/projects/functions/utils.py
makseq: you are right, added created_annotations too and renamed the func.

label_studio/projects/functions/utils.py
jombooth: I think the invariant that this call resets to `{}` if there are no drafts would be better captured in a unit test instead (same for https://github.com/HumanSignal/label-studio/pull/5432#discussion_r1480632602)

label_studio/projects/functions/utils.py
makseq: should be fixed now

label_studio/projects/models.py
makseq: this way we can know that all existing annotations in the project are going to be removed and it's a special case where we can totally reset created_labels = {}

label_studio/projects/api.py
makseq: there is below:
```
  permission_required = ViewClassPermission(
        POST=all_permissions.projects_change,
    )
```
isn't it enough?

label_studio/projects/functions/utils.py
makseq: hm.. sorry, I don't understand what invariant you mean..

label_studio/projects/api.py
makseq: tests are added

label_studio/projects/models.py
makseq: other words - this signals that annotations will be deleted fully

label_studio/projects/models.py
```diff
@@ -1200,6 +1200,15 @@ def update_created_annotations_and_labels(self, annotations):
         self.save(update_fields=['created_annotations', 'created_labels'])
 
     def remove_created_annotations_and_labels(self, annotations):
+        # we are going to remove all drafts, make full refresh
+        if self.project.annotations.count() == len(annotations):
+            # it's hard to calculate created_labels_drafts properly, so it's better to reset it completely
```
jombooth: these comments are copy-pasted from below, please fix them to refer to annotations and labels rather than drafts

label_studio/projects/functions/utils.py
jombooth: The behavior where there are no drafts on the project is now captured by this unit test:

https://github.com/HumanSignal/label-studio/pull/5432/files#diff-fc2a039da0975ee919925c1f66e53984db5044b15d741eed50f66837eae03364R63-R95

The behavior where there are no annotations on the project is now captured by this unit test:

https://github.com/HumanSignal/label-studio/pull/5432/files#diff-fc2a039da0975ee919925c1f66e53984db5044b15d741eed50f66837eae03364R27-R60

Now we don't have to check whether `drafts.exists()` in this function, or call `save()` in a separate codepath, because we know from the tests that the `update_created_labels_drafts` function will do the right thing when the field on the summary has been cleared.

label_studio/projects/api.py
jombooth: https://github.com/HumanSignal/label-studio/pull/5432/files#diff-fc2a039da0975ee919925c1f66e53984db5044b15d741eed50f66837eae03364R141-R148 <- this test specifically covers what happens when a logged out user hits this endpoint

jombooth (approved):

MihajloHoma (approved):

triklozoid (approved):

PR #5450, opened 2024-02-09
chore: LEAP-622: bump LaunchDarkly SDK

jombooth:
Bumping LD python SDK from 7.x.x to 8.2.1 per the advice in their 8->9 [migration guide](https://docs.launchdarkly.com/sdk/server-side/python/migration-8-to-9), which says "Before you migrate to version 9.0, we recommend updating to the latest 8.x version." This seems safer, and the EOL is within a couple months of the EOL of v9

image


jombooth:
closing in favor of #5451

PR #5452, opened 2024-02-10
Added input and classes to the response payload of get compatible projects API

forum-hs:
Added input and classes to the response payload of get compatible projects API and added tests

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit

### DIFF
--- a/label_studio/ml_models/api.py
+++ b/label_studio/ml_models/api.py
@@ -1,6 +1,5 @@
-import drf_yasg.openapi as openapi
+from core.label_config import get_all_labels
 from core.permissions import ViewClassPermission, all_permissions
-from django.conf import settings
 from django.utils.decorators import method_decorator
 from drf_yasg.utils import swagger_auto_schema
 from ml_models.models import ModelInterface
@@ -100,6 +99,28 @@ class ModelCompatibleProjects(generics.RetrieveAPIView):
                         if input.get('type', '') == 'Text':
                             return True
         return False
+    def build_inputs_payload(self, project):
+        parsed_config = project.get_parsed_config()
+        # using this to remove duplicate inputs 
+        input_map = {}
+        for tag in parsed_config:
+            for input in parsed_config[tag].get('inputs',[]):
+                input_type = input.get('type',None)
+                input_name = input.get('value', None)
+                if input_type and input_name and f'{input_type}_{input_name}' not in input_map:
+                    input_map[f'{input_type}_{input_name}'] = {
+                        'type': input_type,
+                        'name': input_name
+                    }
+        return list(input_map.values())
+
+    def build_classes_payload(self, project):
+        labels, _ = get_all_labels(project.label_config)
+        return sorted(list(set([
+                    label
+                    for label_list in labels.values()
+                    for label in label_list
+                ])))
     
     def get_queryset(self):
         return Project.objects.with_counts(fields = ['total_annotations_number']).filter(organization=self.request.user.active_organization, total_annotations_number__gt=1)
@@ -109,10 +130,14 @@ class ModelCompatibleProjects(generics.RetrieveAPIView):
         compatible_project_list = []
         for project in user_projects:
             if self._is_project_compatible(project=project):
+                inputs = self.build_inputs_payload(project=project)
+                classes = self.build_classes_payload(project=project)
                 compatible_project_list.append(
                     {
                         'title': project.title,
                         'id': project.id,
+                        'inputs': inputs,
+                        'classes' : classes
                     }
                 )
         result = {
--- a/label_studio/ml_models/api.py
+++ b/label_studio/ml_models/api.py
@@ -99,7 +99,7 @@ class ModelCompatibleProjects(generics.RetrieveAPIView):
                         if input.get('type', '') == 'Text':
                             return True
         return False
-    def build_inputs_payload(self, project):
+    def _build_inputs_payload(self, project):
         parsed_config = project.get_parsed_config()
         # using this to remove duplicate inputs 
         input_map = {}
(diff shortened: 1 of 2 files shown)


label_studio/ml_models/api.py
```diff
@@ -1,6 +1,5 @@
-import drf_yasg.openapi as openapi
```
pakelley, line 1: I take it these removed imports just weren't being used?

label_studio/ml_models/api.py
```diff
@@ -100,6 +99,28 @@ def _is_project_compatible(self, project):
                         if input.get('type', '') == 'Text':
                             return True
         return False
+    def _build_inputs_payload(self, project):
+        parsed_config = project.get_parsed_config()
+        # using this to remove duplicate inputs 
+        input_map = {}
+        for tag in parsed_config:
```
pakelley, line 106: Up to you what you think is most clear, but I think we can simplify this logic by reducing into a set, then just mapping back into a dict. Something like:
```python
unique_tags = {
    (("type", tag["type"]), ("value", tag["value"]))  # tuple of tuples, to make it hashable and still simple to map back to a dict
    for tag in parsed_config.values() 
    if ("type" in tag) and ("value" in tag)
}
return list(map(dict, unique_tags))
```

label_studio/ml_models/api.py
```diff
@@ -100,6 +99,28 @@ def _is_project_compatible(self, project):
...
+                if input_type and input_name and f'{input_type}_{input_name}' not in input_map:
+                    input_map[f'{input_type}_{input_name}'] = {
+                        'type': input_type,
+                        'name': input_name
+                    }
+        return list(input_map.values())
+
+    def _build_classes_payload(self, project):
+        labels, _ = get_all_labels(project.label_config)
+        return sorted(list(set([
```
pakelley, line 119: Why do we need this to be sorted?

label_studio/ml_models/api.py
pakelley: Ah, I missed one layer of complexity wrt the "inputs" key, I guess it would be more like:
```python
input_tags = [tag["inputs"] for tag in parsed_config.values() if "inputs" in tag]
unique_tags = { 
    (("type", tag["type"]), ("value", tag["value"])) # tuple of tuples, to make it hashable and still simple to map back to a dict 
    for tag in input_tags.values() 
    if ("type" in tag) and ("value" in tag) 
} 
return list(map(dict, unique_tags))
```
Again, whatever seems most clear to you, but imo this makes it very clear that we're explicitly filtering for input tags, imo. 
Also I figure both will run very fast, but if speed is a concern I'd expect these to be about the same. In yours we only loop once but in this I think we do less per iteration (i.e. less string construction and conditionals)

label_studio/ml_models/api.py
forum-hs: yes

label_studio/ml_models/api.py
forum-hs: I have seen this practice be used before when response contains list to retain consistent order, so that we do not change the order in every time we provide the response, will confuse user, we can either sort it in backend or frontend would do before displaying it, and it makes testing easier.

label_studio/ml_models/api.py
hakan458: This is like a leetcode answer haha I dont disagree its cleaner in a way, but less readable IMO 
I'll leave it up to you Forum

label_studio/ml_models/api.py
hakan458: For testing it doesnt matter we can specify that order does not matter 
For displaying consistent results to the use I agree it could be nice to see labels sorted, but Im also thinking that they might have it in their label config in the order that they want already (might not be alphabetical) and we would return it in the same order to them anyway each time

label_studio/ml_models/api.py
forum-hs: @pakelley - the code you wrote doesnt take into consideration that tag['inputs'] is of type list of dictionaries and not dictionary, so input_tags becomes a list of list of dictionaries, that is the reason I created an input map to collapse all the list into unique values. I would keep the code as is since that is more clear to me

label_studio/ml_models/api.py
forum-hs: So since I put it in a set the order would change as compared to label config on every call, are we okay with that?

hakan458: Thats true from a quick test it looks like `list(set())` is already sorting it, so the sorted call is redundant anyway, I think its ok not an important detail for now at least. Easy change in the future if we ever need

label_studio/ml_models/api.py
pakelley: lol I don't know if it's quite leetcode cryptic, but I don't mind either way 😁 
I'd prefer if we at least leave a comment here that we're filtering out non-input tags though, since it's pretty easy to miss this in the current implementation (imo). A brief docstring to this effect would also be fine too, if you'd prefer.

label_studio/ml_models/api.py
pakelley: Ah yeah, makes sense. Yeah, I'd prefer if we remove the `sorted` in this case, but it's not a blocker for me

pakelley (approved):
I'd like to get in the comment on filtering non-input tags and to remove the semi-redundant `sorted` call, but neither of these are blocking imo 👍

label_studio/ml_models/api.py
hakan458: What do you mean, isnt that the whole point, to get input tags? 
`for input in parsed_config[tag].get('inputs',[]):`

label_studio/ml_models/api.py
hakan458: Agree we should remove sorted

PR #5465, opened 2024-02-13
feat: DIA-851: Add ModelVersion

hakan458:

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/ml_models/api.py
```diff
@@ -78,53 +79,136 @@ def perform_create(self, serializer):
...
+        tags=['Models: Model Versions'],
+        operation_summary='List third-party model versions',
+        operation_description='List all third-party versions of a model.',
+        manual_parameters=[
+            openapi.Parameter(
+                name='model_id',
+                type=openapi.TYPE_INTEGER,
+                in_=openapi.IN_QUERY,
+                required=True,
+                description='A unique integer value identifying the model ID to list versions for.',
```
pakelley, line 95: Sorry if we've talked about this before, I can't remember: do we typically use an auto-incrementing int for ids? Would it be simple enough to use a uuid instead? The whole "try the integers less than the id of my latest ModelVersion" strategy is pretty simple to abuse, and it'd be nice to at least make it harder for people to uncover any permission issues we've missed.

label_studio/ml_models/api.py
```diff
@@ -78,53 +79,136 @@ def perform_create(self, serializer):
...
+    decorator=swagger_auto_schema(
+        tags=['Models: Model Versions'],
+        operation_summary='Delete third-party model version',
+        operation_description='Delete a third-party model version by ID',
+    ),
+)
+class ThirdPartyModelVersionAPI(viewsets.ModelViewSet):
+    serializer_class = ThirdPartyModelVersionSerializer
+    permission_required = ViewClassPermission(  # TODO use same permissions or different?
+        GET=all_permissions.model_interface_view,
```
pakelley: Just want to make sure we ask ourselves: would it make sense to rename these "model_interface_xxx" permissions to be more general, if we're using them for both model interfaces and model versions?

label_studio/ml_models/api.py
```diff
@@ -78,53 +79,136 @@ def perform_create(self, serializer):
...
+        POST=all_permissions.model_interface_create,
+    )
+
+    def get_queryset(self):
+        model_interface_pk = self.request.query_params.get('model_id')
+        if model_interface_pk is not None:
+            model_interface = generics.get_object_or_404(ModelInterface, pk=model_interface_pk)
+            model_versions = ThirdPartyModelVersion.objects.filter(parent_model=model_interface.id)
+        else:
+            model_versions = ThirdPartyModelVersion.objects.filter(organization=self.request.user.active_organization)
```
pakelley, line 158: what's the use case for filtering on org rather than model interface? (I'm probably just forgetting)

label_studio/ml_models/migrations/0002_thirdpartymodelversion.py
```diff
@@ -0,0 +1,35 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='ThirdPartyModelVersion',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('title', models.CharField(help_text='Model name', max_length=500, verbose_name='title')),
+                ('prompt', models.TextField(help_text='Prompt to execute', verbose_name='prompt')),
+                ('provider', models.CharField(choices=[('OpenAI', 'OpenAI')], default='OpenAI', help_text='The model provider to use e.g. OpenAI', max_length=255)),
```
pakelley, line 23: Why are there two "OpenAI"s here?

label_studio/ml_models/models.py
```diff
@@ -26,3 +27,51 @@ class ModelInterface(models.Model):
...
+
+class ModelVersion(models.Model):
+    class Meta:
+        abstract = True
+
+    title = models.CharField(_('title'), max_length=500, null=False, blank=False, help_text='Model name')
+
+    parent_model = models.ForeignKey(ModelInterface, related_name='model_versions', on_delete=models.CASCADE)
+
+    # TODO add field containing model run IDs
```
pakelley, line 40: Fyi I made sure this is captured in the implementation details for DIA-860

label_studio/ml_models/serializers.py
```diff
@@ -20,3 +22,36 @@ class Meta:
...
+    created_by = UserSimpleSerializer(default=CreatedByFromContext(), help_text='User who created Dataset')
+
+    class Meta:
+        model = ThirdPartyModelVersion
+        fields = '__all__'
+        read_only_fields = ['created_by', 'created_at', 'updated_at']
+
+    def validate(self, data):
+        third_party_model_version = self.instance
+        if third_party_model_version:
```
pakelley: If you're so inclined, this is basically the exact use case the "walrus operator" was made for:
```python
if third_party_model_version := self.instance:
    ...
```

label_studio/ml_models/serializers.py
```diff
@@ -20,3 +22,36 @@ class Meta:
...
+
+class ThirdPartyModelVersionSerializer(serializers.ModelSerializer):
+    created_by = UserSimpleSerializer(default=CreatedByFromContext(), help_text='User who created Dataset')
+
+    class Meta:
+        model = ThirdPartyModelVersion
+        fields = '__all__'
+        read_only_fields = ['created_by', 'created_at', 'updated_at']
+
+    def validate(self, data):
```
pakelley, line 35: Does the check that the parent model exists happen automagically by django, w/ the foreign key relationship? (didn't see anything explicitly set up in the model/migration, but could be missing something there too)

label_studio/ml_models/api.py
hakan458: We always use auto incrementing IDs yes, we can deal with permissions

label_studio/ml_models/api.py
hakan458: ? Thats exactly that theyre called now, do you mean "model_xxx" instead if we use them for both?

label_studio/ml_models/api.py
hakan458: This get_queryset function is used while listing, and also while getting a specific model version
In the list case, we pass parent_model param. In the get specific case, we are just getting it by the one ID, but that is retrieved from the output of this get_queryset function, so here all we do is filter by org for permissions reasons, then we get the specific id from there

label_studio/ml_models/migrations/0002_thirdpartymodelversion.py
hakan458: This is how choices are specified in django, one is the user facing value, one is the value stored in DB. For simplicity I used same value for both (defined in the class itself)

label_studio/ml_models/models.py
hakan458: Nice thank you

label_studio/ml_models/serializers.py
hakan458: Yes that is handled thanks to the ForeignKey field, and captured in one of the tests I added

label_studio/ml_models/serializers.py
hakan458: True true I will do that, not used to it still but its nice

label_studio/ml_models/api.py
pakelley: Gotcha, and is that just because it's simple? In your opinion, would it be worth considering the use of a UUID instead?

label_studio/ml_models/api.py
pakelley: Yeah sorry, weird sentence structure, but that's what I meant.

label_studio/ml_models/api.py
pakelley: Mind leaving a comment that we're only retrieving ModelVersions the user has access to? (If this is a very very common pattern, don't worry about it though)

label_studio/ml_models/api.py
hakan458: Dont think its worth it no Id like to be consistent with every other django model in LS

label_studio/ml_models/api.py
```diff
@@ -78,53 +79,136 @@ def perform_create(self, serializer):
...
+        PATCH=all_permissions.model_interface_change,
+        PUT=all_permissions.model_interface_change,
+        POST=all_permissions.model_interface_create,
+    )
+
+    def get_queryset(self):
+        model_interface_pk = self.request.query_params.get('model_id')
+        if model_interface_pk is not None:
+            model_interface = generics.get_object_or_404(ModelInterface, pk=model_interface_pk)
+            model_versions = ThirdPartyModelVersion.objects.filter(parent_model=model_interface.id)
```
forum-hs: Even if the model_interface_pk is not None should we not have filter by org for permissions reason, user could invoke a api call with another orgs model_interface_pk ?

label_studio/ml_models/api.py
hakan458: done

pakelley (approved):

forum-hs (approved):

PR #5480, opened 2024-02-17
docs: DOC-189: Clarify adding tags in example for creating regions without labels

caitlinwheeless:
The "Create regions without labels" section needed to refer to the earlier tag example, or it wouldn't work. 

This PR affects:

- [X] Community docs
- [X] Enterprise docs

### DIFF
--- a/docs/source/guide/labeling.md
+++ b/docs/source/guide/labeling.md
@@ -251,20 +251,20 @@ Create a custom template for your labeling interface using the following example
 
 ```xml
 <View>
-    <Image name="image" value="$image" />
-    <Rectangle name="rect" toName="image" />
-    <Ellipse name="ellipse" toName="image" />
-    <KeyPoint name="kp" toName="image" />
-    <Polygon name="polygon" toName="image" />
-    <Brush name="brush" toName="image" />
-    <Choices name="choices" toName="image">
-        <Choice value="yes"></Choice>
-        <Choice value="no"></Choice>
-    </Choices>
-    <Labels name="labels" toName="image" fillOpacity="0.5" strokeWidth="5">
-        <Label value="building" background="green"></Label>
-        <Label value="vehicle" background="blue"></Label>
-    </Labels>
+  <Image name="image" value="$image" />
+  <Rectangle name="rect" toName="image" />
+  <Ellipse name="ellipse" toName="image" />
+  <KeyPoint name="kp" toName="image" />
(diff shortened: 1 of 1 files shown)


docs/source/guide/labeling.md
```diff
@@ -248,25 +248,25 @@ You can add multiple types of regions to image annotations. You can add any of t
...
-  <Brush name="brush" toName="image" />
-  <Choices name="choices" toName="image">
-    <Choice value="yes"></Choice>
-    <Choice value="no"></Choice>
-  </Choices>
-<Labels name="labels" toName="image" fillOpacity="0.5" strokeWidth="5">
-  <Label value="building" background="green"></Label>
-  <Label value="vehicle" background="blue"></Label>
-  </Labels>
+    <Image name="image" value="$image" />
```
hlomzik: do we use 4 spaces indents in other places? I believe it's 2 everywhere

docs/source/guide/labeling.md
```diff
@@ -277,21 +277,19 @@ If you accidentally select a point on an image while creating a polygon, just do
 ### Create regions without labels
 When you're annotating images, you can create regions without applying labels. 
 
-1. Create a region by double-clicking or clicking and dragging to create a bounding box, or click the points necessary to construct a polygon.
-2. Select the created region in the sidebar or on the image.
-3. Select the label that you want to apply to the region.
-4. Repeat these steps for any regions that you want to create.
+1. Create a custom template using the example provided above under [Add multiple types of regions to image annotations](#Add-multiple-types-of-regions-to-image-annotations). 
+2. Create a region by double-clicking or clicking and dragging to create a bounding box, or click the points necessary to construct a polygon.
```
hlomzik: The very first time the tool should be selected in toolbar, and after this in any other project and task it will keep selection from previous tasks. But it's worth to mention to clarify it for good.

hlomzik (changes requested):
Small comments only

docs/source/guide/labeling.md
caitlinwheeless: Updated to use two spaces instead of 4

docs/source/guide/labeling.md
caitlinwheeless: Added:

> 2. Select which tool you want to use in the labeling toolbar. You only need to do this for the first task.

hlomzik (approved):

PR #5493, opened 2024-02-22
[temporary pull req] Adding additional fields to model interface parent model run django model

forum-hs:
https://humansignal.atlassian.net/browse/DIA-950

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit


label_studio/ml_models/api.py
```diff
@@ -69,17 +70,49 @@ class ModelInterfaceAPI(viewsets.ModelViewSet):
...
 
         # we need to save these fields for faster access and filters without excess joins
         serializer.validated_data['organization'] = self.request.user.active_organization
         serializer.validated_data['created_by'] = self.request.user
         serializer.save()
+        instance = serializer.instance
+       
+        model_interface = ModelInterface.objects.filter(pk=instance.pk)[0]
+        for id in associated_projects_data:
+            model_interface.associated_projects.add(id)
```
pakelley, line 112: nit: looks like we're missing some spacing here, mind running the formatter? (i.e. blue/ruff)

label_studio/ml_models/api.py
```diff
@@ -69,17 +70,49 @@ class ModelInterfaceAPI(viewsets.ModelViewSet):
...
         serializer.is_valid(raise_exception=True)
 
         # we need to save these fields for faster access and filters without excess joins
         serializer.validated_data['organization'] = self.request.user.active_organization
         serializer.validated_data['created_by'] = self.request.user
         serializer.save()
+        instance = serializer.instance
+       
+        model_interface = ModelInterface.objects.filter(pk=instance.pk)[0]
+        for id in associated_projects_data:
```
pakelley: nit: if this is just a list of ids, mind if we name the var `associated_project_ids`?

label_studio/ml_models/models.py
```diff
@@ -26,6 +35,12 @@ class ModelInterface(models.Model):
         'organizations.Organization', on_delete=models.CASCADE, related_name='model_interfaces', null=True
     )
 
+    input_fields = models.JSONField(default=list, validators=[validate_string_list])
+
+    output_classes = models.JSONField(default=list, validators=[validate_string_list])
+
+    associated_projects = models.ManyToManyField('projects.Project', blank=True)
```
pakelley, line 43: Why is this allowed to be blank? Seems like we end up having to validate it later too, is this some weirdness with many-to-many fields or something?

label_studio/ml_models/models.py
```diff
@@ -26,6 +35,12 @@ class ModelInterface(models.Model):
         'organizations.Organization', on_delete=models.CASCADE, related_name='model_interfaces', null=True
     )
 
+    input_fields = models.JSONField(default=list, validators=[validate_string_list])
```
pakelley, line 39: Do we need a default here? Seems like this should be a required field, and not have a default?

label_studio/ml_models/models.py
```diff
@@ -6,6 +6,15 @@
 from django.utils.translation import gettext_lazy as _
 from ml_model_providers.models import ModelProviderConnection
 from projects.models import Project
+from rest_framework.exceptions import ValidationError
+
+def validate_string_list(value):
```
pakelley, line 12: Mind putting type definitions, and maybe a docstring here? It wasn't obvious what's supposed to be inside `value` to me.

label_studio/ml_models/models.py
```diff
@@ -6,6 +6,15 @@
...
+    if not value:
+        raise ValidationError('list should not be empty')
+    if not isinstance(value, list):
+        raise ValidationError('Value must be a list')
+    if not all(isinstance(item, str) for item in value):
```
pakelley, line 17: If we're validating that `output_classes` and `input_fields` are lists of strings, why do we store them as a `JSONField`? Is this just the closest type we can get to a list of strings?

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+ 
+    class Meta:
+        model = ModelInterface
+        fields = '__all__'
+        read_only_fields = ['created_by', 'created_at', 'updated_at']
+
+    def check_output_classes(self, project, provided_output_classes):
+        labels, _ = get_all_labels(project.label_config)
+        project_output_classes =  sorted(list(set([label for label_list in labels.values() for label in label_list])))
+        print(project_output_classes)
```
pakelley: nit: let's remove these print statements

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+        print(provided_output_classes)
+        if project_output_classes != provided_output_classes:
+            raise ValidationError(f'output_classes not compatible with Project (id:{project.pk})')
```
pakelley: Mind adding both the `provided_output_classes` and the `project_output_classes` to the error message here?

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+    def check_input_fields(self, project, provided_input_fields):
+        parsed_config = project.get_parsed_config()
+        project_input_fields = set()
+        for tag in parsed_config:
+            for input in parsed_config[tag]['inputs']:
+                if input.get('type',None) and input.get('value', None):
+                    project_input_fields.add(input.get('value'))
+        
+        project_input_fields = sorted(list(project_input_fields))
+        print(project_input_fields, provided_input_fields)
```
pakelley: let's remove this print statement too

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+        if project_input_fields != provided_input_fields:
+            raise ValidationError(f'input_fields do not match inputs in Project (id:{project.pk})')
+        
+    def validate(self, data):        
+        model_obj = getattr(self, 'instance', None)
+        if model_obj:
+            if 'input_fields' in data:
```
pakelley, line 50: `input_fields` and `output_classes` should be required, right? So I figure we can either just get rid of these `if`s, or guard for this case by raising an exception if `input_fields` or `output_classes` are blank.

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+                provided_input_fields = sorted(data['input_fields'])
```
pakelley, line 51: If we're just sorting these for the equality check, can we just put them in a `set` instead? Imo that makes what we're doing slightly more clear, plus should be a bit faster (not that I expect this to be a bottleneck)

label_studio/ml_models/serializers.py
```diff
@@ -16,19 +17,103 @@ class CreatedByFromContext:
...
+            else:
+                if 'input_fields' in data:
+                    for associated_proj in model_obj_associated_projects:
+                        self.check_input_fields(project=associated_proj, provided_input_fields=provided_input_fields)
+                if 'output_classes' in data:
+                    for associated_proj in model_obj_associated_projects:
+                        self.check_output_classes(project=associated_proj, provided_output_classes=provided_output_classes)
+            return data
+        
+class ModelInterfaceCreateSerializer(ModelInterfaceSerializer):
```
pakelley, line 83: It seems like the main reason for having a separate serializer here is to have a simpler validation function, is that correct?
I want to say we can get by with just using the regular validation on create, but maybe I'm missing something?

label_studio/ml_models/api.py
forum-hs: okay

label_studio/ml_models/api.py
forum-hs: will do

label_studio/ml_models/models.py
forum-hs: it is list of string which is being validated for type, incase the value sent is list of integers we will error out. I can add type as list

label_studio/ml_models/models.py
forum-hs: in django that is how list of string is being stored based what I have read.

label_studio/ml_models/models.py
forum-hs: yes that is what indicates to django that this is of list type.

label_studio/ml_models/models.py
forum-hs: yes, because we cannot add the manytomany field when we are creating the object and it would error out if it were False. That was the issue I faced and hence I added a check to ensure user provides us the project without it we will not build the modelrun object

label_studio/ml_models/serializers.py
forum-hs: yes sorry about that I thought I had deleted all of them

label_studio/ml_models/serializers.py
forum-hs: sure

label_studio/ml_models/serializers.py
forum-hs: yes absolutely sorry about that

label_studio/ml_models/serializers.py
forum-hs: we will need to validate during update as well and that is why we have 2 serializers, since for create all the fields are a must but for update only some of them need to be validated

label_studio/ml_models/api.py
```diff
@@ -268,7 +308,6 @@ def perform_create(self, serializer):
         serializer.validated_data['created_by'] = self.request.user
         serializer.validated_data['triggered_at'] = datetime.utcnow()
         serializer.validated_data['organization'] = self.request.user.active_organization
-        print(serializer.validated_data)
```
forum-hs, line 271: This print wont be present in the lso lse split PR

label_studio/ml_models/api.py
forum-hs: done

label_studio/ml_models/serializers.py
forum-hs: done

PR #5562, opened 2024-03-11
feat: LEAP-818: Add X-Api-Key support

### LINKED ISSUE #3349: Authorization Token collision with Google IAP (Identity-Aware Proxy)
**Is your feature request related to a problem? Please describe.**
When Label Studio is deployed to Google Cloud Run with IAP (https://cloud.google.com/iap) active, I see a 302 redirect when I make an API call:

```
curl -X GET http://[MY_CLOUD_RUN_DOMAIN_NAME]/api/projects/ -H 'Authorization: Token <TOKEN>'
```

This is because Google's https://cloud.google.com/iap asks me to log in with my Google account first before I can hit the Label Studio API. 

I can bypass it by passing in a `GCP_IAAP_AUTH_TOKEN_[STRING]` cookie using Thunder Client or Postman, but I don't know of a way to pass a cookie to the Label Studio SDK (https://github.com/heartexlabs/label-studio-sdk) if I'm hitting Label Studio via another service/application.

Three potential solutions:

1) Turn off Google IAP (this may not be feasible depending on product requirements)

2) Figure out a way to pass the `GCP_IAAP_AUTH_TOKEN_[STRING]` cookie to the Label Studio SDK (somehow??)

3) Fork and make a PR to the Label Studio repo so that it accepts `X-Api_Key` as an alternative to `Authorization: Token`. 

There's a possibility this third solution can be implemented via the third-party `djangorestframework-api-key` library at https://florimondmanca.github.io/djangorestframework-api-key/guide/#custom-header.

In other words, currently this is supported:
```
curl -X GET http://[MY_CLOUD_RUN_DOMAIN_NAME]/api/projects/ -H 'Authorization: Token <TOKEN>'
```
We want to be able to support the following as well:
```
curl -X GET http://[MY_CLOUD_RUN_DOMAIN_NAME]/api/projects/ -H 'X-Api-Key: <TOKEN>'
```

**Describe alternatives you've considered**
See above. This feature request focuses on the third solution. 

**Additional context**
1) Slightly related: GCP_IAAP_AUTH_TOKEN Cookies Without Suffix Ignored 
 https://issuetracker.google.com/issues/215575439?pli=1
 
2) The reason the https://florimondmanca.github.io/djangorestframework-api-key/guide/#custom-header looks like it might be a valid solution:

image

jombooth:
If X-API-Key header is specified with `<token>`, overwrite the `Authorization` header with `Token <token>`.

Resolves https://github.com/HumanSignal/label-studio/issues/3349

```
~/Repos/label-studio on  fb-LEAP-818/x-api-key [$?]
❯ curl -X GET http://localhost:8080/api/projects/ -H 'X-API-Key: [redacted]'
{"count":27,"next":null,"previous":null,"results":[...]}
```

`HttpRequest.META` documentation for reference: https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.META

### DIFF
--- a/label_studio/core/middleware.py
+++ b/label_studio/core/middleware.py
@@ -159,23 +159,6 @@ class DatabaseIsLockedRetryMiddleware(CommonMiddleware):
         return response
 
 
-class SimpleMiddleware:
-    def __init__(self, get_response):
-        self.get_response = get_response
-        # One-time configuration and initialization.
-
-    def __call__(self, request):
-        # Code to be executed for each request before
-        # the view (and later middleware) are called.
-
-        response = self.get_response(request)
-
-        # Code to be executed for each request/response after
-        # the view is called.
-
-        return response
-
-
 class XApiKeySupportMiddleware:
     """Middleware that adds support for the X-Api-Key header, by having its value supersede
     anything that's set in the Authorization header."""
--- a/label_studio/core/middleware.py
+++ b/label_studio/core/middleware.py
@@ -159,7 +159,7 @@ class DatabaseIsLockedRetryMiddleware(CommonMiddleware):
         return response
 
 
-class XApiKeySupportMiddleware(CommonMiddleware):
+class XApiKeySupportMiddleware(MiddlewareMixin):
     """Middleware that adds support for the X-Api-Key header, by having its value supersede
     anything that's set in the Authorization header."""
 
--- a/label_studio/core/middleware.py
+++ b/label_studio/core/middleware.py
@@ -159,15 +159,36 @@ class DatabaseIsLockedRetryMiddleware(CommonMiddleware):
         return response
 
 
-class XApiKeySupportMiddleware(MiddlewareMixin):
+class SimpleMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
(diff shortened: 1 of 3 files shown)


label_studio/core/middleware.py
```diff
@@ -159,6 +159,21 @@ def __call__(self, request):
...
+    """Middleware that adds support for the X-Api-Key header, by having its value supersede
+    anything that's set in the Authorization header."""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if 'HTTP_X_API_KEY' in request.META:
+            request.META['HTTP_AUTHORIZATION'] = f'Token {request.META["HTTP_X_API_KEY"]}'
+            del request.META['HTTP_X_API_KEY']
```
niklub, line 172: why should we remove it from the payload?

jombooth: It isn't used otherwise, and I just wanted to make sure that we won't have a situation where different auth headers are read in different spots on the same request. This is also why this middleware is set to run right before AuthenticationMiddleware

niklub (approved):

PR #5567, opened 2024-03-12
feat: DIA-940: Read Adala predictions back in once they are completed

matt-bernstein:
Small db changes to support LSE PR, see that one for details.

Related PRs:
- https://github.com/HumanSignal/Adala/pull/62 rework some Adala API endpoints for clarity, and statuses to match LS better
- https://github.com/HumanSignal/label-studio-enterprise/pull/5391

[DIA-939]: https://humansignal.atlassian.net/browse/DIA-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


label_studio/ml_models/models.py
```diff
@@ -142,3 +142,10 @@ def input_file_name(self):
     @property
     def output_file_name(self):
         return f'{self.project.id}_{self.model_version.pk}_{self.pk}/output_tasks.csv'
+
+    @property
+    def error_file_name(self):
+        return f'{self.project.id}_{self.model_version.pk}_{self.pk}/output_errors.csv'
+
+    # TODO should this be a CharField in case Adala changes its job ID format?
```
pakelley: what does uuidfield give us that charfield doesn't? If nothing valuable, seems worth switching

matt-bernstein: the main thing it gives is validation. eg if you pass it in as a string with or without dashes, with or without the header, it will resolve to the same UUID. Seems worthwhile if we will stick with that format.

matt-bernstein:
will add the Prediction modelrun field in a sec, have to deconflict with the other updates to develop today

matt-bernstein (commented):
not sure this will always be true, but seems reasonable to me for now

label_studio/tasks/models.py
```diff
@@ -976,6 +984,13 @@ def save(self, *args, **kwargs):
             logger.warning('project_id is not set for prediction, project_id being set in save method')
             self.project_id = Task.objects.only('project_id').get(pk=self.task_id).project_id
 
+        # make sure the source of this prediction is set correctly
+        if self.model is not None and self.model_run is not None:
+            logger.error(
+                'Only one of model or model_run should be set in prediction, '
+                'it cannot come from both an MLBackend and a ModelRun.',
+            )
+
```
matt-bernstein: Not sure this will always be true, but seems reasonable to me for now.

forum-hs: Why would this be true? model and model run are different entities right?

label_studio/ml_models/models.py
```diff
@@ -142,3 +144,10 @@ def input_file_name(self):
     @property
     def output_file_name(self):
         return f'{self.project.id}_{self.model_version.pk}_{self.pk}/output_tasks.csv'
+
+    @property
+    def error_file_name(self):
```
hakan458: I added this in 939 PR as well

label_studio/ml_models/models.py
hakan458: I added this in 939 PR as well, and made it a charfield 
I figured we will have more job IDs than just Adala's in the future and it should be flexible

label_studio/tasks/models.py
hakan458: Yeah I dont think it will ever happen, but maybe a good check to keep because we might accidentally set `model` instead of `model_version` :D

matt-bernstein:
rebased on DIA-939 and deconflicted duplicate changes, but this is still NOT mergeable, because it includes conflicting changes to `Prediction` model that were reverted in d1021d643ff33e039505bae56999c68dbfabfecf on develop - need to figure out what the status of those changes is

label_studio/ml_models/models.py
matt-bernstein: changed to charfield

label_studio/tasks/models.py
matt-bernstein: will have to figure out what to do with this now that the previous change is reverted in develop

matt-bernstein: reverted this check until HUMSIG-37 gets remerged

hakan458 (approved):

PR #5570, opened 2024-03-12
feat: HUMSIG-37: ML backend experience updates

niklub:
- Support for HTTP Basic Authentication in ML Backend and Label Studio
- Capability to transfer additional JSON params from Label Studio to the ML Backend
- Enhanced configuration method for selecting predictions or ML Models in Label Studio for prelabelling
- Project settings featuring a list of predictions with options to remove them
- Standard screens for scenarios without connected models or uploaded predictions

## Old interface
image
image

## New interface
image
image


label_studio/data_manager/actions/basic.py
```diff
@@ -138,6 +138,7 @@ def async_project_summary_recalculation(tasks_ids_list, project_id):
         'title': 'Retrieve Predictions',
         'order': 90,
         'dialog': {
+            'modal_title': 'Retrieve Predictions',
```
bmartel:
```suggestion
            'title': 'Retrieve Predictions',
```

label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -95,8 +95,10 @@ def predictions_to_annotations_form(user, project):
         'title': 'Create Annotations From Predictions',
         'order': 91,
         'dialog': {
-            'text': 'This action will create new annotations from predictions with the selected model version '
-            'for each selected task.',
+            'modal_title': 'Create Annotations From Predictions',
```
bmartel:
```suggestion
            'title': 'Create Annotations From Predictions',
```

label_studio/tasks/api.py
```diff
@@ -223,20 +225,21 @@ def get_serializer_class(self):
...
 
-        result = self.get_serializer(task).data
+    #     result = self.get_serializer(task).data
 
-        # use proxy inlining to task data (for credential access)
-        result['data'] = task.resolve_uri(result['data'], project)
-        return Response(result)
+    #     # use proxy inlining to task data (for credential access)
+    #     result['data'] = task.resolve_uri(result['data'], project)
+    #     return Response(result)
```
jombooth: why is this commented out? If it's no longer needed, let's delete it. But we should understand what it was doing before

label_studio/tasks/api.py
```diff
@@ -181,18 +180,21 @@ def get_retrieve_serializer_context(self, request):
 
     def get(self, request, pk):
         context = self.get_retrieve_serializer_context(request)
-        context['project'] = project = self.task.project
+        context['project'] = self.task.project
 
         # get prediction
-        if (
-            project.evaluate_predictions_automatically or project.show_collab_predictions
-        ) and not self.task.predictions.exists():
```
jombooth: Why do we no longer check for `project.evaluate_predictions_automatically`?

label_studio/tasks/api.py
```diff
@@ -181,18 +180,21 @@ def get_retrieve_serializer_context(self, request):
...
-            evaluate_predictions([self.task])
+        if self.task.project.show_collab_predictions:
+            self.task.refresh_predictions()
             self.task.refresh_from_db()
 
+        # predictions = retrieve_predictions([task])
+        # if predictions:
+        #     self.task.refresh_from_db()
```
jombooth: this looks like it can be deleted

.pre-commit-dev.yaml
```diff
@@ -9,3 +9,4 @@ repos:
     rev: v0.9.1
     hooks:
       - id: blue
+        args: [ --verbose ]
```
jombooth: please revert this, it spams my terminal when I run `make fmt-all`

label_studio/data_manager/api.py
```diff
@@ -255,11 +255,15 @@ def get(self, request):
...
+
             # retrieve ML predictions if tasks don't have them
-            if not review and project.evaluate_predictions_automatically:
-                tasks_for_predictions = Task.objects.filter(id__in=ids, predictions__isnull=True)
-                evaluate_predictions(tasks_for_predictions)
-                [tasks_by_ids[_id].refresh_from_db() for _id in ids]
+            # if not review and project.retrieve_predictions_automatically:
+            #     tasks_for_predictions = Task.objects.filter(id__in=ids, predictions__isnull=True)
+            #     retrieve_predictions(tasks_for_predictions)
+            #     [tasks_by_ids[_id].refresh_from_db() for _id in ids]
```
jombooth: What's the outcome of the discussion here? looks like another place where we used to consider `project.retrieve_predictions_automatically` but don't anymore. In the final version, we shouldn't have commented out code here.

label_studio/data_manager/api.py
```diff
@@ -271,13 +275,18 @@ def get(self, request):
             else:
                 serializer = self.task_serializer_class(page, many=True, context=context)
             return self.get_paginated_response(serializer.data)
+
+        # TODO
         # all tasks
-        if project.evaluate_predictions_automatically:
-            evaluate_predictions(queryset.filter(predictions__isnull=True))
+        # if project.retrieve_predictions_automatically:
+        #     retrieve_predictions(queryset.filter(predictions__isnull=True))
```
jombooth: Similarly, we shouldn't merge commented out code here. What's the intent of this part of the change?

label_studio/ml/api.py
```diff
@@ -227,29 +281,64 @@ def post(self, request, *args, **kwargs):
     ),
 )
 class MLBackendInteractiveAnnotating(APIView):
+    """ """
```
jombooth: We can delete this, or better yet, add a docstring :)

jombooth: (applies to all the `""" """` docstrings below also)

label_studio/tests/predictions/test_predictions.py
```diff
@@ -769,6 +769,8 @@ def test_interactive_annotating_failing(business_client, configured_project):
 
         result = r.json()
 
+    print(result)
+
```
jombooth: please remove this print

label_studio/data_manager/functions.py
```diff
@@ -314,16 +314,19 @@ def get_prepared_queryset(request, project):
...
-    """Call ML backend for prediction evaluation of the task queryset"""
+def retrieve_predictions(tasks, backend=None):
+    """Call ML backend to retrieve predictions with the task queryset as an input"""
     if not tasks:
         return
 
-    project = tasks[0].project
+    if not backend:
+        project = tasks[0].project
+        backend = project.ml_backends.first()
```
jombooth: this is a small table so it likely doesn't matter, but it's generally better to use `fast_first` if you don't care about ordering

label_studio/data_manager/functions.py
```diff
@@ -314,16 +314,19 @@ def get_prepared_queryset(request, project):
...
 
-    for ml_backend in project.ml_backends.all():
-        # tasks = tasks.filter(~Q(predictions__model_version=ml_backend.model_version))
-        ml_backend.predict_tasks(tasks)
+    # IMPORTANT change here, ml_backends.all => ml_backends.first
+    # we are using only one ML backend, not multiple
```
jombooth: I think this belongs in the docstring. What I'd recommend:

```
"""
Call the given ML backend to retrieve predictions with the task queryset as an input.
If backend is not specified, we'll assume the tasks' project only has one associated
ML backend, and use that backend.
"""
```

label_studio/ml/api.py
```diff
@@ -202,6 +204,58 @@ def post(self, request, *args, **kwargs):
...
+        random = request.query_params.get('random', False)
+        if random:
+            task = Task.get_random(project=ml_backend.project)
+            if not task:
+                raise Http404
+
+            kwargs = ml_backend._predict(task)
+            return Response(**kwargs)
+
+        # TODO this needs to be implemented and needs to have a specific task param
```
jombooth: Is this still a TODO? I can't really tell what's meant by the "this"

label_studio/tasks/models.py
```diff
@@ -178,6 +179,20 @@ class Meta:
     def file_upload_name(self):
         return os.path.basename(self.file_upload.file.name)
 
+    @classmethod
+    def get_random(cls, project):
+        """Get random task from a project, this should not be used lightly as its expensive method to run"""
+        if project is None:
+            raise Exception('Project id is required')
```
jombooth: The method shouldn't raise an exception for this. `project` should be a required argument, and there shouldn't be any nullchecking.

label_studio/ml/models.py
```diff
@@ -391,6 +529,16 @@ def _validate_ml_api_result(ml_api_result, tasks, curr_logger):
     return True
 
 
+@receiver(pre_delete, sender=MLBackend)
+def modify_project_model_version(sender, instance, **kwargs):
+    """ """
```
jombooth: We generally discourage the use of Django signals because they aren't called under various conditions, eg, a queryset-level delete. why is this necessary? A docstring would really help.

label_studio/ml/models.py
```diff
@@ -391,6 +529,16 @@ def _validate_ml_api_result(ml_api_result, tasks, curr_logger):
...
+    project = instance.project
+
+    if project.model_version == instance.title:
+        project.model_version = None
+        project.save()
```
jombooth: please use update_fields to avoid the potential for clobbering other updates to `project`. like `project.save(update_fields=['model_version'])`

label_studio/ml/models.py
```diff
@@ -87,45 +123,110 @@ class MLBackend(models.Model):
...
+        If the title of the model instance is changed and the model_version
+        of the related project is currently the same as the original title,
+        the project's model_version is updated to the new title.
+        """
+        p = self.project
+
+        if self.title != self.__original_title and p.model_version == self.__original_title:
+            with transaction.atomic():
+                p.model_version = self.title
+                p.save()
```
jombooth: this should also use update_fields as recommended in https://github.com/HumanSignal/label-studio/pull/5570/files#r1522245983

label_studio/ml/models.py
```diff
@@ -87,45 +123,110 @@ class MLBackend(models.Model):
...
+        self.__original_title = self.title
+
+    def save(self, *args, **kwargs):
+        """
+        Overrides the save() method to update the associated project's model_version field.
+
+        If the title of the model instance is changed and the model_version
+        of the related project is currently the same as the original title,
+        the project's model_version is updated to the new title.
+        """
```
jombooth: It's really better to avoid custom `save()` methods if at all possible. Is there a different way we could do this? For example, could we just make the project's model_version optional and defer to the title of the model instance if it is unset?

label_studio/ml/models.py
```diff
@@ -87,45 +123,110 @@ class MLBackend(models.Model):
...
+        return self.get_versions_(
+            self.url,
+            self.project,
+            self.auth_method,
+            basic_auth_user=self.basic_auth_user,
+            basic_auth_pass=self.basic_auth_pass,
+        )
+
     def update_state(self):
+        """ """
```
jombooth: nit: please add a docstring or remove this

label_studio/ml/models.py
```diff
@@ -30,6 +30,11 @@ class MLBackendState(models.TextChoices):
     PREDICTING = 'PR', _('Predicting')
 
 
+class MLBackendAuth(models.TextChoices):
+    NONE = 'NA', _('None')
+    BASIC_AUTH = 'BA', _('Basic Auth')
```
jombooth: We save almost nothing by using two-letter codes in this charfield, why not make the CharField len 255 and make the enum values the same as the enum name? This means nobody ever has to figure out that "NA" is the same as NONE by reading backend code, for example.

label_studio/ml/models.py
```diff
@@ -60,20 +65,51 @@ class MLBackend(models.Model):
...
+        max_length=2,
+        choices=MLBackendAuth.choices,
+        default=MLBackendAuth.NONE,
+    )
+
+    basic_auth_user = models.TextField(
+        _('basic auth user'),
+        blank=True,
+        null=True,
+        default='',
```
jombooth, line 79: On recent versions of postgres, adding a column with a simple default is an efficient operation. I would have preferred specifying either `null=True` or `default=''` but not both; however I recognize the ship may have already sailed on this since it was deployed on a production database.

label_studio/ml/serializers.py
```diff
@@ -2,50 +2,119 @@
...
+            'id',
+            'state',
+            'readable_state',
+            'is_interactive',
+            'url',
+            'error_message',
+            'title',
+            'auth_method',
+            'basic_auth_user',
+            'basic_auth_pass',
```
jombooth, line 101: We need to make sure that these fields are writable but not readable in the API.

label_studio/projects/models.py
```diff
@@ -259,12 +271,17 @@ class SkipQueue(models.TextChoices):
...
+        """ """
+
         super(Project, self).__init__(*args, **kwargs)
         self.__original_label_config = self.label_config
         self.__maximum_annotations = self.maximum_annotations
         self.__overlap_cohort_percentage = self.overlap_cohort_percentage
         self.__skip_queue = self.skip_queue
 
+        # if rpa is not None:
+        #     self.evaluate_predictions_automatically = rpa
```
jombooth: need to remove this too

label_studio/projects/models.py
```diff
@@ -276,18 +293,19 @@ def __init__(self, *args, **kwargs):
...
-        if flag_set(
-            'fflag_perf_back_lsdv_4695_update_prediction_query_to_use_direct_project_relation',
-            user='auto',
-        ):
-            return Prediction.objects.filter(Q(project=self.id) & Q(model_version=self.model_version))
-        else:
-            return Prediction.objects.filter(Q(task__project=self.id) & Q(model_version=self.model_version))
+    @property
+    def ml_backend(self):
+        return self.ml_backends.first()
```
jombooth: please use `fast_first` instead of `first()`

label_studio/projects/models.py
```diff
@@ -644,16 +658,41 @@ def display_count(count: int, type: str) -> Optional[str]:
...
+        with transaction.atomic():
+            # If we are deleting specific model_version then we need
+            # to remove that from the project
+            if self.should_none_model_version(model_version):
+                self.model_version = None
+                self.save()
+
+            predictions.delete()
+
+        ## TODO wtf is that returning
```
jombooth: We should remove this comment

label_studio/projects/models.py
```diff
@@ -644,16 +658,41 @@ def display_count(count: int, type: str) -> Optional[str]:
...
+        predictions = Prediction.objects.filter(**params)
         count = predictions.count()
-        predictions.delete()
+
+        with transaction.atomic():
+            # If we are deleting specific model_version then we need
+            # to remove that from the project
+            if self.should_none_model_version(model_version):
+                self.model_version = None
+                self.save()
```
jombooth: we should be using update_fields here too

label_studio/projects/models.py
```diff
@@ -644,16 +658,41 @@ def display_count(count: int, type: str) -> Optional[str]:
...
+
+            predictions.delete()
```
jombooth: optional: I believe the delete() method returns the number of rows that it deleted, so we can ditch the count() query

label_studio/projects/serializers.py
```diff
@@ -166,6 +167,36 @@ def validate_label_config(self, value):
...
+            else:
+                raise serializers.ValidationError(
+                    "Model version doesn't exist either as live model or as static predictions."
+                )
+
+        return value
+
+    def update(self, instance, validated_data):
+        """ """
+        # assume there is a flag, 'is_name_updated', in model to track if a name has been updated
```
jombooth: This comment seems to be outdated, or is `'show_collab_predictions'` wrong below?

label_studio/tests/ml/test_models.py
```diff
@@ -0,0 +1 @@
+
```
jombooth: please remove this and other empty files from the diff

label_studio/projects/api.py
```diff
@@ -640,5 +642,34 @@ class ProjectModelVersions(generics.RetrieveAPIView):
     queryset = Project.objects.all()
 
     def get(self, request, *args, **kwargs):
+        # TODO make sure "extended" is the right word and is
+        # consistent with other APIs we've got
```
jombooth, line 646: I think this should be addressed before merging

label_studio/ml/serializers.py
jombooth: ok, I do see this is write_only: https://github.com/HumanSignal/label-studio/pull/5570/files#diff-a67b82b0de34ce236f0f9a212e530d2a865c73f603e0e5e6ca485d5c0ea45d10R13

would be great to capture this in a test though.

label_studio/ml/api_connector.py
```diff
@@ -31,18 +31,37 @@
...
+    def __init__(
+        self, url, timeout=None, connection_timeout=None, max_retries=None, headers=None, auth_method=None, **kwargs
+    ):
         self._url = url
         self._timeout = timeout or TIMEOUT_DEFAULT
         self._connection_timeout = connection_timeout or CONNECTION_TIMEOUT
         self._headers = headers or {}
+        self._auth_method = auth_method
+
+        if auth_method == 'BA':
```
jombooth: Here and everywhere in the PR: we shouldn't refer to strings from an enum explicitly. Instead, let's compare `auth_method` with `MLBackendAuth.BASIC_AUTH`

Ideally we would use an enum on the frontend too for the same reason.

label_studio/projects/models.py
```diff
@@ -644,16 +658,41 @@ def display_count(count: int, type: str) -> Optional[str]:
...
-        else:
-            predictions = Prediction.objects.filter(task__project=self)
+    def should_none_model_version(self, model_version):
+        """
+        Returns True if the model version provided matches the object's model version,
+        or no model version is set for the object but model version exists in ML backend.
+        """
+        return self.model_version == model_version or self.ml_backend_in_model_version
+
+    def delete_predictions(self, model_version=None, *args, **kwargs):
```
jombooth: I don't see `*args` or `**kwargs` being used anywhere

label_studio/tasks/api.py
jombooth: I have since learned that it was deprecated as part of this work - we should add this fact to the PR description

label_studio/ml/models.py
```diff
@@ -136,13 +237,12 @@ def update_state(self):
                 self.error_message = setup_response.error_message
             else:
                 self.state = MLBackendState.CONNECTED
-                model_version = setup_response.response.get('model_version')
+                model_version = setup_response.response.get('model_version', 'undefined')
```
niklub: let's fix this

label_studio/ml/models.py
```diff
@@ -156,52 +256,142 @@ def train(self):
...
+                'status': ml_api_result.status_code,
+                'error_message': ml_api_result.error_message,
+                'url': ml_api._get_url(PREDICT_URL),
+                'task': task_ser,
+                'request': request_params,
+                'response': results,
+            },
+        }
+
+    def predict_and_save(self, tasks=None):
```
niklub: maybe rollack it again

label_studio/ml/serializers.py
```diff
@@ -2,50 +2,119 @@
...
-from ml.models import MLBackend
+from ml.models import MLBackend, MLBackendAuth
 from rest_framework import serializers
 
 
 class MLBackendSerializer(serializers.ModelSerializer):
+    """ """
+
+    readable_state = serializers.SerializerMethodField()
+    basic_auth_pass = serializers.CharField(write_only=True, required=False, allow_null=True, allow_blank=True)
```
niklub, line 15: adding test to check GET payload

label_studio/data_manager/functions.py
niklub: since it is only one backend connected all the time, I wouldn't introduce this change rn to avoid big diff

label_studio/projects/serializers.py
```diff
@@ -166,6 +166,34 @@ def validate_label_config(self, value):
...
+
+        # Only run the validation if model_version is about to change
+        # and it contains a string
+        if p is not None and p.model_version != value and value != '':
+            # that model_version should either match live ml backend
+            # or match version in predictions
+
+            if (
+                p.ml_backends.filter(title=value).exists()
+                or p.predictions.filter(project=p, model_version=value).exists()
```
jombooth: I think you could do (pseudocode)

```
MlBackend.objects.filter(project=p, title=value).union(Prediction.objects.filter(project=p, model_version=value)).exists()
```

this would be one query instead of four, assuming p.ml_backends and p.predictions aren't prefetched.

label_studio/tests/next_task.tavern.yml
```diff
@@ -116,31 +97,6 @@ stages:
     json:
       data:
         text: Test example phrase
-      predictions:
```
jombooth, line 119: what's behind the changes to this test?

label_studio/ml/api.py
```diff
@@ -227,29 +283,65 @@ def post(self, request, *args, **kwargs):
...
-        result = ml_backend.interactive_annotating(task, context, user=request.user)
+        if not ml_results:
+            message = f'ML backend has to return a list with at least 1 annotation but it returned: {type(ml_results)}'
+            return self._error_response(message, logger.warning)
 
-        return Response(
-            result,
-            status=status.HTTP_200_OK,
-        )
+        return Response({'data': ml_results}, status=status.HTTP_200_OK)
```
jombooth: There are three test failures coming from this code, because `ml_api_result` at the top is a dictionary, not an `MLApiResult`, and therefore doesn't have any members like `is_error`. The tests that need to be fixed are:

```
tests/test_predictions.py::test_interactive_annotating
tests/test_predictions.py::test_interactive_annotating_failing
tests/test_predictions.py::test_interactive_annotating_with_drafts
```

label_studio/ml/models.py
```diff
@@ -96,36 +131,76 @@ class MLBackend(models.Model):
     def __str__(self):
         return f'{self.title} (id={self.id}, url={self.url})'
 
+    def __init__(self, *args, **kwargs):
+        super(MLBackend, self).__init__(*args, **kwargs)
+        self.__original_title = self.title
+
+    def save(self, *args, **kwargs):
```
jombooth, line 138: I think to be comfortable merging this, we really need a testcase that captures the behavior of this function (ideally something relatively high-level demonstrating the sequence of realistic API calls that causes project.model_version to be renamed, not just a unit test)

label_studio/ml/models.py
```diff
@@ -391,6 +498,15 @@ def _validate_ml_api_result(ml_api_result, tasks, curr_logger):
     return True
 
 
+@receiver(pre_delete, sender=MLBackend)
+def modify_project_model_version(sender, instance, **kwargs):
+    project = instance.project
+
+    if project.model_version == instance.title:
+        project.model_version = None
+        project.save(update_fields=['model_version'])
```
jombooth: as with MLBackend#save above, we really need a testcase that, using realistic API calls, captures the impact that this signal handler has.

jombooth:
Requesting screenshots of all new in-app functionality in the description, two testcases for model_version update behavior on save and delete, and the GET api test for basic_auth_pass mentioned earlier in the review - with these changes, I'm prepared to approve for the backend

label_studio/projects/serializers.py
```diff
@@ -177,8 +177,9 @@ def validate_model_version(self, value):
             # or match version in predictions
 
             if (
-                p.ml_backends.filter(title=value).exists()
-                or p.predictions.filter(project=p, model_version=value).exists()
+                # p.ml_backends.filter(title=value).exists()
+                # or p.predictions.filter(project=p, model_version=value).exists()
```
jombooth: please remove commented out code

jombooth:
Thanks for adding old vs new screenshots, super helpful :+1:

label_studio/ml/api_connector.py
```diff
@@ -31,18 +31,36 @@
...
+    def __init__(
+        self, url, timeout=None, connection_timeout=None, max_retries=None, headers=None, auth_method=None, **kwargs
+    ):
         self._url = url
         self._timeout = timeout or TIMEOUT_DEFAULT
         self._connection_timeout = connection_timeout or CONNECTION_TIMEOUT
         self._headers = headers or {}
+        self._auth_method = auth_method
+
+        # TODO add kwarg check in here
```
jombooth: What does this mean? Please clarify.

jombooth (approved):
Approving for backend changes, please address remaining comments about TODOs etc

bmartel (approved):
Approving for frontend. Thanks @niklub @jombooth 🎉

label_studio/projects/api.py
niklub: I think we don't have consistent naming convention regarding request args

jombooth (approved):
LGTM, also fixed ruff

PR #5603, opened 2024-03-19
fix: OPTIC-553: Import fails when user.active_organization is None

makseq:
This PR fixes a minor bug when 
1. project.created_by.active_organization is None (pretty rare case). 
2. request.user is None (in rq workers it's always None). 
3. the line in the code where this problem happens can **lead to very bad consequences - another organization might be taken instead of the real organization** where the project is created.

The PR makes code more clear and remove "bad-excess" lines of code.


bmartel (approved):
🔥 , also this linked directly to a sentry error as described in your description 🤯. Awesome all around!

PR #5612, opened 2024-03-20
fix: LEAP-884: avoid reading entire files to memory in export api

### LINKED ISSUE #5243: Easy Export to YOLO format causes absurd memory usage
I have a project with 2000 annotated images. 

Following those steps:
1. Go to project
2. click on "export"
3. select "yolo" format (others might work aswell) and confirm export
4. wait for label-studio to create the zip, then use absurd amounts of memory
5. if you are lucky, you get the file, but most times the kernel will kill label-studio because OOM

My system has 32GB RAM, 9GB is used by other applications. 22GB are available to label-studio
it uses *all 22GB* to export the project, and 99% of the times it crashes

**Expected behavior**
reasonable memory usage while exporting a project

**Environment:**
 - OS: Unraid, docker-compose
 - Label Studio Version: 1.10.0

**Additional context**
I couldn't wait for this to be resolved, as it made me unable to export.
So I started debugging this myself, what I discovered is this:
[this line](https://github.com/HumanSignal/label-studio/blob/develop/label_studio/data_export/models.py#L173) starts creating a zip, my first thought was it's doing this in-memory, causing the absurd memory usage. But it isn't.
```py
out = read_bytes_stream(os.path.abspath(tmp_dir + '.zip'))
```
Then I saw this. To me, it's reading a bytes stream. It has the word "stream" in it, so it probably reads in chunks?
[Nevermind](https://github.com/HumanSignal/label-studio/blob/5df9ae3828b98652e9fa290a19f4deedf51ef6c8/label_studio/core/utils/io.py#L143) it’s reading the *entire* zip file into memory for absolutely no reason:
```py

def read_bytes_stream(filepath):
    with open(filepath, mode='rb') as f:
        return io.BytesIO(f.read())
```
With my zip files being 7GB, its enough to somehow fill 22GB of RAM.

# Workaround
the zip is still created, you can go to label-studios tmp dir and download the file yourself, without loading it all into memory

jombooth:
Resolves #5243 

High level idea is to avoid calling `f.read()` in the `ExportAPI`, using `shutil.copy2` instead where we can.

Tested:
- with CI, for single file exports. eg `pytest tests/sdk/test_tasks.py::test_export_tasks`
- manually, for exporting an archive
- using `label-studio export`, for the `tasks.functions` change
- using `pytest tests/data_export.tavern.yml`, for the export mixin change.


label_studio/core/utils/io.py
```diff
@@ -140,9 +140,14 @@ def read_yaml(filepath):
...
-def read_bytes_stream(filepath):
-    with open(filepath, mode='rb') as f:
-        return io.BytesIO(f.read())
+def path_to_open_binary_file(filepath) -> io.BufferedReader:
+    """
+    The result of this call should generally be passed to a FileResponse or similar.
+    Unusually, this call deliberately doesn't close the file; FileResponse is responsible for this.
+    """
+    tmp = NamedTemporaryFile(delete=True)
+    shutil.copy2(filepath, tmp.name)
```
makseq, line 150: what does it do? why do we need it?

label_studio/data_export/api.py
```diff
@@ -197,14 +197,13 @@ def get(self, request, *args, **kwargs):
...
+        export_file, content_type, filename = DataExport.generate_export_file(
             project, tasks, export_type, download_resources, request.GET
         )
 
-        response = HttpResponse(File(export_stream), content_type=content_type)
-        response['Content-Disposition'] = 'attachment; filename="%s"' % filename
-        response['filename'] = filename
-        return response
+        r = FileResponse(export_file, as_attachment=True, content_type=content_type, filename=filename)
+        r['filename'] = filename
```
makseq, line 205: response['Content-Disposition'] = 'attachment; filename="%s"' % filename

it's important to have this header for export, as I remember SDK depends on it and some customers too

label_studio/data_export/mixins.py
```diff
@@ -319,11 +319,11 @@ def convert_file(self, to_format):
...
                 filename = pathlib.Path(input_name).stem + '.zip'
 
-            out = read_bytes_stream(output_file)
-            return File(
-                out,
-                name=filename,
-            )
+            with open(output_file, mode='rb') as f:
+                return File(
+                    io.BytesIO(f.read()),
```
makseq, line 325: is it ok to use f.read() here? it will load the whole file into the memory again..

makseq (changes requested):

label_studio/data_export/api.py
jombooth: FileResponse already does this! 
image

label_studio/core/utils/io.py
jombooth: If I just `open` the file at `filepath` here, it gets cleaned up before `FileResponse` finishes using the file. Instead I'm just copying the tempfile to the `NamedTemporaryFile`, which exists until it's closed (by FileResponse, or the caller).

label_studio/data_export/mixins.py
jombooth: I basically just unrolled what `read_bytes_stream` used to do here. I'm not sure exactly how to fix this part right now, but I don't think it's relevant to the issue I was trying to resolve in this PR. Added a comment suggesting taking a look in future.

label_studio/tasks/functions.py
```diff
@@ -119,14 +120,15 @@ def export_project(project_id, export_format, path, serializer_context=None):
...
         project, tasks, export_format, settings.CONVERTER_DOWNLOAD_RESOURCES, {}
     )
 
     # write to file
     filepath = os.path.join(path, filename) if os.path.isdir(path) else path
-    with open(filepath, 'wb') as file:
-        file.write(export_stream.read())
+    with open(filepath, 'wb'):
+        shutil.copy2(export_file.name, filepath)
+    export_file.close()
```
jombooth: I can confirm that this part works:

```
~/label_studio$ label-studio export 20 yolo
...
{"asctime": "21/Mar/2024:00:35:09 +0000", "name": "label_studio.server", "funcName": "main", "lineno": 335, "levelname": "INFO", "user_id": null, "message": "Project exported successfully: /label-studio/data/export/project-20-at-2024-03-21-00-35-efa9bad9.zip", "request_id": null}
```

And the zip file contents looked good.

makseq (approved):

jombooth:
With the Windows failures, we can't merge this as is. Finding a solution will require a bit more time.

Update 4/24: seems to be fixed via Django's NamedTemporaryFile

wesleylima (approved):

PR #5615, opened 2024-03-21
fix: LEAP-828: Add handling for carriage return in domManager

Gondragos:
1. Update the dom Managed to incorporate the handling of carriage return (CR) characters. This is a necessary change as some operating systems or browsers use both line feed (LF) and carriage return characters.

2. Previously we had a problem with missing regions when we are trying to render annotation from results that have a wrong order. It was caused by the problem with detecting area type (it thought it was an classification) when we have excessive fields in result (like a text). So in this commit I tried to fix it by adding expected values from value after creating an area and directly to the text field of  `RichTextRegion`.  (see https://github.com/HumanSignal/label-studio-frontend/commit/6ef0657ffa22fe4001129a84ef8aa21e430b2a0b#diff-9386d502ec5e5d9e8b50c2c679499430ca58728900aac1a43de33d13ee2648f4)

There was a situation when, when opening an annotation with richtext, some regions could receive incorrect text values, which led to saving the results without text inside.

- [ ] Performance
- [ ] Security
- [ ] UX

- [x] No

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)

`HyperText`, `RichText`, `Text`, `domManager`


web/libs/editor/src/tags/object/RichText/domManager.ts
```diff
@@ -366,7 +368,8 @@ class DomData {
     let fromIdx = this.displayedTextPos;
     const contentParts = [];
 
-    while (displayedText[fromIdx] === LF) {
+    // it should be just LF but in some OS / browsers it's LF and CR
```
hlomzik: From this comment it looks like different OS / browsers have different inner content for the same text, right? And in this case browsers with CRLF will have +1 extra symbol for every new line, producing shifted indices, which will be inconsistent in other OS / browsers and that's a huge risk. maybe we have to check for `CRLF|LF|CR` generating one symbol for CRLF(s)

web/libs/editor/src/regions/RichTextRegion.js
```diff
@@ -61,6 +61,15 @@ const Model = types
       }
     },
 
+    applyAdditionalDataFromResult(result) {
```
hlomzik, line 75: Definitely requires a description + the whole omitting values mechanism (which is precedecor for this method) should also be properly described, providing a reason and examples of possible dangerous behaviour.

web/libs/editor/src/stores/Annotation/Annotation.js
```diff
@@ -1214,6 +1214,7 @@ export const Annotation = types
             object: to_name,
             ...data,
             // We need to omit value properties due to there may be conflicting property types, for example a text.
+            // if we don't it can create a classification instead of proper area
```
hlomzik, line 1270: examples?

hlomzik (changes requested):
This PR definitely requires more description + it has a dangerous change in behaviour

web/libs/editor/src/tags/object/RichText/domManager.ts
Gondragos: Thank you. I'll fix related comments. But for the context here: it's inner mostly counter that do not affect positions in result, but it can provide visual isues with displayed text (in region details) as it will lead to converting /r/n/r/n into /n/n/n/n. But it's historically done thing.

web/libs/editor/src/tags/object/RichText/domManager.ts
```diff
@@ -366,7 +375,11 @@ class DomData {
     let fromIdx = this.displayedTextPos;
     const contentParts = [];
 
-    while (displayedText[fromIdx] === LF) {
+    // it should be just LF but in some OS / browsers it's CRLF (at least)
+    // in the other hand `fromIdx` is only inner counter that does not affect start/end offsets in results
```
hlomzik: "in the other hand" is excess here
```suggestion
    // `fromIdx` is only an inner counter that does not affect start/end offsets in results
```

web/libs/editor/src/tags/object/RichText/domManager.ts
```diff
@@ -1,10 +1,19 @@
...
+const CF = '\r';
 
 type DDExtraText = string;
 
+
+/**
+ * Normalize text for displaying it.
+ * It replaces all line breaks with '\n' symbol.
+ * This is a variant used historically, but it converts /r/n to /n/n which might be not correct
+ * @todo check if we can convert /r/n to /n without getting any problems
```
hlomzik:
```suggestion
 * This is a variant used historically, but it converts \r\n to \n\n which might be not correct
 * @todo check if we can convert \r\n to \n without getting any problems
```

hlomzik (approved):
small corrections, but that's great fix!

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8538260765)
> Successfully merged:  87 files changed, 11989 insertions(+), 8887 deletions(-)

MihajloHoma (approved):

PR #5619, opened 2024-03-22
docs: DIA-963: Add README to the ml_models, lse_ml_models and lse_ml_providers sub-package in LSE

matt-bernstein:
Add docs for new modules with autogenerated Mermaid diagrams. As this is an LSE-only feature for now, all the usage information will go in the LSE repo.


makseq (changes requested):
Could you add basic descriptions here pls? I would write here what model *provider* is and why we need it.

hakan458 (approved):

makseq (approved):

PR #5627, opened 2024-03-25
ci: PLT-287: Add docs tags generation

nikitabelonogov:


bmartel (commented):
🔥Nice work @nikitabelonogov

bmartel (approved):

PR #5646, opened 2024-03-28
fix: LEAP-908: Fix condition for Starter Cloud to fix DM

hlomzik:
Couple of fixes on how we check for expired Starter Cloud:
- main condition for Starter Cloud
- check that billing exists (it doesn't in Open Source)
- don't check for expiration in non Starter Cloud setup

DM was crashing on load because of missed `billing`

### What feature flags were used to cover this change?
`fflag_feat_front_leap_482_self_serve_short`

- [ ] Performance
- [ ] Security
- [x] UX

- [ ] No
- [x] Fixes breaking change


web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
```diff
@@ -21,37 +21,46 @@ const style = {
...
+  const isSelfServe = isFF(FF_SELF_SERVE) && window.APP_SETTINGS.billing?.enterprise === false;
 
-const WithDisabledTooltip = ({ children, ...props }) => {
-  if (!props.disabled) {
-    return children;
-  }
+  if (isOpenSource || !isSelfServe) return simpleButton;
 
+  // Check if user is on trial
+  const isTrialExpired = window.APP_SETTINGS.billing.checks?.is_license_expired;
```
juliosgarbi, line 37:
```suggestion
  const isTrialExpired = window.APP_SETTINGS.billing?.checks.is_license_expired;
```

bmartel (approved):

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
```diff
@@ -21,37 +21,46 @@ const style = {
...
-const isSelfServeExpired = isSelfServeExpiredTrial || isSelfServeExpiredSubscription;
+/**
+ * Checks for Starter Cloud trial expiration.
+ * If expired it renders disabled Import button with a tooltip.
+ */
+const ImportButtonWithChecks = ({ size }) => {
+  const simpleButton = <ImportButton size={size}>Import</ImportButton>;
+  const isOpenSource = !window.APP_SETTINGS.billing;
+  // Check if user is self-serve; Enterprise flag === false is the main condition
+  const isSelfServe = isFF(FF_SELF_SERVE) && window.APP_SETTINGS.billing?.enterprise === false;
```
juliosgarbi, line 32:
```suggestion
  const isSelfServe = isFF(FF_SELF_SERVE) && !window.APP_SETTINGS.billing?.enterprise;
```

juliosgarbi (changes requested):

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
bmartel: This is already covered by the above check to `!APP_SETTINGS.billing`

juliosgarbi: nvm

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
hlomzik: here I explicitly check for `false` because `undefined` here is when billing is not set up. I check this condition separately, so even with `!` it will work, but semantic will be a bit different which I don't want

juliosgarbi, line 32:
```suggestion
  const isSelfServe = isFF(FF_SELF_SERVE) && !window.APP_SETTINGS.billing?.enterprise;
```

juliosgarbi (approved):

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
hlomzik: `billing` is already validated before, so should be safe here. but we also have to check other nested values in case we have broken/different structure

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
bmartel: This would mean OpenSource would be considered based on the loosened falsey check.

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.js
bmartel: If `checks` were then to be nullish, it would break.

vladimirheartex (approved):

PR #5652, opened 2024-03-29
fix: LEAP-846: Fix date filtering in virtual tabs

jombooth:
previously we used packJSON to serialize filters for virtual tabs; this has some problems, in that it seems to get rid of hyphens, and has bugs that result in wrong datetimes if multiple datetimes are selected in the virtual tab. Instead of using packJSON to accomplish this, switch to gzip (by way of `pako`) and base64 serialization (by way of `js-base64`) to compactly serialize filters for URLs.

An example of an old packJSON url with a couple filters:

```
/projects/278389/data?tab=Y2zxDxw4fbr3g041r8bW101GVJPVlHhj4BpPOsyOpRsPOe7andGNPlHXiuWsco.UZL0JwRBzwDVm0wtasks.03gRMYJUYVy0NvDq8vPqQoRvQe9lessGNp9VnGWAOG0Gdm0VmHBe0du0T00VW1q01-05tm004Q2T.Cw48coRqRsOK0012EMiz6jcDJ00
```

and the same with gzip, though with 6 filters:

```
projects/38/data?tab=H4sIAAAAAAAAA62SQWvCQBCF_8uco4zG9rDnXgrFk6cWKWN2ItNmNyE7ESXkv3dii9aDlYKwh919b7_3GLYHFa0YHDxxSV2lkEEplXKbwPVQ1PGji4VKHc1C0ZssysHEt_7HaML3ximlz-S6xpOyf6eRVTfcktajqeKU7GZHVTfmzXG-mGA-meUrXDhEW1NEfDWLHppjI8OoBIYhu1NWjnfI2hwus2xGShJ_5eUn7Isk_QNpQJpKoC3fQgIlX57rPh_fXAeLvwRKPKN6CHZ0swwC7a3scKIuu7Ax1nVs0fJ_xo2PK3y4Me61xdet51bi1j7VevgCXznVcZICAAA
```

for backwards compatibility, we first attempt to deserialize with gzip, and if that fails, fall back to packJSON.

This fixes numerous issues with virtual tabs, and most saliently, allows date filtering to work within them.

The URLs above can be used for testing. Verify that deserialization is successful for both the virtual tabs given, but that the packJSON one results in a `urlJSON.js:15 falling back to packJSON.` message on the console and an error like the below, originating from a wrongly formatted date filter.

image


web/libs/datamanager/src/utils/urlJSON.js
```diff
@@ -0,0 +1,20 @@
...
+    try {
+        return JSON.parse(ungzip(toUint8Array(s), { to: 'string' }));
+    } catch (e) {
+        console.log('Error deserializing gzipped data:', e);
+        console.log('falling back to packJSON.');
+        return packJSON.parse(s)
+    }
+}
+
+export const serializeJsonForUrl = (s) => fromUint8Array(gzip(JSON.stringify(s)), true)
```
jombooth: note that this is url-safe: https://github.com/dankogai/js-base64/blob/main/base64.js#L102-L110

web/libs/datamanager/src/utils/urlJSON.js
```diff
@@ -0,0 +1,20 @@
+// Library that serializes and deserializes JSON data to/from URL-safe strings
+// Uses gzip to minimize the size of the serialized data
+
+import { gzip, ungzip } from 'pako'
+import { fromUint8Array, toUint8Array } from 'js-base64'
+import { packJSON } from './packJSON';
+
+export const deserializeJsonFromUrl = (s) => {
+    // Fall back to packJSON if deserialization fails; this probably means
+    // we are deserializing an old virtual view. TODO(jo): deprecate packJSON entirely
```
nehalecky: @jombooth, what's a path to deprecating all old virtual views?

web/libs/datamanager/src/utils/urlJSON.js
```diff
@@ -0,0 +1,20 @@
...
+    try {
+        return JSON.parse(ungzip(toUint8Array(s), { to: 'string' }));
+    } catch (e) {
+        console.log('Error deserializing gzipped data:', e);
```
nehalecky: Will the managed exception provide us a reference the the old virtual view that is causing the error?

web/libs/datamanager/src/utils/urlJSON.js
jombooth: Whenever we're ready, we can cut a new version that removes deserialization via packJSON from the exception handler below, and at that point we can delete packJSON entirely. I've chatted with @amymrou about when we can do this, following the change management best practices that we're discussing - we haven't locked those yet, so for now, I've just cut a ticket for the deprecation work (LEAP-914) and whenever we pick that up, some lucky engineer can enjoy ripping ~600 lines of buggy code out of LS :)

web/libs/datamanager/src/utils/urlJSON.js
jombooth: not currently - if we get here, all we know is that we couldn't deserialize with gzip, so we try packJSON instead; for properly encoded packJSON query params (ie, no datetimes because those are broken), the user will get their intended virtual tab seamlessly. If the query string is neither valid gzip nor valid packjson, the packjson call fails silently and we get an empty virtual tab (which is actually not a terrible experience)

web/libs/datamanager/src/utils/urlJSON.js
hlomzik, line 1: let's make it in typescript

web/libs/datamanager/src/utils/packJSON.js
```diff
@@ -1,3 +1,5 @@
+// DEPRECATED; don't use this. use utils/urlJSON instead.
```
hlomzik: better to use jsdoc comment here
```suggestion
/** @deprecated Buggy legacy library, don't use it. Use utils/urlJSON instead. */
```

web/libs/datamanager/src/utils/urlJSON.js
hlomzik: linting issues:
- we use 2 spaces indent
- add trailing `;`

hlomzik (changes requested):
some comments to improve the code

web/libs/datamanager/src/utils/urlJSON.ts
```diff
@@ -0,0 +1,21 @@
+/** Library that serializes and deserializes JSON data to/from URL-safe strings
+ *  using gzip to minimize the size of the serialized data.
+ */
```
hlomzik:
```suggestion
/**
 * Library that serializes and deserializes JSON data to/from URL-safe strings
 * using gzip to minimize the size of the serialized data.
 */
```

web/libs/datamanager/src/utils/urlJSON.ts
```diff
@@ -0,0 +1,21 @@
...
+  try {
+    return JSON.parse(ungzip(toUint8Array(s), { to: 'string' }));
+  } catch (e) {
+    console.log('Error deserializing gzipped data:', e);
+    console.log('falling back to packJSON.');
+    return packJSON.parse(s);
+  }
+}
+
+export const serializeJsonForUrl = (o: object) => fromUint8Array(gzip(JSON.stringify(o)), true);
```
hlomzik, line 22: basically that should be also `unknown`
```suggestion
export const serializeJsonForUrl = (o: unknown) => fromUint8Array(gzip(JSON.stringify(o)), true);
```

hlomzik (approved):

web/libs/datamanager/src/utils/urlJSON.ts
hlomzik: Yeah, `unknown` here is too much, strange to force users to type value for JSON.stringify which is already `any`. And given the usage context `obejct` is a good for here

hlomzik (approved):

jombooth:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8516380391)

bmartel (approved):
Nicely done! This drastically reduces the complexity and maintenance of virtual tab filtering.

PR #5692, opened 2024-04-08
fix: LEAP-859: Fix data manager breaking on displaying specific data

Gondragos:
The commit replaces usage of `toStudlyCaps` across several files with a new function `normalizeCellAlias` to standardize cell alias normalization across the app. This change ensures a consistent approach to handling cell alias normalization.

`CellViews` is used only to chose the right way do display the column and for aliases there are just a couple of variants that are apllicable in that case. So that changing any other string will not break anything.

`toStudlyCaps` includes [code](https://github.com/dleitee/strman/blob/master/src/tostudlycaps.js#L30)
```JavaScript
value.replace(/[-_\s]+(.)?/g, function (match, chr) {
  return toUpperCase(chr);
});
``` 
which does not work with values like `"smthng_"` as `chr` should be a string but we get `undefined`. To prevent this we need to remove unsafe trailing special characters before using this function.

- [ ] Performance
- [ ] Security
- [ ] UX

- [x] No

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)

`CellViews`


web/libs/datamanager/src/components/CellViews/index.js
```diff
@@ -1,3 +1,5 @@
+import {toStudlyCaps} from "strman";
```
juliosgarbi:
```suggestion
import { toStudlyCaps } from "strman";
```

web/libs/datamanager/src/components/CellViews/index.js
```diff
@@ -15,3 +17,9 @@ export { StringCell as Text } from "./StringCell";
 export { VideoCell as Video } from "./VideoCell";
 export { ProjectCell as Project } from './ProjectCell';
 
+export function normalizeCellAlias(alias) {
+  const safeAlias = alias
```
juliosgarbi:
```suggestion
  // remove trailing separators to make `toStudlyCaps` safe
  const safeAlias = alias.replace(/[-_\s]+$/g, '');
```

I believe that we could not break this line, since this line is not more than 50 chars, I don't think we should break it; doing so makes reading easier

juliosgarbi (approved):
LGTM, with just a few minor adjustments needed to improve readability but not a blocket

juliosgarbi (commented):
Also that would be nice if we had some tests here

web/libs/datamanager/src/components/CellViews/index.js
yyassi-heartex: great observation @juliosgarbi +1 to this

web/libs/datamanager/src/stores/Tabs/tab_column.js
```diff
@@ -214,7 +223,7 @@ export const TabColumn = types
     },
 
     get filterable() {
-      return ((CellViews[self.type] ?? CellViews[toStudlyCaps(self.alias)])?.filterable) !== false;
+      return ((CellViews[self.type] ?? CellViews[normalizeCellAlias(self.alias)])?.filterable) !== false;
```
yyassi-heartex: lets break this down a little for readability
```suggestion
      const cellView = CellViews[self.type] ?? CellViews[normalizeCellAlias(self.alias)];
      
      return cellView?.filterable !== false;
```

yyassi-heartex: also this code implies that its possible for cellView to be undefined meaning we could get `undefined !== false` which would give us true - is that desired? do want to switch to something like 
```suggestion
      const cellView = CellViews[self.type] ?? CellViews[normalizeCellAlias(self.alias)];
      
      return !!cellView?.filterable;
```
this would implicitly make it return boolean every time

Gondragos: I can't say if that is desired. I'm not the one who what originally an author of that part of code, so I wouldn't like to change this logic to avoid unexpected results.

web/libs/datamanager/src/components/CellViews/index.js
```diff
@@ -15,3 +17,10 @@ export { StringCell as Text } from "./StringCell";
 export { VideoCell as Video } from "./VideoCell";
 export { ProjectCell as Project } from './ProjectCell';
 
+export function normalizeCellAlias(alias) {
+  // remove trailing separators to make `toStudlyCaps` safe
+  const safeAlias = alias.replace(/[-_\s]+$/g, '');
+    // remove trailing separators to make `toStudlyCaps` safe
+    .replace(/[-_\s]+$/g, '');
```
yyassi-heartex: with the applied replacement we wont need this anymore - infact as it is it would cause an error
```suggestion
 
```

yyassi-heartex (approved):

MihajloHoma (approved):

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8649273085)
> Successfully merged:  108 files changed, 2197 insertions(+), 1985 deletions(-)

PR #5719, opened 2024-04-11
feat: DIA-957: Delete predictions when ModelVersion updated or ModelR…

hakan458:
…un deleted

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/ml_models/models.py
+++ b/label_studio/ml_models/models.py
@@ -65,8 +65,9 @@ class ModelVersion(models.Model):
         """
         Deletes Predictions associated with ModelVersion
         """
-        model_run = ModelRun.objects.get(model_version=self.id)
-        model_run.delete_predictions()
+        model_runs = ModelRun.objects.filter(model_version=self.id)
+        for model_run in model_runs:
+            model_run.delete_predictions()
         super().delete(*args, **kwargs)
 
 
--- a/label_studio/ml_models/models.py
+++ b/label_studio/ml_models/models.py
@@ -184,4 +184,3 @@ class ModelRun(models.Model):
         """
         self.delete_predictions()
         super().delete(*args, **kwargs)
-
--- a/label_studio/ml_models/models.py
+++ b/label_studio/ml_models/models.py
@@ -3,10 +3,13 @@
 
 from django.conf import settings
 from django.db import models
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
 from ml_model_providers.models import ModelProviderConnection
 from projects.models import Project
 from rest_framework.exceptions import ValidationError
+from tasks.models import Prediction
 
 
 def validate_string_list(value):
@@ -59,6 +62,16 @@ class ModelVersion(models.Model):
         return f'{self.parent_model.title}__{self.title}'
 
     prompt = models.TextField(_('prompt'), null=False, blank=False, help_text='Prompt to execute')
+
+    def delete_predictions(self):
+        """
+        Deletes any predictions that have originated from a ModelVersion
+
+        Currently assumes that we are setting the `model_version` field of Prediction
+        to ModelVersion.full_title
+        """
+
+        predictions = Prediction.objects.filter(model_version=self.full_title).delete()
 
 
 class ThirdPartyModelVersion(ModelVersion):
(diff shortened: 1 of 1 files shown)


label_studio/ml_models/models.py
```diff
@@ -159,3 +172,27 @@ def output_file_name(self):
...
+def delete_predictions_model_version(sender, instance, **kwargs):
+    """
+    Deletes all Prediction objects associated with a ModelVersion when
+    deleting that ModelVersion
+    """
+
+    instance.delete_predictions()
+
+
+@receiver(pre_delete, sender=ModelVersion)
```
forum-hs: I am guessing the sender will be ModelRun here?

label_studio/ml_models/models.py
```diff
@@ -60,6 +63,16 @@ def full_title(self):
...
 
+    def delete_predictions(self):
+        """
+        Deletes any predictions that have originated from a ModelVersion
+
+        Currently assumes that we are setting the `model_version` field of Prediction
+        to ModelVersion.full_title
+        """
+
+        predictions = Prediction.objects.filter(model_version=self.full_title).delete()
```
forum-hs: do we need to have check here that modelrun status is completed to allow delete on predictions?

label_studio/ml_models/models.py
```diff
@@ -60,6 +63,16 @@ def full_title(self):
 
     prompt = models.TextField(_('prompt'), null=False, blank=False, help_text='Prompt to execute')
 
+    def delete_predictions(self):
+        """
+        Deletes any predictions that have originated from a ModelVersion
+
+        Currently assumes that we are setting the `model_version` field of Prediction
+        to ModelVersion.full_title
```
matt-bernstein: I don't think this is true. Very confusingly, `Prediction.model_version` was added during the recent MLBackend changes, it describes the version of an MLBackend, and is entirely unrelated to this ModelVersion. We are setting `Prediction.model_run`: https://github.com/HumanSignal/label-studio/blob/fb-dia-957/label_studio/tasks/models.py#L869 

I don't think we are setting `Prediction.model_version`, and if we are, we should stop, since it'd show up in the UI for autolabeling using an MLBackend and confuse people, afaiui.

label_studio/ml_models/models.py
```diff
@@ -159,3 +172,27 @@ def output_file_name(self):
     @property
     def error_file_name(self):
         return f'{self.project.id}_{self.model_version.pk}_{self.pk}/error.csv'
+
+
+@receiver(pre_delete, sender=ModelVersion)
```
matt-bernstein: Consider switching from using a signal to putting this inside `ModelVersion.delete()`: https://www.notion.so/humansignal/Backend-Handbook-ce7324a497cb4b40b91911a429e34fcb?pvs=4#052c958f2f9c48b6b6f1b207bca3674d

label_studio/ml_models/models.py
hakan458: Ok good to know. We are currently setting it for sure

matt-bernstein: Yeah, it's possible that it's actually ok, but that's not what the field was added for

hakan458: Yeah its not the exact use of this field, but I think we do need to set it actually. Its whats shown on the project side as what made this prediction, its just used as the name of the model essentially. 
If we dont want to reference it this way, we need to add a new foreign key for prompter ModelVersion as well, just ModelRun wont do it, since when we update a ModelVersion we want to delete existing Predictions. 
But I guess we enforce that there is only one ModelRun per ModelVersion, so could be ok. Will try it

label_studio/ml_models/models.py
```diff
@@ -3,10 +3,13 @@
 
 from django.conf import settings
 from django.db import models
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
```
matt-bernstein: unused imports

label_studio/ml_models/models.py
```diff
@@ -54,11 +57,19 @@ class Meta:
...
     @property
     def full_title(self):
         return f'{self.parent_model.title}__{self.title}'
 
-    prompt = models.TextField(_('prompt'), null=False, blank=False, help_text='Prompt to execute')
+    def delete(self, *args, **kwargs):
+        """
+        Deletes Predictions associated with ModelVersion
+        """
+        model_run = ModelRun.objects.get(model_version=self.id)
```
matt-bernstein: wouldn't this throw an error if there are 2 model runs for this model version, one on GT and one on all tasks? Or did we make that impossible?

label_studio/ml_models/models.py
```diff
@@ -159,3 +168,17 @@ def output_file_name(self):
     @property
     def error_file_name(self):
         return f'{self.project.id}_{self.model_version.pk}_{self.pk}/error.csv'
+
+    def delete_predictions(self):
+        """
+        Deletes any predictions that have originated from a ModelRun
+        """
+
```
matt-bernstein, line 177: does [updating the task counters](https://github.com/HumanSignal/label-studio-enterprise/commit/fc0daaea9fef1f9b3a5df8e868d39a8b832a57c1) also need to happen here?

label_studio/ml_models/models.py
hakan458: No, this is for one model version object. 
For separate subsets, there will be a different model version object

label_studio/ml_models/models.py
hakan458: No since we're calling it after calling delete on each model run. I wanted to call it once versus after each object deletion. If we have another flow of deleting model runs later on we will need to add it there as well. For now we're good

matt-bernstein: ah, makes sense

matt-bernstein (approved):

forum-hs (approved):

pakelley (approved):

PR #5729, opened 2024-04-15
fix: LEAP-1004: Show prediction with selected project.model_version in QuickView

niklub:
In "QuickView" mode, user now sees the prelabeled data from the prediction based on selected model version.
If prediction from this model is not presented it fallbacks to old behavior and uses the first prediction.

**Model version selected**:
image

**Correct prediction is shown first**:
image
image

- [x] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend


hlomzik (approved):

deppp:
Why do we think this is a correct behavior? That should be an EXPLICIT user action to select what predictions should be used?

MihajloHoma (approved):

niklub:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8705834891)
> Successfully merged:  327 files changed, 70 insertions(+), 14 deletions(-)

niklub:
> Why do we think this is a correct behavior? That should be an EXPLICIT user action to select what predictions should be used?

In the current implementation, the only explicit user action is to set the model version in project settings, which produces the bug fixed in this PR.

Two alternatives that could be better from UX perspective:
- either let the user create annotation from prediction manually, without pre-initializing it (downside: it can break existing QuickView labeling flow without or with a single prediction, or introduce inconsistency)
- or let the user copy predictions into the draft annotation (downside: this tool has to be implemented)

Is there any other scenario that can better address the problem?

Anyway, this PR obviously fixes an incorrect behavior, and we can address future enhancements in the next ones

PR #5741, opened 2024-04-17
fix: LEAP-947: Check task when we get a response from ML

hlomzik:
Add a wrapper to check that we are going to apply response from ML-backend to the same task we sent a request from in a first place.

If ML-backend responds long enough and we switched to another task (by submit, prev, postpone, etc) then its response will be applied to a wrong annotation, not only making a confusion but potentially distorting the image because of image sizes embedded into regions.

We should also stop requests when we leave a task, this will be done in a future.

- [ ] Performance
- [ ] Security
- [ ] UX

- [ ] No
- [x] Fixes breaking change (data corruption)

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)


web/libs/editor/src/stores/AppStore.js
```diff
@@ -798,11 +798,18 @@ export default types
...
+      try {
+        const response = yield request;
+
+        if (requestId === self.suggestionsRequest) {
+          self.annotationStore.selected.setSuggestions(dataParser(response));
+          self.setFlags({ awaitingSuggestions: false });
+        }
+      } catch(e) {
         self.setFlags({ awaitingSuggestions: false });
+        // @todo handle errors + situation when task is changed
```
yyassi-heartex, line 811: since there is a todo here it might be prudent to add a `console.error` temporarily as a placeholder while a more elegant solution is put in place
```suggestion
        // @todo handle errors + situation when task is changed
        console.error('Error:', e);
```

web/libs/editor/src/stores/AppStore.js
```diff
@@ -798,11 +798,18 @@ export default types
...
+        return;
```
yyassi-heartex: maybe im misreading/misunderstanding the code but is this return needed? doesn't really appear to be giving us an early return. Am i missing something?
```suggestion
```

Gondragos (approved):

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -140,7 +140,20 @@ export const DataManagerPage = ({ ...props }) => {
...
+        // we'll check that we are processing the same task
+        const wrappedRequest = new Promise(async (resolve, reject) => {
+          const response = await suggestionsRequest;
+
+          // right now task might be an old task,
+          // so in order to get a current one we need to get it from lsf
+          if (task.id === dataManager.lsf.task.id) {
+            resolve(response);
+          } else {
+            reject();
```
yyassi-heartex, line 152: if we `reject()` then we move on to `lsf.loadSuggestions(wrappedRequest, (response) => {` wouldn't response be `undefined`? if thats the case line 157 `if (response.data) {` will throw an error most likely

yyassi-heartex: ahh, disregard - we catch in loadSuggestions so we dont fire `dataParser`. Resolving convo

web/libs/editor/src/stores/AppStore.js
hlomzik: ah, yeah, forgot to remove it after rearranging the code, thank you!

web/libs/editor/src/stores/AppStore.js
hlomzik: state here is not always an error like in this specific case we are fixing here — it's just a way to ignore the wrong request. later we'll come up with two different things:
- proper handling for errors from ML backend
- proper request management and cancellation

yyassi-heartex: ok sounds good

yyassi-heartex (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8729250192)
> Successfully merged:  5 files changed, 41 insertions(+), 1 deletion(-)

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/8792350026)
> Successfully merged:  53 files changed, 1618 insertions(+), 898 deletions(-)

MihajloHoma (approved):

PR #5742, opened 2024-04-17
fix: LEAP-947: Check task when we get a response from ML

hlomzik:
Add a wrapper to check that we are going to apply response from ML-backend to the same task we sent a request from in a first place.


Gondragos (approved):

yyassi-heartex (approved):
approving because these changes should get us unstuck, but original LSO PR had some questions that should still be addressed but shouldn't have an impact on outcome

PR #5769, opened 2024-04-23
feat: LEAP-1119: Feature/gridview thumbnail

FrsECM:

- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

When using multilayered images, the original behaviour was to create a src that separate each image by a ','.

This way, it's not displayable.
image

#### What does this fix?
It fixes an issue in the gridview when there is multiple layers image.
image
It's even possible to use a specific "thumbnail" field, that allows to add a specific thumbnail to be use.
image
If you want the thumbnail to be displayed in the column view, just declare it as an image in the taxonomy : 
image
image

There is two different new behaviors.
If a "thumbnail" is provided : 
- The gridview uses the thumbnail field.
The advantage is the ability to use a thumbnail that has a lower resolution than the original image.
else:
We use the first image of the multilayer images.

The current behavior is that in this case, images url are separated by a ',' in the image src.
What display nothing at all

It can improve the performance because thumbnail can be a lower resolution image.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

GridView => Data Exploration


sajarin:
/jira create
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/9289869308)
> Jira issue [TRIAG-577](https://humansignal.atlassian.net/browse/TRIAG-577) is created

[TRIAG-577]: https://humansignal.atlassian.net/browse/TRIAG-577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

juliosgarbi (approved):
I added to the PR unit tests, also I removed the feature about show thumbnails, we already have something similar to it on the list view that doesn't need add any tags on the config, you just need to add a new node on the data (e.g. thumb), enable this column on the columns dropdown, and change the view mode from `str` to `img` on the thumb column, overall it looks good, thanks for the contribution!

FrsECM:
> I added to the PR unit tests, also I removed the feature about show thumbnails, we already have something similar to it on the list view that doesn't need add any tags on the config, you just need to add a new node on the data (e.g. thumb), enable this column on the columns dropdown, and change the view mode from `str` to `img` on the thumb column, overall it looks good, thanks for the contribution!

If you have a piece of document about that I'm very interessed. It can help when we have very high resolution images. 

Thanks !

PR #5784, opened 2024-04-25
fix: LEAP-1036: Revert validation on attr name

juliosgarbi:
We are reverting the validation of the attribute name on visual tags because it is disrupting the workflow. We plan to devise a method to validate these attributes by showing a warning, instead of breaking the workflow, and will implement this solution afterwards


hlomzik (approved):

makseq (approved):

jombooth:
Force merging due to incident

PR #5819, opened 2024-05-03
fix: pre-commit when --all-files is use on manual, for example in 'make fmt-al'

luarmr:
When you execute 
`pre-commit run --hook-stage manual "biome-check" --config .pre-commit-dev.yaml --all-files`
for example using  'make fmt-al' biome hangs because files outside the main configuration are being pass.


bmartel (approved):

luarmr:
@bmartel I tested with no files modified, and also with an intentional error and looks it works as intended.
Can you double check?

PR #5820, opened 2024-05-03
fix: OPTIC-712: Remove warnings in the app related to injection occur…

luarmr:
Removing warnings in console:

Screenshot 2024-05-03 at 9 52 27 AM


bmartel (approved):

hlomzik (approved):
This can even be improved further because some of components (`RelationsTab`, `InfoTab`) don't need `store`, so `inject()` can also be removed; plus for the rest `observer()` can be moved to component declaration as well for consistency.
But it's ok like this to fix the bug!

PR #5843, opened 2024-05-07
feat: LEAP-1027: Add option for including annotation history

triklozoid:


hlomzik:
fixed and merged already in https://github.com/HumanSignal/label-studio/pull/5861

PR #5844, opened 2024-05-07
fix: LEAP-889: remove env_vars from context log

sajarin:

_(check all that apply)_
- [ ] Product design
- [X] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

Based on the current code, context logs currently collect environment variables that are prefixed with `LABEL_STUDIO`. There may be a case where someone creates a non label studio related env variable and that accidentally gets exposed to us. This PR is meant to remove env_vars from the payload entirely. 

#### What does this fix?
This removes the env_var object from the payload and also removes the get_label_studio_env function (as this is no longer needed). 

The new behavior is that environmental variables are no longer present in our context logs. 

Yes, this minimizes our exposure to sensitive user data. 

- [ ] No

- [X] unit


label_studio/feature_flags.json
```diff
@@ -12,10 +12,7 @@
         "variation": 0
```
jombooth: Can we revert the changes in this file? Looks like it was autoformatted

label_studio/tests/test_contextlog.py
```diff
@@ -0,0 +1,25 @@
+"""This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
+"""
```
jombooth:
```suggestion
```

Dockerfile
```diff
@@ -37,6 +37,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
     POETRY_VIRTUALENVS_CREATE=false \
     DJANGO_SETTINGS_MODULE=core.settings.label_studio \
     LABEL_STUDIO_BASE_DATA_DIR=/label-studio/data \
+    LABEL_STUDIO_TEST_ENV_VAR=true \
```
jombooth:
```suggestion
```

label_studio/tests/test_contextlog.py
```diff
@@ -0,0 +1,25 @@
...
+
+import pytest
+import responses
+from django.test import override_settings
+
+
+@responses.activate
+@pytest.mark.django_db
+@override_settings(CONTEXTLOG_SYNC=True)
+def test_contextlog(business_client):
```
jombooth:
```suggestion


@responses.activate
@pytest.mark.django_db
def test_contextlog(business_client, settings):
    settings.CONTEXTLOG_SYNC = True
```

I realized that we could leverage pytest's `settings` fixture instead to make this a little simpler

label_studio/tests/test_contextlog.py
```diff
@@ -0,0 +1,25 @@
+"""This file and its contents are licensed under the Apache License 2.0. Please see the
+included NOTICE for copyright information and LICENSE for a copy of the license.
+"""
```
jombooth: Ah I was actually proposing we remove the license string, I don't think these are needed

label_studio/core/settings/base.py
```diff
@@ -739,3 +739,7 @@ def collect_versions_dummy(**kwargs):
 
 CLOUD_STORAGE_CHECK_FOR_RECORDS_PAGE_SIZE = get_env('CLOUD_STORAGE_CHECK_FOR_RECORDS_PAGE_SIZE', 10000)
 CLOUD_STORAGE_CHECK_FOR_RECORDS_TIMEOUT = get_env('CLOUD_STORAGE_CHECK_FOR_RECORDS_TIMEOUT', 60)
+
+CONTEXTLOG_SYNC = False
```
wesleylima, line 743: Do we want to `get_bool_env` for this one too? Or is it not supposed to be settable through env variables? How would this be set then?

sajarin: This variable shouldn't really be settable, we specify it here in the settings to that we can override it via the settings fixture in conftest.

wesleylima (approved):

jombooth (approved):
LGTM, nice work!

PR #5847, opened 2024-05-08
fix: OPTIC-719: TypeError: e.target.className.includes is not a function

luarmr:
Error coming from Sentry 
https://heartex.sentry.io/issues/5039994894/

There is no impact on the user.

This happens when a user drags and drops, for example, when the column header is an SVG. I confirm that the drag-and-drop feature continues to work well.

Also, `?` wont work because in the case of failure is an image that is an svg and className is a `SVGAnimatedString` object. No need to pursue baseVal since we don't have the className `handle`


web/libs/datamanager/src/components/Common/Table/TableHead/TableHead.jsx
```diff
@@ -218,7 +218,10 @@ export const TableHead = observer(
 
         useEffect(() => {
           ref.current?.addEventListener("mousedown", (event) => {
-            if (event.target.className.includes("handle")) event.preventDefault();
+            const className = event.target.className;
+            if (typeof className === "string" && className.includes("handle")) {
```
bmartel, line 222:
```suggestion
            if (className?.includes("handle")) {
```

web/libs/datamanager/src/components/Common/TableOld/TableHead/TableHead.jsx
```diff
@@ -212,7 +212,10 @@ export const TableHead = observer(
 
         useEffect(() => {
           ref.current?.addEventListener("mousedown", (event) => {
-            if (event.target.className.includes("handle")) event.preventDefault();
+            const className = event.target.className;
+            if (typeof className === "string" && className.includes("handle")) {
```
bmartel, line 216:
```suggestion
            if (className?.includes("handle")) {
```

bmartel (approved):

web/libs/datamanager/src/components/Common/Table/TableHead/TableHead.jsx
bmartel: This would also work, as in the case of className being anything other than a string the method includes will also be undefined.
```suggestion
            if (className?.includes?.("handle")) {
```

web/libs/datamanager/src/components/Common/TableOld/TableHead/TableHead.jsx
bmartel:
```suggestion
            if (className?.includes?.("handle")) {
```

web/libs/datamanager/src/components/Common/Table/TableHead/TableHead.jsx
luarmr: Do you feel strong about this? I will go either way.
I'd like to show that I'm using "includes" because it's a string. But, again,  if you feel strong about it I will change it.

hlomzik: I would also vote for null check, because it's more semantically correct. In your code it feels like `className` can be some other type which is not the case, right? We are just protecting us against the target without class name (what the case is it btw?)

hlomzik: ah, sorry, I re-read the description and now see that it can actually be a different kind of "string". Then both solutions are good, but they both require a comment. (still vote for shorter version though)

hlomzik:
```suggestion
            const className = event.target.className;

            if (typeof className === "string" && className.includes("handle")) {
```
linter eyes...
(no need in this fix with the Brandon's colution)

hlomzik (approved):

web/libs/datamanager/src/components/Common/Table/TableHead/TableHead.jsx
luarmr: We are in the after biome era, It is a rule that we should enable?

luarmr: Update to use the shorter version since both of you agree :)

luarmr: Also added a comment

hlomzik: we had this rule before: https://eslint.org/docs/latest/rules/newline-after-var
for some reason can't see it in biome rules

PR #5861, opened 2024-05-08
feat: LEAP-1026: Add option for including annotation history

triklozoid:
added parameter for including annotation history in export, it does not change the export format for LSO, because history does not exist here. Related methods will be overridden in LSE.

Mirroring https://github.com/HumanSignal/label-studio/pull/5843

### DIFF
--- a/label_studio/data_export/mixins.py
+++ b/label_studio/data_export/mixins.py
@@ -135,6 +135,11 @@ class ExportMixin:
             options['context'] = {'interpolate_key_frames': settings.INTERPOLATE_KEY_FRAMES}
             if 'interpolate_key_frames' in serialization_options:
                 options['context']['interpolate_key_frames'] = serialization_options['interpolate_key_frames']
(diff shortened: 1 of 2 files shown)


label_studio/data_export/mixins.py
```diff
@@ -135,6 +135,11 @@ def _get_export_serializer_option(serialization_options):
             options['context'] = {'interpolate_key_frames': settings.INTERPOLATE_KEY_FRAMES}
             if 'interpolate_key_frames' in serialization_options:
                 options['context']['interpolate_key_frames'] = serialization_options['interpolate_key_frames']
+            if (
+                'include_annotation_history' in serialization_options
+                and serialization_options['include_annotation_history'] is False
+            ):
```
jombooth:
```suggestion
            if serialization_options.get('include_annotation_history') is False:
```

jombooth (approved):

Issue #5704, opened 2024-04-09
DOMAIN_FROM_REQUEST=True is maybe inconsistent with HOSTNAME handling


curlup:
is it me or is this code https://github.com/HumanSignal/label-studio/blob/develop/label_studio/core/settings/base.py#L111-112

clashes with this code just couple lines above
https://github.com/HumanSignal/label-studio/blob/develop/label_studio/core/settings/base.py#L88-93
? 

My reading is that if `DOMAIN_FROM_REQUEST=True`
by the time line 111 get's executed, the line 92 or `else` clause on line 93 **guarantee** that 112 will raise an error. 

So basically there's no configuration with `DOMAIN_FROM_REQUEST=True` that will not raise on L112.

oreilles:
Facing this as well, lines 107:114 should be moved above line 87

curlup:
I guess somehow it doesn't make sense to the maintainers?

makseq:
Could you please describe more details what settings are you trying to use? 
Have you tried to use something like LABEL_STUDIO_HOST=http://domain/subpath instead of LABEL_STUDIO_HOST=/subpath?

oreilles:
If you set `LABEL_STUDIO_HOST=http://domain/subpath` and `DOMAIN_FROM_REQUEST=True`, the result will be an exception because:

https://github.com/HumanSignal/label-studio/blob/39a92aef950cd2c6a8ac92a39b17b7e1716b52f9/label_studio/core/settings/base.py#L109-L112

If you set `LABEL_STUDIO_HOST=/subpath` and `DOMAIN_FROM_REQUEST=True`, it won't work either cause the variable `LABEL_STUDIO_HOST` will be reset to an empty string:

https://github.com/HumanSignal/label-studio/blob/39a92aef950cd2c6a8ac92a39b17b7e1716b52f9/label_studio/core/settings/base.py#L87-L92

I made a PR to address this but it has been closed: https://github.com/HumanSignal/label-studio/pull/5874

PR #5874, opened 2024-05-13
fix: #5704 Fix DOMAIN_FROM_REQUEST validation for HOSTNAME

oreilles:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [ ] Frontend

https://github.com/HumanSignal/label-studio/issues/5704

#### What does this fix?
When the `DOMAIN_FROM_REQUEST`  env variable is set, `HOSTNAME` is expected to start with `/` (so that the subpath is added to domain). However, that constraint validation happens *after* another validation for `HOSTNAME`, which is otherwise expected to start with `https?://`, and resets it to an empty string if that isn't the case - meaning that `DOMAIN_FROM_REQUEST` cannot be used successfully.

The validation for a `HOSTNAME` subpath used with `DOMAIN_FROM_REQUEST` is moved above the check for a `https://?` `HOSTNAME`.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/core/settings/base.py
+++ b/label_studio/core/settings/base.py
@@ -84,8 +84,13 @@ SILENCED_SYSTEM_CHECKS = []
 
 # Hostname is used for proper path generation to the resources, pages, etc
 HOSTNAME = get_env('HOST', '')
+DOMAIN_FROM_REQUEST = get_bool_env('DOMAIN_FROM_REQUEST', False)
+
 if HOSTNAME:
-    if not HOSTNAME.startswith('http://') and not HOSTNAME.startswith('https://'):
+    if DOMAIN_FROM_REQUEST and not HOSTNAME.startswith('/'):
+        # in this mode HOSTNAME can be only subpath
+        raise ImproperlyConfigured('LABEL_STUDIO_HOST must be a subpath if DOMAIN_FROM_REQUEST is True')
+    else if not HOSTNAME.startswith('http://') and not HOSTNAME.startswith('https://'):
         logger.info(
             '! HOST variable found in environment, but it must start with http:// or https://, ignore it: %s', HOSTNAME
         )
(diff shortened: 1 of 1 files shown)


triklozoid:
You can set up HOSTNAME manually, or read it from request, but not both. 
This check intentionally denies enabling `DOMAIN_FROM_REQUEST` if HOSTNAME is already configured. 
The only case when it's possible is when label-studio is deployed to subpath `example.com/something/labelstudio`

oreilles:
This is the condition currenlty being checked:

```python
if DOMAIN_FROM_REQUEST:
    if HOSTNAME and not HOSTNAME.startswith('/'):
        raise ImproperlyConfigured('LABEL_STUDIO_HOST must be a subpath if DOMAIN_FROM_REQUEST is True')
```

This condition will __*always*__ raise when `DOMAIN_FROM_REQUEST` is set to true, because above is:

```py
if HOSTNAME:
    if not HOSTNAME.startswith('http://') and not HOSTNAME.startswith('https://'):
        HOSTNAME = ''
```

If you set HOSTNAME to a string starting with`/` it will be reset to an empty string by this block. But `DOMAIN_FROM_REQUEST` requires that `HOSTNAME` start with `/`, so it will never work.

Either `DOMAIN_FROM_REQUEST` can be used with `HOSTNAME` as a subpath, in which case there is a bug in the current implementation that this PR fixes...
Or `DOMAIN_FROM_REQUEST` cannot be used at all with `HOSTNAME` and the validation and error message should reflect that:

```diff
if DOMAIN_FROM_REQUEST:
-    # in this mode HOSTNAME can be only subpath
-    if HOSTNAME and not HOSTNAME.startswith('/'):
-       raise ImproperlyConfigured('LABEL_STUDIO_HOST must be a subpath if DOMAIN_FROM_REQUEST is True')
- 
+    if HOSTNAME:
+       raise ImproperlyConfigured('LABEL_STUDIO_HOST must be empty if DOMAIN_FROM_REQUEST is True')
```

>  The only case when it's possible is when label-studio is deployed to subpath `example.com/something/labelstudio`

This is exactly what I'm trying to achieve by setting these env, and which doesn't work because of #5704 

```
LABEL_STUDIO_DOMAIN_FROM_REQUEST=true
LABEL_STUDIO_HOST=/something/labelstudio
```

PR #5877, opened 2024-05-13
fix: RND-56: Fix some errors with api schema

niklub:
Some of the parameters provided in swagger decorators are incorrect and break openapi3 validation checks.

### DIFF
--- a/label_studio/tests/sdk/test_tasks.py
+++ b/label_studio/tests/sdk/test_tasks.py
@@ -6,7 +6,6 @@ from label_studio.tests.sdk.common import LABEL_CONFIG_AND_TASKS
 
 pytestmark = pytest.mark.django_db
 from label_studio_sdk import Client
-
 from tests.sdk.utils import sdk_logs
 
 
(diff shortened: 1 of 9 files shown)


label_studio/tests/sdk/test_tasks.py
```diff
@@ -6,6 +6,7 @@
 
 pytestmark = pytest.mark.django_db
 from label_studio_sdk import Client
+
```
jombooth: let's revert the changes here and below in the diff view, they are causing ruff to fail

jombooth (commented):
LGTM modulo getting rid of spurious imports changes

label_studio/tests/sdk/test_tasks.py
niklub: idk why did these changes appear, since I haven't touch it manually but only with `make fmt-all`

jombooth: What directory are you running `make fmt-all` from? Ruff uses the following section order for dependencies:

https://docs.astral.sh/ruff/settings/#lint_isort_section-order

And I wonder if it thinks `tests` is a third-party dependency because there is no `tests` folder in the directory `ruff` is being executed from

niklub: Cool, thanks, it is actually the opposite: another `tests` folder is in the same root directory where I executed the script, so it obfuscated the third-party

jombooth (approved):

PR #5882, opened 2024-05-14
fix: DIA-1116: Deleting predictions timing out

hakan458:

_(check all that apply)_
- [ ] Product design
- [x] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

After evaluating a model on all tasks, and running a second time, we first delete all predictions before starting thew new job. However, this was taking a super long time for projects with a lot of tasks - causing a timeout in some cases. Now we are performing a _raw_delete() on these predictions, which bypasses the ORM. So we need to manually handle any foreign key relationships - which in this case we have one. When we create an annotation from a prediction, the parent_prediction field is set. When deleting through ORM, we are setting that field to null - so doing the same thing here. Operation is much much faster almost instant now.

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/ml_models/models.py
+++ b/label_studio/ml_models/models.py
@@ -2,7 +2,7 @@
 """
 
 from django.conf import settings
-from django.db import connection, models
+from django.db import models
 from django.utils.translation import gettext_lazy as _
 from ml_model_providers.models import ModelProviderConnection
 from projects.models import Project
--- a/label_studio/ml_models/models.py
+++ b/label_studio/ml_models/models.py
@@ -2,12 +2,12 @@
 """
 
 from django.conf import settings
-from django.db import models
+from django.db import connection, models
 from django.utils.translation import gettext_lazy as _
 from ml_model_providers.models import ModelProviderConnection
 from projects.models import Project
 from rest_framework.exceptions import ValidationError
-from tasks.models import Prediction
+from tasks.models import Annotation, Prediction
 
 
 def validate_string_list(value):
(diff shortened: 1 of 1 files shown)


label_studio/ml_models/models.py
```diff
@@ -173,9 +173,16 @@ def error_file_name(self):
...
 
-        Prediction.objects.filter(model_run=self.id).delete()
+        Executing a raw SQL query here for speed. This ignores any foreign key relationships
+        so if another model has a Prediction fk and set to on_delete=CASCADE for example,
+        it will not take affect. The only relationship like this that currently exists
+        is in Annotation.parent_prediction, which we are handling here
+        """
+        predictions = Prediction.objects.filter(model_run=self.id)
+        prediction_ids = [p.id for p in predictions]
+        Annotation.objects.filter(parent_prediction__in=prediction_ids).update(parent_prediction=None)
```
forum-hs, line 184: I think we can do an on_delete update to None/null in django?
parent = models.ForeignKey(Parent, on_delete=models.SET_NULL, null=True)

hakan458: Yes we can, and we already have that. But the docstring explains why we need this
raw_delete is fast, because it skips these checks, so need to do it manually here

forum-hs: oh we are moving away from delete to speedup

forum-hs (approved):

pakelley (approved):

PR #5913, opened 2024-05-22
fix: LEAP-854: NER Annotations disappear from UI when switching between annotations

juliosgarbi:

This PR addresses the issue where NER annotations disappear from the UI when switching between annotations and predictions.

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

When the user is using the HyperText tag with the property `inline` set to false, the HyperText component loads content into an iframe. When the user switches between annotation tabs, the regions are not being cleared from memory. Consequently, the regions are not reloaded properly when switching back to annotations.

Screen-Recording-2024-05-22-at-12 20 29

#### What does this fix?
The fix involves validating the component within the componentWillUnmount method and disposing of the spans. This process cleans the variable, ensuring that self._spans is empty each time the component is loaded, allowing the regions to be reloaded correctly.

Screen-Recording-2024-05-22-at-12 21 23

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit


web/libs/editor/src/tags/object/RichText/model.js
```diff
@@ -179,6 +179,17 @@ const Model = types
         }
       },
 
+      onDispose() {
+        self.regs.forEach((region) => {
+          try {
```
hlomzik: do we expect errors there so we have to catch them?

bmartel (approved):

bmartel:
I have the same question as @hlomzik but outside of that, the rest looks good. Approving based on the resolution of that comment.

web/libs/editor/src/tags/object/RichText/model.js
juliosgarbi: it makes sense @hlomzik, I was trying to make things safer, but seeing it up close there is no reason for this validation, thanks! :)

Gondragos (approved):

MihajloHoma (approved):

PR #5919, opened 2024-05-24
docs: DOC-216: Move LDAP sections to a separate page

caitlinwheeless:
Sections on LDAP should not be under the SSO page. 

This PR affects:
- [X] Enterprise docs
- [ ] Community docs

### DIFF
--- a/docs/source/guide/auth_setup.md
+++ b/docs/source/guide/auth_setup.md
@@ -14,7 +14,7 @@ section: "Manage Your Organization"
 Set up single sign-on using SAML to manage access to Label Studio using your existing Identity Provider (IdP).
 
 !!! error Enterprise
-    SSO authentication are only available in Label Studio Enterprise Edition. If you're using Label Studio Community Edition, see <a href="https://labelstud.io/guide/label_studio_compare.html">Label Studio Features</a> to learn more.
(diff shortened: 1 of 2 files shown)


docs/source/guide/auth_setup.md
```diff
@@ -100,86 +98,10 @@ The details will vary depending on your IdP, but in general you will complete th
...
-
-```bash
-AUTH_LDAP_ORGANIZATION_WORKSPACES='{"Workspace 1":"cn=team1,ou=users,o=60cbc901ec2e8e387a3b2d3e,dc=jumpcloud,dc=com","Workspace 2":"cn=team2,ou=users,o=60cbc901ec2e8e387a3b2d3e,dc=jumpcloud,dc=com"}'
-```
-
-
-## Manage user access only with LDAP or SSO 
-
-If you want to manage Label Studio roles and workspaces entirely with LDAP or single sign-on (SSO), add the following to your environment variable file:
+If you want to manage Label Studio roles and workspaces entirely with single sign-on (SSO), add the following to your environment variable file:
```
makseq, line 104: btw, let's add here to "Manage user access only with SSO", that if a customer uses SaaS, he can ask us to enable this flags for his organization (we can do it via billing settings). Also we should mention here, that we can disable "Common login" by their request and users must use SSO login only. Common login will be completely disabled.

makseq (changes requested):

docs/source/guide/auth_setup.md
caitlinwheeless: Added a note

makseq (approved):

PR #5933, opened 2024-05-30
feat: DIA-1075: Adding base_model column to ModelInterface to indicate which model from the model provider this ModelInterface uses

wesleylima:
This enabled flexibility in picking different "base models" from an API provider. (like OpenAPI's `gpt-3.5-turbo` )


label_studio/ml_models/migrations/0006_modelinterface_base_model.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.24 on 2024-05-30 18:11
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ml_models', '0005_auto_20240319_1738'),
```
pakelley, line 9: I actually think it would be best to add this to the `ThirdPartyModelVersion` class: the base model can change from version to version (e.g. I want to test out a version with the GPT5 to see how much better it might be), and the "base model" would be an attribute of what we've been referring to as a "3rd party model" (where 1st-party would be one trained in-house by a user, and a 2nd-party model would be trained in-house by Human Signal)

PR #5974, opened 2024-06-11
fix: RND-90: Create lsdk>=1 integration tests

niklub:
Support sdk integration tests with label-studio-sdk>=1 interface


jombooth (approved):
hard to argue with this one - nice work!

PR #5975, opened 2024-06-12
fix: RND-94: Fix API docs for actions.create() and prepare params in DM

makseq:
- [x] Docs have been added/updated (for bug fixes/features)

### DIFF
--- a/label_studio/data_manager/prepare_params.py
+++ b/label_studio/data_manager/prepare_params.py
@@ -162,20 +162,10 @@ class Type(CustomEnum):
 example_request_1 = {
     'filters': {
         'conjunction': 'or',
-        'items': [
-            {
-                'filter': 'filter:tasks:id',
-                'operator': 'greater',
-                'type': 'Number',
-                'value': 123
-            }
-        ]
+        'items': [{'filter': 'filter:tasks:id', 'operator': 'greater', 'type': 'Number', 'value': 123}],
     },
-    'selectedItems': {
-        'all': True,
-        'excluded': [124, 125, 126]
-    },
-    'ordering': ['tasks:total_annotations']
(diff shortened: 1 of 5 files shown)


label_studio/data_manager/prepare_params.py
```diff
@@ -41,3 +42,249 @@ class PrepareParams(BaseModel):
...
+)
+
+prepare_params_schema = openapi.Schema(
+    type=openapi.TYPE_OBJECT,
+    properties={
+        'filters': filters_schema,
+        'selectedItems': selected_items_schema,
+        'ordering': ordering_schema
+    },
+    description='Data payload containing task filters, selected task items, and ordering'
```
niklub: Can we add examples here? similar to https://github.com/HumanSignal/label-studio/blob/develop/label_studio/tasks/openapi_schema.py#L70
Then they would become visible under `-d` here:
image

niklub (changes requested):

label_studio/tasks/api.py
```diff
@@ -119,7 +119,8 @@
                 type=openapi.TYPE_STRING,
                 in_=openapi.IN_QUERY,
                 description='Additional query to filter tasks. It must be JSON encoded string of dict containing '
-                'one of the following parameters: `{"filters": ..., "selectedItems": ..., "ordering": ...}`\n\n'
+                'one of the following parameters: `{"filters": ..., "selectedItems": ..., "ordering": ...}`. '
+                'Check Data Manager > Create View for more details about filters, selectedItems and ordering.\n\n'
```
niklub: where is that reference?

niklub (changes requested):

niklub (approved):

PR #5976, opened 2024-06-12
fix: RND-93: Fix response schema for delete_all_tasks

niklub:
Response schema without content for `delete_all_tasks()`


niklub:
Bypassing the rules as it only introduces schema fix

PR #5982, opened 2024-06-13
fix: RND-97: Support for multiple ML backend predictions per task

niklub:
- Support for multiple predictions response from ML backend. Current version only supports ML backend /predict response with the following payload: `"results": [{}, {}, ...]` where it expects the list of the predictions of the same length as the number of submitted tasks. Now it can optionally handle the case when ML backend responds with `"results": [ [{}, {}], ...]` with each item represents the list of Predictions per task (e.g. with different model versions)
- refactor multiple calls to ML backend to simplify the code (most of the changes)
- added tests for single and multiple predictions, on task opening as using "actions" with sdk

### DIFF
--- a/label_studio/ml/models.py
+++ b/label_studio/ml/models.py
@@ -1,8 +1,9 @@
 """This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
 """
 import logging
+from typing import Dict, List
 
-from core.utils.common import conditional_atomic, db_is_not_sqlite, load_func, safe_float
+from core.utils.common import conditional_atomic, db_is_not_sqlite, load_func
 from django.conf import settings
 from django.db import models, transaction
 from django.db.models import Count, JSONField, Q
@@ -11,7 +12,6 @@ from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
 from ml.api_connector import PREDICT_URL, TIMEOUT_PREDICT, MLApi
 from projects.models import Project
-from tasks.models import Prediction
 from tasks.serializers import PredictionSerializer, TaskSimpleSerializer
 from webhooks.serializers import Webhook, WebhookSerializer
 
@@ -258,6 +258,60 @@ class MLBackend(models.Model):
             },
         }
 
+    def _get_predictions_from_ml_backend(self, serialized_tasks: List[Dict]) -> List[Dict]:
+        result = self.api.make_predictions(serialized_tasks, self.project)
+
+        # response validation
+        if result.is_error:
+            logger.error(f'Error occurred: {result.error_message}')
+            return []
+        elif not isinstance(result.response, dict) or 'results' not in result.response:
+            logger.error(f'ML backend returns an incorrect response, it must be a dict: {result.response}')
+            return []
+        elif not isinstance(result.response['results'], list) or len(result.response['results']) == 0:
+            logger.error('ML backend returns an incorrect response, it must be a list with at least one result')
+            return []
+
+        responses = result.response['results']
+
+        predictions = []
+        if len(responses) == 1 and len(serialized_tasks) > 1:
+            # In case ML backend doesn't support batch of tasks, do it one by one
+            # TODO: remove this block after all ML backends will support batch processing
+            logger.warning(
+                f"'ML backend '{self.title}' doesn't support batch processing of tasks, "
+                f'switched to one-by-one task retrieval'
+            )
+            for serialized_task in serialized_tasks:
+                # get predictions per task
+                predictions.extend(self._get_predictions_from_ml_backend([serialized_task]))
+
+            return predictions
+
+        # ML backend supports batch processing
+        for task, response in zip(serialized_tasks, responses):
+            if isinstance(response, dict):
+                # ML backend can return single prediction per task or multiple predictions
+                response = [response]
+
+            # get  all predictions per task
+            for r in response:
+                if 'result' not in r:
+                    logger.error(
+                        f"ML backend returns an incorrect prediction, it should be a dict with the 'result' field:"
+                        f' {r}'
+                    )
+                    continue
+                predictions.append(
+                    {
+                        'task': task['id'],
+                        'result': r['result'],
+                        'score': r.get('score'),
+                        'model_version': r.get('model_version', self.model_version),
+                    }
+                )
+        return predictions
+
     def predict_tasks(self, tasks):
         model_version = self.update_state()
         if self.not_ready:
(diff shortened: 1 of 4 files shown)


CrisRodriguez:
Hi @niklub,

I would want to ask your help on something. I tried to launch a Github codespace instance using the Dev container configuration inside Label Studio project. However, it does not allow me to run the tests in the local machine (it does not even contain pip so I can do pip install .[test]). Am I missing something? Could you please tell me how you do to get an environment to dev and launch locally tests with a Github codespace instance?

Thanks,

niklub:
> Hi @niklub,
> 
> I would want to ask your help on something. I tried to launch a Github codespace instance using the Dev container configuration inside Label Studio project. However, it does not allow me to run the tests in the local machine (it does not even contain pip so I can do pip install .[test]). Am I missing something? Could you please tell me how you do to get an environment to dev and launch locally tests with a Github codespace instance?
> 
> Thanks,

Hi @CrisRodriguez !

Are you referring to the particular issue related to this PR? If your question is more general, I'd recommend asking in our [Slack](https://slack.labelstud.io/?source=github-1)

label_studio/ml/models.py
```diff
@@ -277,100 +331,13 @@ def predict_tasks(self, tasks):
...
-                return
-            predictions.append(
-                {
-                    'task': task['id'],
-                    'result': response['result'],
-                    'score': response.get('score'),
-                    'model_version': response.get('model_version', self.model_version),
-                }
-            )
+        predictions = self._get_predictions_from_ml_backend(tasks_ser)
```
makseq, line 361: pls put it behind the FF

label_studio/ml/models.py
```diff
@@ -277,100 +331,13 @@ def predict_tasks(self, tasks):
...
         with conditional_atomic(predicate=db_is_not_sqlite):
             prediction_ser = PredictionSerializer(data=predictions, many=True)
             prediction_ser.is_valid(raise_exception=True)
             instances = prediction_ser.save()
         return instances
 
-    def predict_one_task(self, task, model_version):
```
makseq, line 330: we have to do something with predict_one_task in lse:
lse_ml::functions.py::run_single_task_prediction

label_studio/ml/models.py
```diff
@@ -258,6 +258,60 @@ def _predict(self, task):
...
+
+            return predictions
+
+        # ML backend supports batch processing
+        for task, response in zip(serialized_tasks, responses):
+            if isinstance(response, dict):
+                # ML backend can return single prediction per task or multiple predictions
+                response = [response]
+
+            # get  all predictions per task
```
makseq:
```suggestion
            # get all predictions per task
```

label_studio/ml/models.py
```diff
@@ -258,6 +258,60 @@ def _predict(self, task):
...
+                        f"ML backend returns an incorrect prediction, it should be a dict with the 'result' field:"
+                        f' {r}'
+                    )
+                    continue
+                predictions.append(
+                    {
+                        'task': task['id'],
+                        'result': r['result'],
+                        'score': r.get('score'),
+                        'model_version': r.get('model_version', self.model_version),
```
makseq, line 336: project must be specified, we use prediction.project actively for filtering and etc

```suggestion
                        'model_version': r.get('model_version', self.model_version),
                        'project': task['project']
```

label_studio/ml/models.py
```diff
@@ -277,100 +331,13 @@ def predict_tasks(self, tasks):
...
-                logger.debug(f'ML backend {self} is not ready to predict {task}')
-                return
-
-        if task.predictions.filter(model_version=self.model_version).exists():
-            # prediction already exists
-            logger.info(
-                f'Skip creating prediction with ML backend {self} for task {task}: model version '
-                f'{self.model_version} is up-to-date'
-            )
-            return
```
makseq, line 343: what about these 2 things? they are missed in the new code

label_studio/ml/models.py
```diff
@@ -277,100 +331,13 @@ def predict_tasks(self, tasks):
...
-        elif len(responses) == 1 and len(tasks) != 1:
-            logger.warning(
-                f"'ML backend '{self.title}' doesn't support batch processing of tasks, "
-                f'switched to one-by-one task retrieval'
-            )
-            instances = [self.predict_one_task(task, model_version=model_version) for task in tasks]
-            return instances
-
-        # wrong result number
-        elif len(responses) != len(tasks_ser):
```
makseq, line 305: this check len(responses) == 1 and len(tasks) != 1 this check may be useful - it helps to avoid user mistakes

makseq (changes requested):

label_studio/ml/models.py
niklub: why do we need it? additionally, @luarmr is currently working on ff cleanup, I'd not introduce any additional flag at the moment, especially for those changes that don't bring any new functionality

label_studio/ml/models.py
niklub: I'll change it

label_studio/ml/models.py
niklub: isn't it already supported in `Prediction.save()` ?

label_studio/ml/models.py
niklub: I don't think they are lost:

- 1st "if": https://github.com/HumanSignal/label-studio/blob/fb-RND-97/label_studio/ml/models.py#L316
- 2nd "if": https://github.com/HumanSignal/label-studio/blob/fb-RND-97/label_studio/ml/models.py#L327

am I missing something?

label_studio/ml/models.py
niklub: We have this check https://github.com/HumanSignal/label-studio/blob/fb-RND-97/label_studio/ml/models.py#L278

or are you talking about `len(responses) != len(tasks_ser)` ?

CrisRodriguez:
> > Hi @niklub,
> > I would want to ask your help on something. I tried to launch a Github codespace instance using the Dev container configuration inside Label Studio project. However, it does not allow me to run the tests in the local machine (it does not even contain pip so I can do pip install .[test]). Am I missing something? Could you please tell me how you do to get an environment to dev and launch locally tests with a Github codespace instance?
> > Thanks,
> 
> Hi @CrisRodriguez !
> 
> Are you referring to the particular issue related to this PR? If your question is more general, I'd recommend asking in our [Slack](https://slack.labelstud.io/?source=github-1)

Thanks for the link.
Best,

label_studio/ml/models.py
niklub: @makseq https://github.com/HumanSignal/label-studio-enterprise/pull/5851/commits/d3d04cd5abb8435cd56b195db12a825014592f1f

label_studio/ml/models.py
makseq: yes, but with warning - 
image
it seems it is better to set it explicitly

label_studio/ml/models.py
makseq: I mean this : `len(responses) != len(tasks_ser)`
Imagine this: 

```
def predict(tasks): 
  results = []
  for task in tasks:
    try: 
      results.append(mode.predict(task))
    except:  # sometimes ml models can fail
      continue

  return results 
```

len(results) != len(task)

niklub: The warning logs won't help too much, as in this situation, the task. IDs will be misaligned with predictions anyway. We should either completely fail here, or just ignore bad code on the ML backend side. I'd prefer the first approach instead of how it was implemented before.

makseq (approved):

label_studio/ml/models.py
```diff
@@ -258,6 +258,70 @@ def _predict(self, task):
...
+
+        # response validation
+        if result.is_error:
+            logger.error(f'Error occurred: {result.error_message}')
+            return []
+        elif not isinstance(result.response, dict) or 'results' not in result.response:
+            logger.error(f'ML backend returns an incorrect response, it must be a dict: {result.response}')
+            return []
+        elif not isinstance(result.response['results'], list) or len(result.response['results']) == 0:
+            logger.error('ML backend returns an incorrect response, it must be a list with at least one result')
```
jombooth:
```suggestion
            logger.error('ML backend returns an incorrect response, results field must be a list with at least one result')
```

jombooth: nit: It might be preferable to just go ahead and define a DRF serializer for this, instead of hand-rolling validation

label_studio/ml/models.py
makseq: I believe we have to do lots of validations here and it would be one of them..

makseq (approved):

label_studio/ml/models.py
```diff
@@ -258,6 +258,70 @@ def _predict(self, task):
...
+
+                return predictions
+            else:
+                # complete failure - likely ML backend skipped some tasks, we can't match them
+                logger.error(
+                    f'Number of tasks and responses are not equal: '
+                    f'{len(serialized_tasks)} tasks != {len(responses)} responses. '
+                    f'Returning empty predictions.'
+                )
+                return []
```
jombooth: This function is a bit long - could we maybe factor out this part I've highlighted to a helper method?

label_studio/tests/ml/test_predict.py
```diff
@@ -0,0 +1,151 @@
...
+                </View>""",
+            title='test_get_single_prediction_on_task',
+        ),
+        user=business_client.user,
+        use_ml_backend=False,
+    )
+
+    make_task({'data': {'text': 'test 1'}}, project)
+    make_task({'data': {'text': 'test 2'}}, project)
+    make_task({'data': {'text': 'test 3'}}, project)
```
jombooth: Why are three tasks created in this + the other testcase? It's not obvious to me how the downstream test code depends on this happening.

label_studio/tests/sdk/test_ml.py
```diff
@@ -0,0 +1,198 @@
+import pytest
+from label_studio_sdk.client import LabelStudio
+
+from label_studio.tests.utils import register_ml_backend_mock
+
+
+@pytest.fixture
+def ml_backend_for_test_batch_predictions(ml_backend):
```
jombooth: Is this any different to the pytest fixture defined in test_predict.py? If not, I think it'd be better to have one fixture for both test suites to use.

label_studio/ml/models.py
niklub: Agree that the code looks clunky. However we are currently using DRF serializers around data models stored in db, and this response is coming from connected ML backend. Additionally, it would make the logic less clear. Refining how we connect and receive predictions from the model first is better than making the code more concise. It is not part of the work in this PR, and I don't want to increase the complexity of refactoring here as the current change does not deteriorate the current state anyway.

label_studio/ml/models.py
niklub: fixed

label_studio/tests/ml/test_predict.py
niklub: fixed

label_studio/tests/sdk/test_ml.py
niklub: fixed

label_studio/ml/models.py
jombooth: Fair, but fwiw DRF serializers don't have to be model-based only and I've found them really useful for implementing all kinds of request validations in a declarative and maintainable way

jombooth:
LGTM

PR #5991, opened 2024-06-14
fix: Return users in project.all_members()

makseq:
_(check all that apply)_
- [x] Backend (API)

project.all_members() returned member ids instead of user ids

#### What does this fix?
all_members() returns user ids now. 

Data manager filters display the correct user list. 

Data manager filters displayed empty user items. 
image

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/label_studio/projects/mixins.py
+++ b/label_studio/projects/mixins.py
@@ -1,7 +1,8 @@
-from typing import Mapping, Optional, List
+from typing import List, Mapping, Optional
+
+from django.utils.functional import cached_property
 
 from core.redis import start_job_async_or_sync
-from django.utils.functional import cached_property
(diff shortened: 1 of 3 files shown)


triklozoid (approved):

gdoermann:
This works for me. Thank you Max!

label_studio/projects/mixins.py
```diff
@@ -85,9 +86,9 @@ def _can_use_overlap(self):
         return True
 
     @cached_property
-    def all_members(self):
+    def all_members(self) -> List[User]:
```
jombooth: This type should be `QuerySet[User]` not `List[User]`

PR #5993, opened 2024-06-14
docs: DOC-214: Prompter MVP docs

caitlinwheeless:
Initial docs for Prompts tool

#### Affects
- [ ] Community docs
- [X] Enterprise docs


docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+meta_title: Prompts overview
+meta_description: An overview of the Prompts feature in the HumanSignal platform
+section: Prompts
+date: 2024-05-15 14:30:14
+---
+
+
+Use Prompts to evaluate and refine your LLM prompts, and then generate predictions to automate your labeling process. 
+
+All you need to get started is an OpenAI API key and a dataset with some ground truth annotations.
```
shereezhang: we can remove the GT annotation restriction now :)

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+section: Prompts
+date: 2024-05-15 14:30:14
+---
+
+
+Use Prompts to evaluate and refine your LLM prompts, and then generate predictions to automate your labeling process. 
+
+All you need to get started is an OpenAI API key and a dataset with some ground truth annotations. 
+
+With Prompts, you can:
```
shereezhang, line 20: we can add a bullet point about leveraging subject matter expertise to rapidly bootstrap projects with labels to get started with ML development faster

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+Prompts allows you to leverage LLMs to swiftly generate accurate predictions, enabling instant labeling of thousands of tasks. 
+
+By utilizing AI to handle the bulk of the annotation work, you can significantly enhance the efficiency and speed of your data labeling workflows. This is particularly valuable when dealing with large datasets that require consistent and accurate labeling. Automating this process reduces the reliance on manual annotators, which not only cuts down on labor costs but also minimizes human errors and biases. With AI's ability to learn from the provided ground truth annotations, you can maintain a high level of accuracy and consistency across the dataset, ensuring high-quality labeled data for training machine learning models.
+
+
+
+## Prompt evaluation and fine-tuning
+
+!!! info Tip
+    While you can use Prompts to do basic prompt evaluation and fine-tuning, we recommend using [Evals](evals_overview), which has been specifically designed for this purpose and has much more robust analytics.
```
shereezhang: i wouldn't say Evals is designed for finetuning, but i'd agree it's better for prompt evaluation. As such, i'd suggest perhaps emphasize Prompts as the interface for building prompts, and evals as the tool to evaluate - maybe wording to something along the lines of -
While you can use Prompts to conduct prompt fine-tuning and basic evaluations, we recommend using Evals for deeper analyses on LLM prompt performance and evaluations.

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+    While you can use Prompts to do basic prompt evaluation and fine-tuning, we recommend using [Evals](evals_overview), which has been specifically designed for this purpose and has much more robust analytics. 
+
+As you evaluate your prompt against the ground truth annotations, we return an accuracy score for each version of your prompt. You can use this to iterate your prompt versions for [clarity, specificity, and context](prompts_draft#Drafting-effective-prompts). 
+
+![Screenshot of accuracy score](/images/prompts/accuracy_score.png)
+
+This accuracy score provides a measurable way to evaluate and refine the performance of your prompt. By tracking accuracy, you can ensure that the automated labels generated by the LLM are consistent with ground truth data. 
+
+This feedback loop allows you to iteratively fine-tune your prompts, optimizing the accuracy of predictions and enhancing the overall reliability of your data annotation processes. In industries where data accuracy directly impacts decision-making and operational efficiency, this capability is invaluable.
+
```
shereezhang, line 67: Maybe one more section on 'Bootstrapping Projects with Prompts'
This is for the new feature where no GT is needed, and Prompts will not conduct evaluations like accuracy, but the user can generate predictions with a prompt and convert them to annotations.

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+## Model types
+
+### Text classification 
+
+At present, the Prompts feature only supports text classification labeling tasks.
```
shereezhang: specifically 'single label text classification'

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+Text classification labeling tasks are fundamental in many applications, enabling efficient data organization, improving searchability, and providing valuable insights through data analysis. Some examples include:
+
+* **Spam Detection**: Classifying emails as "spam" or "not spam."
+* **Sentiment Analysis**: Categorizing user reviews as "positive," "negative," or "neutral."
+* **Topic Categorization**: Assigning articles to categories like "politics," "sports," "technology," etc.
+* **Support Ticket Classification**: Labeling customer support tickets based on the issue type, such as "billing," "technical support," or "account management."
+* **Content Moderation**: Identifying and labeling inappropriate content on social media platforms, such as "offensive language," "hate speech," or "harassment."
+
+
+## Workflow
```
shereezhang: do we need a workflow for the bootstrapping mode? the only steps that diverge are 2 (GT not needed), 4 (no evaluate step) and 5 (no accuracy, just vibe check)

docs/source/guide/prompts_model.md
```diff
@@ -0,0 +1,47 @@
...
+---
+
+
+## Prerequisites
+
+* An OpenAI API key. 
+* You need to have a project that meets the following criteria:
+  * Text-based data set (meaning you are annotating text and not image or video files). 
+  * The labeling configuration for the project must be set up to use single-class classification (`choice="single"`). 
+  * You should have at least one task with a [ground truth annotation](quality#Define-ground-truth-annotations-for-a-project). Only tasks with a ground truth will be listed in the prompt interface.
```
shereezhang: maybe this can be conditional - like "if you want to evaluate for accuracy for your prompt, you'll need GT"

docs/source/guide/prompts_predictions.md
```diff
@@ -0,0 +1,46 @@
...
+order_enterprise: 234
+meta_title: Generate predictions
+meta_description: Generate predictions to auto-label project tasks
+section: Prompts
+date: 2024-06-13 21:54:22
+---
+
+You can continue to save and evaluate your prompts until you reach your desired level of accuracy. When your prompt is ready, you can use it to generate predictions for all tasks within your project. 
+
+!!! note
```
shereezhang, line 16: i'm not sure i understand this note? would i be able to get some more context?

docs/source/guide/prompts_predictions.md
```diff
@@ -0,0 +1,46 @@
...
+## Create annotations from predictions
+
+Once you have your predictions in place, you still need to convert them to annotations. You can review predictions by opening tasks. The predictions are listed under the model name and are grayed out: 
+
+![Screenshot of the prediction preview](/images/prompts/prediction.png)
+
+
+From the Data Manager, select all the tasks you want to label and then select **Actions > Create Annotations from Predictions**. You are asked to select the model and version you want to use. 
+
+![Gif of the of create annotations action](/images/prompts/create_annotations_1.png)
```
shereezhang: maybe just me but this gif didnt render on my screen :(

caitlinwheeless: That's because I didn't change the file extension. I'll fix!

docs/source/guide/prompts_predictions.md
caitlinwheeless: I don't think I wrote the note well, and I was missing a bit of context. I was trying to explain why their ground truth tasks might have multiple predictions. 
For example, in the demo project I'm currently working in, each ground truth task I have has 3 predictions:

1 prediction is from Model A, from when I clicked "Evaluate"
1 prediction is from Model B, from when I clicked "Evaluate"
1 prediction is from Model A, from when I clicked "Get Predictions for All Tasks" <-- this scenario I didn't include, because I'm not sure if that's the expected behavior.

docs/source/guide/prompts_overview.md
pakelley: Another thing here is that we should probably be careful to say "project" here instead of "dataset" (since you can't use Prompts with a Data Discovery "Dataset" yet)

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+Use Prompts to evaluate and refine your LLM prompts, and then generate predictions to automate your labeling process. 
+
+All you need to get started is an OpenAI API key and a dataset with some ground truth annotations. 
+
+With Prompts, you can:
+
+* Drastically increase the speed and efficiency of annotations, transforming subject matter experts (SMEs) into highly productive data scientists while reducing the dependency on non-SME manual annotators.
+* Enhance annotation throughput and accuracy, making the process faster and more scalable.
```
pakelley: nit: I think something like "Increase" might be more accurate than "Enhance" here. I know we just used "Drastically increase" though, so maybe "improve" or something?

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+* Enhance annotation throughput and accuracy, making the process faster and more scalable. 
+* Empower users to harness the full potential of AI-driven text labeling, setting a new standard for efficiency and innovation in data labeling.
+
+## Auto-labeling with Prompts
+ 
+Prompts allows you to leverage LLMs to swiftly generate accurate predictions, enabling instant labeling of thousands of tasks. 
+
+By utilizing AI to handle the bulk of the annotation work, you can significantly enhance the efficiency and speed of your data labeling workflows. This is particularly valuable when dealing with large datasets that require consistent and accurate labeling. Automating this process reduces the reliance on manual annotators, which not only cuts down on labor costs but also minimizes human errors and biases. With AI's ability to learn from the provided ground truth annotations, you can maintain a high level of accuracy and consistency across the dataset, ensuring high-quality labeled data for training machine learning models.
```
pakelley, line 36: The point here about consistency for large datasets is actually a really good point. 
I'm on the fence, but it almost seems worth calling out in the bullet points above (i.e. we have a lot of features dedicated to comparing labels across annotators, and users spend a lot of time/energy ensuring consistency. Having one entity label large amounts of data might be a really nice selling point. Though it's balanced against the fact that we generally don't generally consider LLMs accurate yet)

docs/source/guide/prompts_overview.md
pakelley: Imo it might be helpful to put this in an "info" admonition too. Just my opinion though, if this doesn't seem right to you

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Text classification 
+
+At present, the Prompts feature only supports text classification labeling tasks.  
+
+Text classification is the process of assigning predefined categories or labels to segments of text based on their content. This involves analyzing the text and determining which category or label best describes its subject, sentiment, or purpose. The goal is to organize and categorize textual data in a way that makes it easier to analyze, search, and utilize. 
+
+Text classification labeling tasks are fundamental in many applications, enabling efficient data organization, improving searchability, and providing valuable insights through data analysis. Some examples include:
+
+* **Spam Detection**: Classifying emails as "spam" or "not spam."
```
pakelley: Generally "not spam" is called "ham" in this context, if we want to use that term. Definitely not the hill I'd die on though.

docs/source/guide/prompts_overview.md
pakelley: Yep, I don't think we've filled Caitlin in on this, but definitely seems important to capture in the docs (as I think it's likely to be one of the more confusing parts of the workflow).
Basically, there are two different styles of workflow:
1. You use a dataset with Ground Truth, and follow the workflow we have outlined here
2. You use a dataset without Ground Truth (or use a project with Ground Truth, but change the baseline you're using to "Sample 20" or "All"), and then you don't have accuracy displayed (i.e. both the accuracy column and the accuracy summary stats at the bottom will be omitted). The user workflow is as Sheree described, but I do think we should point out that you can change the baseline (and that you _can_ use this workflow for projects with Ground Truth if you change the baseline)

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+* **Sentiment Analysis**: Categorizing user reviews as "positive," "negative," or "neutral."
+* **Topic Categorization**: Assigning articles to categories like "politics," "sports," "technology," etc.
+* **Support Ticket Classification**: Labeling customer support tickets based on the issue type, such as "billing," "technical support," or "account management."
+* **Content Moderation**: Identifying and labeling inappropriate content on social media platforms, such as "offensive language," "hate speech," or "harassment."
+
+
+## Workflow
+
+1. Create a project and import a text-based dataset.
```
pakelley: I think it may be worth calling out something like "If you don't have one already, create a project..." to make sure users don't think they need to make a separate project to use Prompts

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+1. Create a project and import a text-based dataset. 
+
+    * [Create a project](setup_project)
+    * [Sync data from external storage](storage)
+2. Annotate a subset of tasks, marking as many as possible as ground truth annotation. The more ground truths you have in your dataset, the better your results will be. 
+
+    * [Labeling guide](labeling)
+    * [Define ground truth annotations for a project](quality#Define-ground-truth-annotations-for-a-project)
+    * [Blog - What's a ground truth dataset?](https://humansignal.com/blog/what-s-a-ground-truth-dataset/)
+3. Go to the Prompts page and create a new model. The first time you create a model, you will be asked to provide an OpenAI API key.
```
pakelley: Splitting hairs here, but technically "... you will be asked to provide an OpenAI API key if you haven't set one up already." (There's a separate button to add an API key, though I figure most people will add it by clicking "Create Model" rather than adding their API key manually)

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+    * [Define ground truth annotations for a project](quality#Define-ground-truth-annotations-for-a-project)
+    * [Blog - What's a ground truth dataset?](https://humansignal.com/blog/what-s-a-ground-truth-dataset/)
+3. Go to the Prompts page and create a new model. The first time you create a model, you will be asked to provide an OpenAI API key. 
+
+    * [Create a model](prompts_model)
+    * [Where do I find my OpenAI API Key?](https://help.openai.com/en/articles/4936850-where-do-i-find-my-openai-api-key)
+4. Write a prompt and evaluate it against your ground truth dataset. 
+
+    * [Draft a prompt](prompt_draft)
+5. When your prompt is returning an overall accuracy that is acceptable, you can apply it to the rest of the tasks in your project.
```
pakelley: Can we say "...you can choose to apply it..." to avoid implying that there is some check on our end that keeps users from applying it when their accuracy is low?

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,105 @@
...
+
+| Feature | Support |
+| --- | --- |
+| **Supported data types** | Text |
+| **Supported model types** | Text Classification |
+| **Class selection** | Single selection (the LLM can apply one label per task)|
+| **Supported base models** | OpenAI gpt-3.5-turbo-16k <br>OpenAI gpt-3.5-turbo-instruct <br>OpenAI gpt-4-turbo <br>OpenAI gpt-3.5-turbo <br>OpenAI gpt-4o <br>OpenAI gpt-4 |
+| **Number of API keys per org** | 1 (OpenAI API key only) |
+| **Required permissions** | **Owners, Administrators, Managers** -- Can create Prompt models and update projects with auto-annotations. Managers can only apply models to projects in which they are already a member. <br><br>**Reviewers and Annotators** -- No access to the Prompts tool, but can see the predictions generated by the prompts from within the project (depending on your [project settings](project_settings_lse)).  |
+| **Enterprise vs. Open Source** | Label Studio Enterprise only |
```
pakelley: Might be worth documenting that on-prem isn't currently supported

docs/source/guide/prompts_model.md
```diff
@@ -0,0 +1,47 @@
...
+meta_description: How to create a Prompts model
+section: Prompts
+date: 2024-06-11 16:53:16
+---
+
+
+## Prerequisites
+
+* An OpenAI API key. 
+* You need to have a project that meets the following criteria:
```
pakelley: nit: might be worth matching the style of the first bullet with something like "A project that meets the following criteria:"

docs/source/guide/prompts_model.md
```diff
@@ -0,0 +1,47 @@
...
+  * Text-based data set (meaning you are annotating text and not image or video files). 
+  * The labeling configuration for the project must be set up to use single-class classification (`choice="single"`). 
+  * You should have at least one task with a [ground truth annotation](quality#Define-ground-truth-annotations-for-a-project). Only tasks with a ground truth will be listed in the prompt interface. 
+
+## API key
+
+You can only specify one OpenAI API key per organization, and it only needs to be added once. 
+
+Once added, it is automatically used for all new models.
```
pakelley: "automatically used for all new Prompts"

docs/source/guide/prompts_model.md
```diff
@@ -0,0 +1,47 @@
...
+From the Prompts page, click **Create Model** in the upper right and then complete the following fields:
+
+<div class="noheader rowheader">
+
+| | |
+| --- | --- |
+| Name | Enter a name for the prompt model. |
+| Description | Enter a description for the prompt model.  |
+| Type | Select the model type. At this time, we only support [text classification](prompts_overview#Text-classification). |
+| Target Project| Select the project you want to use. The projects that are listed must have a text-based dataset (meaning you are labeling text, not images).<br><br>You must have access to the project. If you are in the Manager role, you need to be added to the project to have access. <br><br>If you don't see a project listed, then you do not have any eligible projects available. This is likely because the project is set up for multiple selection (e.g. `choice="multiple"`), is using a data type that isn't supported (video, audio), or you do not have access to the project.   |
```
pakelley: I think we should phrase this as "The projects listed will only include text-based projects which you have access to. If you are in the Manager role..." Wdyt?
Also, I think we should phrase the part about video/audio/etc as something like "doesn't include text data, e.g. it only includes data types that aren't supported (image, video, audio)...". Users could use projects that have video/etc if they want, as long as there's _also_ text data.

docs/source/guide/prompts_model.md
```diff
@@ -0,0 +1,47 @@
...
+| Model Classes | This list is generated from the labeling configuration of the target project. |
```
pakelley: nit: "This list is automatically generated from the..."

docs/source/guide/prompts_draft.md
```diff
@@ -0,0 +1,54 @@
...
+type: guide
+order: 0
+order_enterprise: 231
+meta_title: Draft a prompt
+meta_description: Create and evaluate an LLM prompt
+section: Prompts
+date: 2024-06-12 14:09:09
+---
+
+With your [model created](prompts_model), you can begin drafting prompts.
```
pakelley: I figure this should be "With your Prompt created", but I do see how the wording gets awkward...

docs/source/guide/prompts_draft.md
```diff
@@ -0,0 +1,54 @@
...
+
+With your [model created](prompts_model), you can begin drafting prompts. 
+
+## Draft a prompt and generate predictions
+
+
+1. Select your base model. For a description of all OpenAI models, see [OpenAI's models overview](https://platform.openai.com/docs/models/models-overview).
+2. In the **Prompt** field, enter your prompt. Keep in mind the following:
+    * You must include the text class. (In the demo below, this is the `review` class.) Click the text class name to insert it into the prompt. 
+    * Although not strictly required, you should provide definitions for each class.
```
pakelley: I think it'd be helpful to clarify that the reason to provide these definitions is for model accuracy. Maybe something like "...you should provide definitions for each class, to ensure prediction accuracy."

docs/source/guide/prompts_draft.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. Select your base model. For a description of all OpenAI models, see [OpenAI's models overview](https://platform.openai.com/docs/models/models-overview).
+2. In the **Prompt** field, enter your prompt. Keep in mind the following:
+    * You must include the text class. (In the demo below, this is the `review` class.) Click the text class name to insert it into the prompt. 
+    * Although not strictly required, you should provide definitions for each class. 
+3. Click **Save**. 
+4. Click **Evaluate**. 
+
+!!! note
+    When you click **Evaluate**, you will create predictions for each ground truth task. When you return to the project, you will see this reflected in the project. You can see how many predictions a task has using the **Predictions** column in the Data Manager.
```
pakelley: This will mostly be helpful for clarifying which records will be evaluated when you choose other baselines than "Ground Truth", but I think saying something like "...you will create predictions for each task included in the baseline you've selected" (then maybe linking to the section with info about the two different workflows, depending on the baseline the user has selected)

docs/source/guide/prompts_draft.md
```diff
@@ -0,0 +1,54 @@
...
+3. Click **Save**. 
+4. Click **Evaluate**. 
+
+!!! note
+    When you click **Evaluate**, you will create predictions for each ground truth task. When you return to the project, you will see this reflected in the project. You can see how many predictions a task has using the **Predictions** column in the Data Manager. 
+
+<br><br>
+<video src="../images/prompts/prompts.mp4" controls="controls" style="max-width: 800px;" class="gif-border" />
+
+## Drafting effective prompts
```
pakelley, line 44: Might be worth linking to The Prompt Report here, as that's a pretty popular reference for prompting techniques now: https://arxiv.org/abs/2406.06608

docs/source/guide/prompts_draft.md
pakelley: Also, might be helpful to link to the "Add context" section below, if possible

docs/source/guide/prompts_predictions.md
pakelley: So yeah, this is the expected behavior at the moment (and I think it's worth including a note about it). In the future we want to improve this behavior, but doing it differently gives different weird user experiences, so this is kind of the lesser of the evils atm.
I think phrasing this by directly pointing out the weirdness might be helpful, something like:

"Evaluating your Prompts can result in multiple predictions on your tasks: if you have multiple Prompts for one Project, or if you click both "Evaluate" and "Get Predictions for All Tasks" from a Prompt, you will see multiple predictions for tasks in the Data Manager"

Wdyt?

pakelley: Might even be worth making this a "warning" admonition.

docs/source/guide/prompts_predictions.md
```diff
@@ -0,0 +1,46 @@
...
+!!! note 
+    Predictions are not cumulative when using this action. If you get predictions for all tasks, change the prompt, and then generate predictions for all tasks again, you will overwrite your previous predictions and each task will only reflect the most recent prediction. 
+
+Once complete, you can return to the project and open the Data Manager. Use the **Total predictions per task** column to confirm that each task has at least one prediction:
+
+![Screenshot of the prediction preview](/images/prompts/prediction_column.png)
+
+## Remove predictions
+
+If you prematurely generated predictions or want to use a new prompt, simply select all tasks and select **Actions > Delete Predictions**. To only remove predictions from certain models or model versions, use the **Predictions** page in the project settings.
```
pakelley: Is there a page we can link to that explains how to use the Predictions page for this?

docs/source/guide/prompts_predictions.md
```diff
@@ -0,0 +1,46 @@
...
+
+![Screenshot of the prediction preview](/images/prompts/prediction_column.png)
+
+## Remove predictions
+
+If you prematurely generated predictions or want to use a new prompt, simply select all tasks and select **Actions > Delete Predictions**. To only remove predictions from certain models or model versions, use the **Predictions** page in the project settings. 
+
+## Create annotations from predictions
+
+Once you have your predictions in place, you still need to convert them to annotations. You can review predictions by opening tasks. The predictions are listed under the model name and are grayed out:
```
pakelley: I don't think they necessarily "need" to convert them (if their desired workflow can work with predictions rather than annotations.
Imo we could say something like "...you still need to convert them to annotations if you want to use the predictions along with your other annotations" or something like that. This is pretty ham-fisted still though, and I'm not sure what the best way is to phrase this (might be worth figuring out which downstream workflows require annotations rather than predictions to determine what's best to put here)

docs/source/guide/prompts_overview.md
pakelley, line 1: Not sure where this is exactly this is, but I think we should add the "Beta" flag that we have on Datasets to the Prompts section in the nav sidebar

caitlinwheeless: Good call out - updated the title and EJS file with this

docs/source/guide/prompts_overview.md
caitlinwheeless: Tentatively removing that tip until I can be sure that Evals docs will need to be out

docs/source/guide/prompts_overview.md
```diff
@@ -0,0 +1,148 @@
...
+
+By utilizing AI to handle the bulk of the annotation work, you can significantly enhance the efficiency and speed of your data labeling workflows. This is particularly valuable when dealing with large datasets that require consistent and accurate labeling. Automating this process reduces the reliance on manual annotators, which not only cuts down on labor costs but also minimizes human errors and biases. With AI's ability to learn from the provided ground truth annotations, you can maintain a high level of accuracy and consistency across the dataset, ensuring high-quality labeled data for training machine learning models.
+
+#### Workflow
+
+1. If you don't already have one, create a project and import a text-based dataset. 
+
+    * [Create a project](setup_project)
+    * [Sync data from external storage](storage)
+2. Annotate a subset of tasks, marking as many as possible as ground truth annotations. The more ground truths you have in your dataset, the more accurate your prompt evaluation results will be.
```
shereezhang: "... the more data you have for prompt evaluation for more confident results"
or smthing like that?
just dont wanna guarantee accuracy

shereezhang (approved):
tiny comment, overall amazing, thank you!

PR #5996, opened 2024-06-15
fix: LEAP-1161: fix User import in mixin

jombooth:
Need to late import the User model to avoid circular import issues in LSE. Also some misc formatting fixes from ruff/blue, as well as an issue with a fixture that's not available in some tests.


jombooth:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/9525422576)

PR #6032, opened 2024-06-25
fix: RND-102: Ensure SDK compatibility with LSE

niklub:
- update fixture for sdk/test_ml to avoid SSRF protection
- fix bug with `projects.update()` when using LSE


niklub:
Merging it as it doesn't contain functional changes - only tests and API schema

PR #6078, opened 2024-07-10
fix: OPTIC-875,OPTIC-876,OPTIC-877,OPTIC-878,OPTIC-879: Visual conflicts due to design updates

bmartel:
Fixes design update issues pertaining to the following:

- OPTIC-875 Close button is not visible at Export Modal 
- OPTIC-876 Project settings icons are displayed only on initial load. 
- OPTIC-877 Gray background is missing on pagination bar
- OPTIC-878 LABEL_STUDIO_DISABLE_SIGNUP_WITHOUT_LINK property is not taken into consideration in new app redesign
- OPTIC-879 '+' icon is missing on Add People button

### DIFF
--- a/web/apps/labelstudio/src/components/Button/Button.styl
+++ b/web/apps/labelstudio/src/components/Button/Button.styl
@@ -17,6 +17,7 @@
   --button-margin-left 0
   --button-margin-right 0
   --button-border 1px solid var(--button-color)
+  --button-gap 6px
 
   border var(--button-border)
   cursor pointer
@@ -35,6 +36,7 @@
   height var(--button-height)
   min-width var(--button-min-width)
   padding var(--button-padding)
+  gap var(--button-gap)
   align-items var(--button-content-align)
   justify-content var(--button-content-justify)
   margin-left var(--button-margin-left)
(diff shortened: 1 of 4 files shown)


web/apps/labelstudio/src/components/Button/Button.styl
```diff
@@ -107,10 +107,10 @@
 
   &_type
     &_text
-      padding 0
+      --button-padding 0
```
yyassi-heartex: do we need this? 
```
&_text, &_link
      --button-padding 0
```
would handle this no?

bmartel: yep I do believe you are correct, nice catch!

web/libs/datamanager/src/components/DataManager/Toolbar/instruments.jsx
```diff
@@ -70,20 +70,23 @@ export const instruments = {
     return <ViewToggle size={size} style={style} />;
   },
   columns: ({ size }) => {
-    const iconProps = {};
-    const isNewUI = isFF(FF_LOPS_E_10);
-
-    if (isNewUI) {
+    const iconProps = {
```
yyassi-heartex, line 73: short term this is fine - but ideally I think this kind of stuff needs to get into the css and we toggle on css classes - maybe part of a greater initiative as I see this alot through out our code

yyassi-heartex (approved):

hlomzik (approved):

vladimirheartex (approved):

vladimirheartex (approved):

PR #6079, opened 2024-07-10
feat: LEAP-1199: Add drag and drop functionality on DM tabs

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [x] Backend (API)
- [x] Frontend

Currently, users cannot move tabs around, which is inconvenient. They must delete ("Close") tabs and recreate them in the desired order. This update allows users to reorder tabs by dragging and dropping them.

Key features:

- Drag and drop functionality for tabs.
- Visual indicator to show the drop location.
- Dragging not confined to vertical tab limits.
- Tabs can be reordered regardless of visibility.
- Order of tabs is consistent across the organization after a hard refresh.
- No disruption for concurrent users.
- ESC key cancels the drag action, returning the tab to its original position.

Screen-Recording-2024-07-16-at-18 09 18


triklozoid:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/9957138605)
> Successfully merged:  rename web/dist/apps/labelstudio/{book.d787a03585e3dfbb093186a4f7e882d3.svg => book.23db5e3751679a820c6e975e99f29323.svg} (70%)

juliosgarbi:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/9960418462)

label_studio/data_manager/models.py
```diff
@@ -9,6 +9,7 @@
 class ViewBaseModel(models.Model):
     data = models.JSONField(_('data'), default=dict, null=True, help_text='Custom view data')
     ordering = models.JSONField(_('ordering'), default=dict, null=True, help_text='Ordering parameters')
+    order = models.IntegerField(_('order'), default=0, help_text='Order of view')
```
jombooth: This migration isn't compatible with our migration guidelines - we shouldn't be adding new fields that aren't nullable, because they cause problems if we have to revert.

jombooth (changes requested):
We need to rework the migration that's included in this PR to make the new field nullable; code depending on the new field should handle nulls.

label_studio/data_manager/models.py
jombooth: While we're here, I think the help text could clarify the purpose of this field more; "order of view" could mean a lot of different things. We should spell out that it's the position of the tab, starting at the left in data manager and increasing as the tabs go left to right.

triklozoid: Fixed.

jombooth (approved):

yyassi-heartex (approved):

web/libs/datamanager/src/components/Common/Tabs/Tabs.jsx
```diff
@@ -7,12 +7,13 @@ import { Icon } from "../Icon/Icon";
 import Input from "../Input/Input";
 import "./Tabs.styl";
 import { TabsMenu } from "./TabsMenu";
+import { DragDropContext, Droppable } from "react-beautiful-dnd";
```
hlomzik: could you please move it up to be in order?

web/libs/datamanager/src/components/DataManager/DataManager.jsx
```diff
@@ -10,6 +10,7 @@ import { FiltersSidebar } from "../Filters/FiltersSidebar/FilterSidebar";
 import { DataView } from "../MainView";
 import "./DataManager.styl";
 import { Toolbar } from "./Toolbar/Toolbar";
+import { DragDropContext, Draggable, Droppable } from "react-beautiful-dnd";
```
hlomzik: and here as well

web/libs/datamanager/src/stores/Tabs/store.js
```diff
@@ -333,6 +333,23 @@ export const TabStore = types
       return view;
     }),
 
+    updateViewOrder: flow(function* (source, destination) {
+      // Detach the view from the original position
+      const [removed] = self.views.splice(source, 1);
+      const sn = getSnapshot(removed);
+
+      // Insert the view at the new position
+      self.views.splice(destination, 0, sn);
```
hlomzik, line 342: Probably these manipulations are the reason for errors on tab duplication.
`self.views` is changed in place, so no re-renders happen. So rendered tabs are the same old objects even after updates, because original tab is removed and new one is created from snapshot. So rendered tab doesn't have parent anymore.
To fix this you can first try to update `views` in immutable manner by updating it: `self.views = [...self.views]`. Then all rendered tabs will be the actual ones, but this can slow tabs down. You can also try to change the way you access parent, basically it's just `views`.

hlomzik (changes requested):

web/libs/datamanager/src/stores/Tabs/store.js
juliosgarbi: Yeah, I started using views instead tab.parent, it's easier and it doesn't depend of any reference, I believe that in this case this solution is better. Thanks @hlomzik

web/libs/datamanager/src/components/Common/Tabs/Tabs.jsx
```diff
@@ -37,8 +47,16 @@ export const Tabs = ({ children, activeTab, onChange, onAdd, tabBarExtraContent,
     <TabsContext.Provider value={contextValue}>
       <Block name="tabs">
         <Elem name="list">
-          {children}
-
+          <DragDropContext onDragEnd={onDragEnd}>
+            <Droppable droppableId="droppable" direction="horizontal">
+              {(provided) => (
+                <Elem ref={provided.innerRef} name={"droppable"} {...provided.droppableProps}>
```
hlomzik:
```suggestion
                <Elem ref={provided.innerRef} name="droppable" {...provided.droppableProps}>
```

web/libs/datamanager/src/components/DataManager/DataManager.jsx
```diff
@@ -66,32 +67,56 @@ const TabsSwitch = switchInjector(
...
+              >
+                <TabsItem
+                  key={tab.key}
+                  tab={tab.key}
+                  title={tab.title}
+                  onFinishEditing={(title) => {
+                    tab.setTitle(title);
+                    tab.save();
+                  }}
+                  onDuplicate={() => views.duplicateView(tab)}
```
hlomzik, line 109: so it works like this? even if the `tab` is potentially removed from the tree?
what about closing tab which was moved? and can we also apply this change to `onClose()`? so it will be `views.deleteView(tab)`

juliosgarbi: The tab is always on the tree when we try to duplicate it, even if we move the position and reorder the array, the problem that we were having is that for some reason we were loosing the `tab.parent` reference, and because of this, we couldn't get the duplicateView method from views. Yeah make totally sense, I already applied the same to onClose(), thanks @hlomzik

juliosgarbi:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10061829249)
> Successfully merged:  29 files changed, 302 insertions(+), 96 deletions(-)

hlomzik (approved):

AndrejOros (approved):

RiccardoMaistri:
Good morning!
Does the drag and drop funcitonality has to be enabled or enteprse version is required? I am running Label studio v 1.13.1 deployed using helm chart. I am using Version 130.0.6723.91 (Official Build) (64-bit)

PR #6085, opened 2024-07-11
fix: OPTIC-883: Export data modal footer not visible when scrollable

bmartel:
App redesign had changed a couple of the overflow values on modals, which broke the ability to provide a position sticky footer. This made it difficult to present modal forms with scrolling content and have the form submit button visible at all times.

_Before_
image

_After_
image

### DIFF
--- a/web/apps/labelstudio/src/components/Modal/Modal.styl
+++ b/web/apps/labelstudio/src/components/Modal/Modal.styl
@@ -13,8 +13,7 @@
   justify-content center
   background-color rgba(#000, 0.3)
   will-change opacity
-  overflow hidden
-  
+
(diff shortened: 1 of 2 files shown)


yyassi-heartex (approved):
negative margins are weird - but given its been around for a while we should take another look at that later

hlomzik (approved):

hlomzik:
> negative margins are weird - but given its been around for a while we should take another look at that later

@yyassi-heartex it's a popular workaround with outer paddings you can't control. it would be better without it with better layout for sure, but I think they are just okay

AndrejOros (approved):

PR #6091, opened 2024-07-12
feat: LEAP-1241: Views ordering

triklozoid:
* added `order` field for model View
* added `/api/dm/views/order` API for reordering views for one project
* views returned by `/api/dm/views' are now ordered by `order` field.


label_studio/data_manager/models.py
```diff
@@ -22,6 +23,7 @@ class ViewBaseModel(models.Model):
     )
 
     class Meta:
+        ordering = ['order']
```
wesleylima, line 26: Did we look into how changing the ordering affects the query performance? I think we should make sure that we aren't missing indexes because of the new ORDER BY.

label_studio/data_manager/api.py
```diff
@@ -163,8 +168,39 @@ def reset(self, request):
...
+        serializer = ViewOrderSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        project_id = serializer.validated_data['project']
+        view_ids = serializer.validated_data['ids']
+
+        project = generics.get_object_or_404(Project.objects.for_user(request.user), pk=project_id)
+
+        queryset = self.filter_queryset(self.get_queryset()).filter(project=project)
+        views = list(queryset.filter(id__in=view_ids))
```
wesleylima, line 189: [select_for_update()](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#select-for-update) may be useful here to make sure that the orders being updated are exactly the same as the ones we juts fetched

label_studio/data_manager/models.py
triklozoid: Added index for order field.

label_studio/data_manager/api.py
triklozoid: I'm not sure we need it here. I update only `order` field, in case of race condition of reordering requests, we will get "last db request win" situation anyway. 
If some other user want's to change filters(for example) of one particular tab, why do we need to deny it by locking the record?

wesleylima: Yeah that's a very good point. It would be over-locking for non-sorting/non-conflict causing updates. We could normalize the ordering into a different table to get around this, but that's probably way too excessive for what would likely be a very unlikely edge case.

label_studio/data_manager/models.py
wesleylima: I think a two_column index of `project_id` and `order` would be _best_? I'm not sure the index in just the `order` field does anything? Just a guess, I haven't seen the queryplan for this. It may be an unecessary over-optimization too

triklozoid: views is a relatively small table and we return only views for one project per request, so I don't expect performance problems here, I'm not sure we need to dig into query plan. 
I changed index to project + order.

wesleylima (approved):

triklozoid:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/9956742991)
> Successfully merged:  8 files changed, 183 insertions(+), 89 deletions(-)

PR #6103, opened 2024-07-16
feat: LEAP-1151: Show region indices on regions

hlomzik:
Display region index from Outliner on regions itself, so it will help to find regions visually.
Index is stored in region and updated every time the order changes (new region added, order direction is changed)

- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

With lot of regions and/or with overlapping regions it's hard to select exact region you need. This fix adds region indices from regions tree, so you can select the region you need precisely in regions tree, without interaction with image, just by looking for the number.

It writes `region_index` on every regions tree change, but that's a volatile property + it's updated only if index is changed, so should be a very small effect.

- [x] No

- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/web/libs/editor/src/mixins/AreaMixin.js
+++ b/web/libs/editor/src/mixins/AreaMixin.js
@@ -92,12 +92,15 @@ export const AreaMixinBase = types
       return Array.from(self.labeling?.mainValue ?? []);
     },
 
+    // used only in labels on regions for Image and Video tags
     getLabelText(joinstr) {
+      const index = self.region_index;
       const label = self.labeling;
       const text = self.texting?.mainValue?.[0]?.replace(/\n\r|\n/, " ");
       const labelNames = label?.getSelectedString(joinstr);
       const labelText = [];
 
+      if (index) labelText.push(String(index));
       if (labelNames) labelText.push(labelNames);
       if (text) labelText.push(text);
       return labelText.join(": ");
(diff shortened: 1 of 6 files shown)


nick-skriabin (approved):

web/libs/editor/src/mixins/AreaMixin.js
```diff
@@ -155,9 +158,17 @@ export const AreaMixinBase = types
     },
   }))
   .volatile(() => ({
-    // selected: false,
+    // index of the region in the regions tree (Outliner); will be updated on any order change
+    region_index: null,
   }))
   .actions((self) => ({
+    setCurrentIndex(index) {
```
Gondragos: Could it be better to name it `setRegionIndex`? We have `setItemIndex` in `RegionsMixin` to indicate the object it relates to. So it can be confusing.

Gondragos (approved):

web/libs/editor/src/mixins/AreaMixin.js
hlomzik: the idea was to indicate that this region is dynamic. On the other hand if we have a setter then it's probably dynamic by nature, huh. I'll change it, thank you

MihajloHoma (approved):

PR #6111, opened 2024-07-19
fix: OPTIC-683: Detect and convert large integers automatically in API responses before parsing as JSON to retain precision

bmartel:
- [X] Docs have been added/updated (for bug fixes/features)

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

When data within JSONB fields of the database would store large integers from other processes, when it came time to displaying these visually it would be done with loss of precision. The JS support for large integers is not very good being it only actually stores floats and has only 32bit support. Therefore to fix the display issues of these, I found that we were already performing a 2 stage response with all api calls from datamanager, therefore I added a method to replace large integers that are present in the textual data as the reviver function to JSON.parse running. This solves the issue and ultimately ensures we do not modify unknowingly data stored on the server while still retaining the visual accuracy for consumption in the UI.

_Before_
image

_After_
image

- Support BigInt in JS.
    - Problem here is it doesn't actually solve the original problem.
    - Can be explicitly supported in addition to this solution but this requires that all other Number values be converted to BigInt to be able to perform mathematical operations on them.
    - Many caveats to cover which ultimately the solution we need is to be as low risk to data corruption as possible.

Don't have one yet, will entertain one if we feel it necessary.

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [X] unit

DataManager API proxy.

### DIFF
--- a/web/libs/datamanager/src/utils/__tests__/helpers.test.ts
+++ b/web/libs/datamanager/src/utils/__tests__/helpers.test.ts
@@ -4,7 +4,7 @@ describe("helpers", () => {
   const originalMaxSafeInteger = Number.MAX_SAFE_INTEGER;
 
   beforeEach(() => {
-    // Mock Number.MAX_SAFE_INTEGER to a smaller value for testing purposes
+    // Mock Number.MAX_SAFE_INTEGER to a smaller value for testing purposes as NodeJS supports larger integers than Web.
     Object.defineProperty(Number, "MAX_SAFE_INTEGER", {
       value: 9007199254740991,
     });
--- a/web/libs/datamanager/src/utils/__tests__/helpers.test.ts
+++ b/web/libs/datamanager/src/utils/__tests__/helpers.test.ts
@@ -1,0 +1,60 @@
+import * as helpers from "../helpers";
+
+describe("helpers", () => {
+  const originalMaxSafeInteger = Number.MAX_SAFE_INTEGER;
+
+  beforeEach(() => {
+    // Mock Number.MAX_SAFE_INTEGER to a smaller value for testing purposes
+    Object.defineProperty(Number, "MAX_SAFE_INTEGER", {
+      value: 9007199254740991,
+    });
+  });
+
+  afterEach(() => {
+    Object.defineProperty(Number, "MAX_SAFE_INTEGER", {
+      value: originalMaxSafeInteger,
+    });
+  });
+
+  describe("jsonReviverWithBigInt", () => {
+    it("should return the source value if the value is a number and the source is defined and exceeds MAX_SAFE_INTEGER", () => {
+      const key = "key";
+      const value = 9007199254740992;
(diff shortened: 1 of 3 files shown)


yyassi-heartex (approved):

web/libs/datamanager/src/utils/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,60 @@
+import * as helpers from "../helpers";
+
+describe("helpers", () => {
+  const originalMaxSafeInteger = Number.MAX_SAFE_INTEGER;
+
+  beforeEach(() => {
+    // Mock Number.MAX_SAFE_INTEGER to a smaller value for testing purposes
```
bmartel:
```suggestion
    // Mock Number.MAX_SAFE_INTEGER to a smaller value for testing purposes as NodeJS supports larger integers than Web.
```

yyassi-heartex:
/selenium test
> See '/help'

AndrejOros (approved):

web/libs/datamanager/src/utils/helpers.ts
```diff
@@ -12,6 +12,35 @@ export const formDataToJPO = (formData: FormData) => {
   return formData;
 };
 
+/**
+ * Hydrate JSON values that are large integers to strings.
+ */
+export const jsonReviverWithBigInt = (_key: any, value: any, context: any) => {
```
hlomzik: easy fix for types problem:
```suggestion
export const jsonReviverWithBigInt = (_key: any, value: any, context?: any) => {
```

not all the browsers support context and while your code will work everywhere the typing for `parse()` has no `context` yet, so marking it as an optional will calm TS down.

web/libs/datamanager/src/utils/helpers.ts
```diff
@@ -12,6 +12,35 @@ export const formDataToJPO = (formData: FormData) => {
...
+ * This is a workaround for the fact that JSON.parse does not directly support big integers and will
+ * immediately convert them to numbers (losing precision).
+ *
+ * ex. { "id": 12345678901234567890 } => { "id": "12345678901234567890" }
+ *     { "id": -12345678901234567890 } => { "id": "-12345678901234567890" }
+ *     { "meta": { "id": 12345678901234567890 } } => { "meta": { "id": "12345678901234567890" } }
+ *     { "meta": { "id": -12345678901234567890 } } => { "meta": { "id": "-12345678901234567890" } }
+ **/
+export const parseJson = <T = any>(jsonString: string): T => {
+  // @ts-ignore The types were complaining about the signature which I believe is due to the NodeJS types of JSON.parse getting in the way
```
hlomzik: and then we don't need this
```suggestion
```

hlomzik (approved):

PR #6135, opened 2024-07-25
fix: LEAP-762: Fix offsets in domManager for emojis

Gondragos:
There were some problems with calculating offsets and applying text regions to the text/html containing emojis after domManager was introduced.
- here it might be fixed
- there are new tests for that
- RichText helper for testing is introduces

- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

N/A

- [ ] No

I hope nobody would like to keep that bug.

- [ ] e2e
- [x] integration
- [ ] unit

`Text`, `RichText`, `HyperText`, `domManager`

### DIFF
--- a/web/libs/editor/src/tags/object/RichText/domManager.ts
+++ b/web/libs/editor/src/tags/object/RichText/domManager.ts
@@ -8,6 +8,17 @@ const CF = "\r";
 type DDExtraText = string;
 
 /**
+ * Array of all characters and dummy placeholders
+ *
+ * Content is a way to store information about the displayed text
+ * and be able to restore global offsets and relative offsets in the same time.
+ * All hidden characters as "/n" or spaces at the start/end are stored as "" (dummy)
+ * but we keep in mind that it is a character with `length` == 1,
+ * and it affects both global and relative offsets
+ */
+type Content = string[];
+
+/**
  * Normalize text for displaying it.
  * It replaces all line breaks with '\n' symbol.
  * This is a variant used historically, but it converts \r\n to \n\n which might be not correct
@@ -21,11 +32,10 @@ class DDTextElement {
   public node: Text;
   public start: number;
   public end: number;
-  // array of all characters and dummy placeholders
-  public content: string[];
+  public content: Content;
   public path?: string;
 
-  constructor(node: Text, start: number, end: number, content: string[], path?: string) {
+  constructor(node: Text, start: number, end: number, content: Content, path?: string) {
     this.node = node;
     this.start = start;
     this.end = end;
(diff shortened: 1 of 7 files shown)


web/libs/editor/src/tags/object/RichText/domManager.ts
hlomzik, line 1: Overall suggestion is to add documentation for the whole file — what are the main ideas, what is the structure and main functions here. Like "domManager keeps the internal structure of all elements inside it with offset mapping to find element by offset in O(1) and get offset of the element in zero time. Provides mappings via methods: offsetToRange, <...>, etc".

hlomzik: Also I suggest to separate string offsets and code points to understand the difference and to draw a line between methods operating one or another and to highlight the conversion points. This can be done via virtual alias types for example: `type CodePoint = number; type StringOffset = number` at least for readability, but also maybe it can really be used to typesafe the code.

hlomzik (approved):

web/libs/editor/src/tags/object/RichText/domManager.ts
```diff
@@ -720,14 +788,21 @@ export default class DomManager {
...
+    let endEl = this.domData.findElementByPath(end);
 
     if (!startEl || !endEl) {
       return undefined;
     }
+    if (!(startEl instanceof DDDynamicBlock)) {
+      startEl = this.domData.findTextBlock(startEl.start, "end") as DDDynamicBlock;
+    }
+    if (!(endEl instanceof DDDynamicBlock)) {
+      // It really should be "end" and not "start" as we are looking for the exact container by the start position
```
hlomzik, line 802: this and next comments makes it even more confusing. what `startEl` and `endEl` are and why do we use `"end"` for start? `"end"` for `endEl` seems more obvious. Also why do you convert text block to dynamic block? Is it possible to define interfaces and use mostly them? like if any element in `DomData.elements` could have the same interface that would be awesome.

bmartel (approved):

MihajloHoma (approved):

web/libs/editor/tests/integration/e2e/ner/emoji.cy.ts
```diff
@@ -0,0 +1,338 @@
...
+        expect(resultValue.globalOffsets.start).to.eq(refHyperTextMultilineResultValue.globalOffsets.start);
+        expect(resultValue.globalOffsets.end).to.eq(refHyperTextMultilineResultValue.globalOffsets.end);
+        expect(resultValue.startOffset).to.eq(refHyperTextMultilineResultValue.startOffset);
+        expect(resultValue.endOffset).to.eq(refHyperTextMultilineResultValue.endOffset);
+        expect(resultValue.text).to.eq(refHyperTextMultilineResultValue.text);
+      });
+    });
+  });
+
+  it.only('Heuristic edge case', () => {
```
Gondragos:
```suggestion
  it('Heuristic edge case', () => {
```

PR #6139, opened 2024-07-26
feat: LEAP-1266: Video frame classification

hlomzik:
New type of regions — quazi-classification per-frame choices.
VideoClassification region type is introduced with new property `perframe` for regions and for tags.
These regions are displayed as normal video region with some style adjustments, but they don't appear in regions tree.
Multiple classification methods and getters were adjusted to properly react on user aactions.

New structure is a mix of choices and video rectangles regions:
Every Choice in Choices tag creates separate region+result with `choices` field as an array of one element, this choice value. Plus this result will also have `sequence` similar to video rectangle, but just with `time`, `frame` and `enabled`.

- [x] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

It's highly requested feature.

This feature is just an add-on, so it should not change any existing functionality. No FFs introduced.

- [x] No

- [ ] e2e
- [ ] integration
- [ ] unit


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10171905260)
> Successfully merged:  create mode 100644 docs/themes/v2/source/images/releases/2-14-UI-settings2.png

web/libs/editor/src/components/Node/Node.tsx
```diff
@@ -80,6 +81,12 @@ const NodeViews = {
     getContent: (node) => <span style={{ color: "#5a5a5a" }}>from {node.sequence[0]?.frame} frame</span>,
   }),
 
+  VideoClassificationModel: NodeView({
+    name: "Video",
+    icon: (...args: any) => <LsChevron {...args} style={{ rotate: "180deg" }} />,
```
yyassi-heartex, line 86: I like this idea of using the icon and rotating to direction we want but I want to call out that between LSO and LSE we have different ways to accomplish the same thing and even within LSO we have this method and `FaCheveronUp`, down, left and right.
maybe this should be a thing we look at more closely as part of an FE guild initiative as its definitely out of scope for this ticket

bmartel: My two cents on this one is slightly tangential but focuses on this area, we shouldn't be using any other icon sets outside of the ones curated by the design team. So the usage of different icons from different sets (this leads to importing and not tree shaking not just one full icon set, but actually 2 which are large). I think we raise Icons as its own focal point, to make sure we are all on the same page I think there is low hanging fruit to save massively on both our bundle size and design consistency.

web/libs/editor/src/tags/object/Video/Video.js
```diff
@@ -197,13 +207,87 @@ const Model = types
         }
       },
 
-      addRegion(data) {
-        const control = self.videoControl() ?? self.control();
+      /**
+       * Update per frame classifications view; analogue for `updateFromResult()` but with data for current frame
+       */
+      updatePerFrameViews() {
```
yyassi-heartex, line 213: there is alot going on here - would be great to have a test of some sort to check this going forward

yyassi-heartex (approved):
looks good - but adding some tests would go a long way here

hlomzik:
Outdated approach cancelled in favor of TimelineLabels: https://github.com/HumanSignal/label-studio/pull/6191

PR #6152, opened 2024-07-30
fix: OPTIC-945: Improve the bundling strategy and ship a single entrypoint with a common vendor chunk across all libs

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

This change aligns the cross library component usage with React, such that it can and is now bundling only 1 set of dependencies across all of the code shipped as part of the frontend. The primary driving force of this change is to enable better code bundling strategies, reduce the amount of code shipped, fix our React usage which had become tied to workarounds from the triplicated definitions of React in various bundles and align the codebase to being able to start extracting common components and utils to a shared space. This PR now allows us to adjust, measure and update the bundling strategy further than proposed here, and will also allow a component by component reduction in the codebase. The end goal being, to ship less code to the end user more granularly with improved performance and consistency.

Functionally and stylistically this should remain exactly as previous. The only difference is in how we are bundling the code, and with which parts are put in specific bundles to retain a better longer term caching strategy between releases.

To achieve this single bundle we had to opt to de-duplicate the components which conflicted in css definitions, which ultimately based on the findings of the current state was more tenable to hardcode the prefix of a given component in conflict as a postfix of the block. This is backwards compatible and also addresses the issue in unifying our build.

#### Notes to reviewers
Explicitly this change has a direct dependency on the decommissioning and removal of the FF `fflag_fix_front_lsdv_4620_memory_leaks_100723_short`. The reason is, we are now only using a single React and this change severs React fibers which are still in use by the remainder of page and does not allow React to properly flush pendingContext as a result, which causes errors and functionality to break. As part of this change we are going to be rolling this out with the FF turned off, so that it operates correctly for everyone. I tested this branch, and there are no memory leaks with this FF turned off, which aligns with the thinking that the original need to have an experimental cleanup was due to our React usage which caused things to linger in memory without ability for the browser to ever garbage collect it.

None. Webpack config was the largest difference here, same version is used for all libs and plugins at this time.

It can increase performance due to the removal of many duplicated portions of code, but ultimately the performance gain initially here is in the more deterministic cache strategy around the code bundles shipped being more resistant to unnecessary change. This means less code being reloaded between versions shipped, and stronger cache lifetimes between releases which improves load times in any given page by shipping less code over time for all users.

No.

The one other alternative to the particular solution here was to fix the CSS prefixing collisions with a modified webpack config for the css_prefix value injected in css/scss/styl files during processing. This was attempted and blocked by the fact that in a given build, if the file which would be produced is a culmination of multiple other css from multiple potential sources (LS App, Editor, DM) it only could be made to apply a single css prefix holistically to the file output, and didn't allow for proper distinguishing between which files had sourced the particular css chunk. This ultimately meant that it was non trivial to achieve this level of change with Webpack, and decided to not pursue any further.

None, it wasn't possible to entertain many builds to possibly FF copies all over the place of similar files to retain the previous and the new. This would only further complicate the situation and make it even harder to move forward.

_(check only one)_
- [X] No

All.


label_studio/feature_flags.json
```diff
@@ -4432,7 +4432,7 @@
     },
     "fflag_fix_front_lsdv_4620_memory_leaks_100723_short": {
       "key": "fflag_fix_front_lsdv_4620_memory_leaks_100723_short",
-      "on": true,
+      "on": false,
```
yyassi-heartex, line 4138: we need to remember to turn this off in LF on delivery

web/apps/labelstudio/src/pages/Settings/DangerZone.jsx
```diff
@@ -106,7 +106,7 @@ export const DangerZone = () => {
 
             return (
               btn.disabled !== true && (
-                <div className={cn("settings-wrapper")}>
+                <div className={cn("settings-wrapper")} key={btn.type}>
```
yyassi-heartex, line 109: this addressed a React warning about having child elements with no distinct keys

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -59,22 +60,22 @@ const buildLink = (path, params) => {
...
   const [crashed, setCrashed] = useState(false);
+  const [loading, setLoading] = useState(!window.DataManager || !window.LabelStudio);
   const dataManagerRef = useRef();
   const projectId = project?.id;
 
   const init = useCallback(async () => {
-    if (!LabelStudio) return;
-    if (!DataManager) return;
+    if (!window.LabelStudio) return;
+    if (!window.DataManager) return;
```
yyassi-heartex, line 78: ideally we should sub these checks out for a check against `dependencies` this will help us move towards removing dependency on these window declarations of LabelStudio and DataManager

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -59,22 +60,22 @@ const buildLink = (path, params) => {
...
   const root = useRef();
   const params = useParams();
   const history = useHistory();
   const api = useAPI();
   const { project } = useProject();
-  const LabelStudio = useLibrary("lsf");
-  const DataManager = useLibrary("dm");
   const setContextProps = useContextProps();
   const [crashed, setCrashed] = useState(false);
+  const [loading, setLoading] = useState(!window.DataManager || !window.LabelStudio);
```
yyassi-heartex, line 72: same as line 77 + 78

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
yyassi-heartex: alternatively we hold off and dig deeper into it on a follow up ticket

label_studio/feature_flags.json
Gondragos: Did you find another way to solve memory leaks? Or what is the plan here?

bmartel: There is no memory leaks when we are not building 3 separate React scripts + mobx-state-tree. When bundled together, and served properly this yielded what is expected, that React lifecycles are respected and ultimately MST was not retaining memory. In fact, it now throws errors if running this code because it fundamentally disrupts normal React usage by ripping out the fibers.

Gondragos: As I know part of the problem is in React 17 itself.  https://github.com/HumanSignal/label-studio-frontend/pull/1297/files#diff-0a9a22cedc083cfd1838c700569c35f3163ab833e220a32fa42a78432d4427b4R87-R95
If we do not move to React 18 there is no reason to say that there are no memory leaks.

bmartel: I had mentioned 2 things which of those 2 I still very much believe our issues have always been the usage of React. Most of my comment there and reading and observations is that if we are compiling and shipping and interoperating with 3 separate React code points, plus 2 different instances of Mobx+MobxStateTree, this would be a completely obvious reason as to how React would become unable to properly cleanup itself, because in almost all handoff points between the previously separate React versions we would be doing something very fundamental to the problem:

Closing quickview and decommissioning editor code and transitioning to DataManager. Or closing Preview and transitioning to any other screen. Anytime this would happen, we were pruning React ourselves, because we had to, React entirely was going away at the same time as the code we were having stuck in memory, so React couldn't ever run its flush cycle and MobxStateTree would be attached forever to a portion of memory that would never release due to the Map with string key usage within a React context. Now, I'm not seeing that retention, because I firmly believe that this issue was inside of that handoff which we had 2 moving parts previously, now only one, the decommissioning of the node out of React.

If you have any methods of asserting the presence or absence of the memory leak scenario from when you came up with this solution, I would love to run it and investigate further. As of now, I have not been able to reproduce this on this branch.

label_studio/templates/base.html
```diff
@@ -84,6 +77,9 @@
   {% block context_menu_right %}{% endblock %}
 </template>
 
+{% block global-js %}
+{% endblock %}
+
```
bmartel:
```suggestion
```

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
bmartel: This makes sense, we came up with the alternative approach later and this is an adage of a previous incarnation. We should address it here, and use this as the canonical approach for LSE as well.

yyassi-heartex:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10200524226)
> Successfully merged:  44 files changed, 1826 insertions(+), 372 deletions(-)

yyassi-heartex (approved):
Approving to get QA eyes on this

yyassi-heartex:
/docker build
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10289004363)
> Docker image was pushed with the tag `20240807.173543-fb-optic-945-8f0feec8d`

yyassi-heartex:
/docker build
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10293228982)
> Docker image was pushed with the tag `20240807.233130-fb-optic-945-ae4739667`

yyassi-heartex:
/docker build
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10301867574)
> Docker image was pushed with the tag `20240808.121759-fb-optic-945-36ea7280c`

yyassi-heartex:
/docker build
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10302370312)
> Docker image was pushed with the tag `20240808.124951-fb-optic-945-d428c41db`

yyassi-heartex:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10307949143)
> Successfully merged:  5 files changed, 8 insertions(+), 114 deletions(-)

label_studio/feature_flags.json
hlomzik, line 1: This file is autogenerated and it's better to fix the state of the FF via stale flags

label_studio/templates/base.html
```diff
@@ -119,21 +112,38 @@
...
+  If for any reason we need to invalidate the cache, we can do so by changing the version number.
+  The libs which are contained here which if updated require this version to also be updated are:
+  - react
+  - react-dom
+  - react-router
+  - react-router-dom
+  - mobx
+  - mobx-react
+  - mobx-react-lite
+  - mobx-state-tree
```
yyassi-heartex: to avoid future confusion - removing the specific list

yyassi-heartex:
```suggestion
```

label_studio/feature_flags.json
hlomzik: or just LD state for next release

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
```diff
@@ -10,9 +9,11 @@ import { useAPI } from "../../../providers/ApiProvider";
 
 const configClass = cn("configure");
 
+const loadDependencies = async () => import("@humansignal/editor");
+
 export const Preview = ({ config, data, error, loading, project }) => {
-  const LabelStudio = useLibrary("lsf");
   const lsf = useRef(null);
+  const [resolvingEditor] = useState(loadDependencies);
```
hlomzik: it's not used as a proper state, better to use `useMemo()` maybe and it will be called only once?

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -59,22 +60,22 @@ const buildLink = (path, params) => {
 };
 
 export const DataManagerPage = ({ ...props }) => {
+  const [dependencies] = useState(loadDependencies);
```
hlomzik: and the same here

farioas (approved):

label_studio/feature_flags.json
yyassi-heartex: so I added the hard override + i updated LD for test environment, will update community merge

yyassi-heartex:
/docker build
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10323550154)
> Docker image was pushed with the tag `20240809.175843-fb-optic-945-db8ac02a0`

nick-skriabin (approved):

vladimirheartex (approved):

PR #6189, opened 2024-08-12
feat: DIA-1122: Add ability to store FailedPredictions

hakan458:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/tasks/models.py
```diff
@@ -992,6 +992,44 @@ class Meta:
...
+        help_text='The message explaining why generating this prediction failed'
+    )
+    error_type = models.CharField(
+        _('error_type'),
+        max_length=512,
+        default=None,
+        null=True,
+        help_text='The type of error that caused prediction to fail'
+    )
+    model_version = models.TextField(
```
niklub, line 1030: isn't `model_run` already containing reference `model_version` ?

label_studio/tasks/models.py
```diff
@@ -992,6 +992,44 @@ class Meta:
...
+        _('model version'),
+        default='',
+        blank=True,
+        null=True,
+        help_text='A string value that for model version that produced the failed prediction. Used in both live models and when uploading offline predictions.',
+    )
+    model_run = models.ForeignKey(
```
niklub, line 1037: if we're going to use this in MLbackend, shall we avoid assigning any value here?

niklub: and in this case, how do we reference ML backend ?

niklub (changes requested):

label_studio/tasks/migrations/0048_failedprediction.py
```diff
@@ -0,0 +1,30 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='FailedPrediction',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('message', models.TextField(blank=True, default='', help_text='The message explaining why generating this prediction failed', null=True, verbose_name='message')),
+                ('error_type', models.CharField(default=None, help_text='The type of error that caused prediction to fail', max_length=512, null=True, verbose_name='error_type')),
+                ('model_version', models.TextField(blank=True, default='', help_text='A string value that for model version that produced the failed prediction. Used in both live models and when uploading offline predictions.', null=True, verbose_name='model version')),
+                ('created_at', models.DateTimeField(auto_now_add=True, verbose_name='created at')),
+                ('updated_at', models.DateTimeField(auto_now=True, verbose_name='updated at')),
```
niklub: do we need updated_at if they are immutable?

niklub (changes requested):

label_studio/tasks/models.py
hakan458: Discussed on Slack - will add a reference to ML backend model to FailedPrediction 
On delete will set to null instead of cascade 

Model run will not be set in case that we are using ML backend yes

label_studio/tasks/models.py
hakan458: Yes that is true, however this field will be used in case we are using a ML backend as well. Reference to the model itself + model version (just like we do for predictions)

label_studio/tasks/migrations/0048_failedprediction.py
hakan458: Yeah I was thinking about this as well, I think for now you are correct. I will remove

matt-bernstein (approved):

label_studio/tasks/migrations/0048_failedprediction.py
```diff
@@ -0,0 +1,31 @@
...
+        ('ml', '0007_auto_20240314_1957'),
+        ('tasks', '0047_merge_20240318_2210'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='FailedPrediction',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('message', models.TextField(blank=True, default='', help_text='The message explaining why generating this prediction failed', null=True, verbose_name='message')),
```
pakelley: My understanding is that it's preferred in django to use null=False, any reason we want to use null here (and for error_type)?

label_studio/tasks/migrations/0048_failedprediction.py
```diff
@@ -0,0 +1,31 @@
...
+                ('error_type', models.CharField(default=None, help_text='The type of error that caused prediction to fail', max_length=512, null=True, verbose_name='error_type')),
+                ('model_version', models.TextField(blank=True, default='', help_text='A string value that for model version that produced the failed prediction. Used in both live models and when uploading offline predictions.', null=True, verbose_name='model version')),
```
pakelley: blank=True to account for non-Prompts use cases, I take it? I figure we should do the same for model_run

label_studio/tasks/models.py
```diff
@@ -847,7 +847,7 @@ def delete(self, *args, **kwargs):
 
 
 class Prediction(models.Model):
-    """ML backend predictions"""
+    """ML backend / prompts predictions"""
```
pakelley: nit: can you capitalize to make it clear we mean the "Prompts" project, not the general idea of prompts (I know the context is pretty clear, just want it to be stupidly obvious)

label_studio/tasks/models.py
```diff
@@ -992,6 +992,50 @@ class Meta:
...
 
 
+class FailedPrediction(models.Model):
+    """
+    Class for storing failed prediction(s) for a task
+    """
+    message = models.TextField(
+        _('message'),
+        default='',
+        blank=True,
```
pakelley, line 1012: I figure we shouldn't allow this to be blank, right? Is there a case it could be?

label_studio/tasks/models.py
```diff
@@ -992,6 +992,50 @@ class Meta:
...
+        help_text='The message explaining why generating this prediction failed'
+    )
+    error_type = models.CharField(
+        _('error_type'),
+        max_length=512,
+        default=None,
+        null=True,
+        help_text='The type of error that caused prediction to fail'
+    )
+    ml_backend_model = models.ForeignKey(
```
pakelley, line 1023: Ah... is this what we do for regular Predictions? Unfortunate that we have different attributes for Prompts vs ML Backend, we'll have to fix that eventually.
I figure blank=True here though, since it would be blank for Prompts-based failed preds?

label_studio/tasks/models.py
```diff
@@ -992,6 +992,50 @@ class Meta:
...
+        help_text='A string value that for model version that produced the failed prediction. Used in both live models and when uploading offline predictions.',
+    )
+    model_run = models.ForeignKey(
+        'ml_models.ModelRun',
+        on_delete=models.CASCADE,
+        related_name='failed_predictions',
+        null=True,
+        help_text='A run of a ModelVersion that created the failed prediction.',
+    )
+    project = models.ForeignKey('projects.Project', on_delete=models.CASCADE, related_name='failed_predictions', null=True)
```
pakelley: naive question: do we need the project if we have the task? Can't we get the project from the task?

label_studio/tasks/migrations/0048_failedprediction.py
matt-bernstein: https://www.notion.so/humansignal/Django-Database-Migrations-Best-Practices-42109ea03c9543dc84b0bf43e5150ce5?pvs=4#ed560b0240b04a44aa12704e7fa30356

hakan458: Everywhere we have a text field we are using null=True

label_studio/tasks/migrations/0048_failedprediction.py
hakan458: Also copying what were doing everywhere else, all TextFields have this option. We can enforce it being required in the serializer if needed, not at DB level

label_studio/tasks/models.py
hakan458: Will do

label_studio/tasks/models.py
hakan458: Yes on Adala side if the Exception for some reason does not have this we will be sending an empty string. Unlikely, but could happen.

label_studio/tasks/models.py
hakan458: blank is not an option for FKs, null is

label_studio/tasks/models.py
hakan458: We could, but doesnt hurt to store, can get all failed predictions for a project without first getting a task and doing some weird queries. Tried to keep it similar to normal Prediction model so we can easily get both if/when needed

pakelley (approved):

pakelley (approved):

niklub (approved):

forum-hs (approved):

label_studio/tasks/models.py
```diff
@@ -1001,6 +1001,50 @@ class Meta:
         db_table = 'prediction'
 
 
+class FailedPrediction(models.Model):
+    """
+    Class for storing failed prediction(s) for a task
+    """
+    message = models.TextField(
+        _('message'),
+        default='',
```
jombooth: When adding columns to existing models, it's not safe to use non-null defaults at all per the migration guide; in this case I don't think we have a problem because this is a brand new model, but I would prefer we didn't mix null and empty string defaults for what are essentially string fields

jombooth:
```suggestion
        default=None,
```

label_studio/tasks/models.py
```diff
@@ -1001,6 +1001,50 @@ class Meta:
...
+    ml_backend_model = models.ForeignKey(
+        'ml.MLBackend',
+        on_delete=models.SET_NULL,
+        related_name='failed_predictions',
+        null=True,
+        help_text='An ML Backend instance that created the failed prediction.',
+    )
+    model_version = models.TextField(
+        _('model version'),
+        default='',
```
jombooth: same as https://github.com/HumanSignal/label-studio/pull/6189/files#r1717446565
```suggestion
        default=None,
```

PR #6201, opened 2024-08-13
feat: DIA-1321: [BE] Create Prompt with the NER problem type

matt-bernstein:
First step to support multiple skills in Prompts backend.

- [] Commit message(s) and PR title follows the format `[fix|feat|ci|chore|doc]: TICKET-ID: Short description of change made` ex. `fix: DEV-XXXX: Removed inconsistent code usage causing intermittent errors`

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/ml_models/migrations/0010_modelinterface_skill.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ml_models', '0009_alter_thirdpartymodelversion_provider'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='modelinterface',
+            name='skill',
```
pakelley: 🎊 🥳 🪩

label_studio/ml_models/models.py
```diff
@@ -13,6 +13,12 @@
 logger = logging.getLogger(__name__)
 
 
+# skills are partitions of projects (label config + input columns + output columns) into categories of labeling tasks
+class Skill(models.TextChoices):
```
pakelley: ok, maybe getting ahead of myself, but the more I think about how we'll probably want to handle Skills more broadly, the more I think we'll probably want a separate `Skill` class, proper, and this is more of a `SkillChoice` or `SkillType` or something...
For example, If we do it that way then we can have an object that encodes all of the input classes of a Skill, and can be generated from this with something like `Skill.from_choice(SkillChoice.TEXT_CLASSIFICATION`. I don't love "SkillChoice", so definitely up for a different name, but my feeling is that we'll want to reserve the actual `Skill` class for the actual business logic.
Wdyt?

matt-bernstein: That's reasonable, `SkillName` is a good fit here.

The API sketch I have in mind is something like
```
skill_types = Skill.compatible_with(project.label_config)
skill_name = skill_types[0]
skill = Skill.from_label_config(skill_name, project.label_config)
# or, depending if we use subclassing/mixins/etc
skill_type = skill_types[0]
skill_name = skill_type.name
skill = skill_type.from_label_config(project.label_config)
```

label_studio/ml_models/migrations/0010_modelinterface_skill_name.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.25 on 2024-08-13 22:56
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
niklub, line 6: Can we avoid migrations if possible?

niklub (changes requested):

label_studio/ml_models/migrations/0010_modelinterface_skill_name.py
matt-bernstein: This is needed, we kept the new fields added to a minimum but we do have to have a way to distinguish between types of projects at least. @pakelley can chime in w other design considerations but we discussed this as a team

label_studio/ml_models/models.py
```diff
@@ -39,6 +45,8 @@ class ModelInterface(models.Model):
         'organizations.Organization', on_delete=models.CASCADE, related_name='model_interfaces', null=True
     )
 
+    skill_name = models.CharField(max_length=255, choices=SkillNames.choices, default=SkillNames.TEXT_CLASSIFICATION)
```
jombooth: `default=` is unsafe if we need to roll back; until we get on django 5 and can use `db_default`, every new field needs to be nullable and have no default set.

matt-bernstein: changed

matt-bernstein: (default value is set in the serializers in LSE now, would like to remove it in future but that'd be too big of a change surface area for this PR, especially FE)

jombooth (approved):
LGTM, nice work!

niklub (approved):
lgtm ensuring tests are green and we can verify migration locally

label_studio/ml_models/migrations/0010_modelinterface_skill_name.py
pakelley: yeah, this is a necessary addition at this point. We basically always knew we'd need a field like this, but avioded adding it before in case it played out some other way.
But we need some way to distinguish a NER ModelInterface from a Text Classification one without parsing the label config each time

pakelley (approved):

hakan458 (approved):

PR #6203, opened 2024-08-14
feat: LEAP-1351: LSF customization for bulk annotation 

Gondragos:
This PR adds
- `"annotation:bulk"` interface that allows to hide all panels and elements that can't be set equally to group of annotaions
- small fixes for custom controll buttons in bottom bar to be able to use them
- fix the way we deliver config in cypress tests

- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

`fflag_feat_all_leap_1181_bulk_annotation_short`

- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Editor`, `BottomBar`


web/libs/editor/src/components/App/App.jsx
```diff
@@ -260,16 +268,23 @@ class App extends Component {
...
-                  panelsHidden={viewingAll}
-                  currentEntity={as.selectedHistory ?? as.selected}
-                  regions={as.selected.regionStore}
-                  showComments={store.hasInterface("annotations:comments")}
-                  focusTab={store.commentStore.tooltipMessage ? "comments" : null}
-                >
-                  {mainContent}
-                  {store.hasInterface("topbar") && <BottomBar store={store} />}
-                </SideTabsPanels>
+                shouldShowPanels ? (
```
yyassi-heartex: should this be applied for situations where `newUIEnabled` is false? IE are we supporting this for situations where FF_DEV_3873 is turned off?

web/libs/editor/src/tags/visual/Collapse.jsx
```diff
@@ -23,76 +24,107 @@ const { Panel } = Collapse;
...
 
 const HtxCollapse = observer(({ item }) => {
   return (
     <Collapse bordered={item.bordered} accordion={item.accordion}>
       {item.children
-        .filter((i) => i.type === "panel")
+        .filter(
+          (i) =>
+            i.type === "panel" &&
+            ((!isFF(FF_BULK_ANNOTATION) && !item.store.hasInterface("bulk:annotation")) || i.isIndependent),
```
yyassi-heartex: can we move `(!isFF(FF_BULK_ANNOTATION) && !item.store.hasInterface("bulk:annotation")` to a variable outside the filter that describes what this means? - this way we dont need to recalculate it for every `i` when 'i' doesn't affect it.
this will also help better define what this is making it more readable

web/libs/editor/src/tags/visual/Filter.jsx
```diff
@@ -59,6 +59,11 @@ const Model = types
     get toTag() {
       return self.annotation.names.get(self.toname);
     },
+
+    // Indicates that it could exist without information about objects, taskData and regions
+    get isIndependent() {
+      return self.toTag ? self.toTag.isIndependent : false;
```
yyassi-heartex: same as this no?
```suggestion
      return self.toTag && self.toTag.isIndependent;
```

web/libs/editor/src/tags/visual/Collapse.jsx
Gondragos: Now I think that the condition was also wrong.

web/libs/editor/src/tags/visual/Filter.jsx
hlomzik: @yyassi-heartex your version returns mixed values instead of plain boolean, so I'd prefer the original version

hlomzik:
Would not it be better to name interface "annotation:bulk"? That will keep it in a shape of namespace:detail like for lot of other interfaces. I know they are a mess but that's a really lowcost change on an early stage with small benefit

web/libs/editor/src/tags/visual/Filter.jsx
yyassi-heartex: ahh i see - good point

Gondragos: @hlomzik From the other side, for some tags it can be `undefined` anyhow.

yyassi-heartex: since its really just checking if toTag is defined maybe we can do something like - feels like thats the intent - is that spot on?
```suggestion
      return self.toTag?.isIndependent ?? false;
```

web/libs/editor/src/components/App/App.jsx
Gondragos: Thank you! I have checked we have this flag switched on not everywhere.
So I've changed the code and now it might be displayed right for both cases if `FF_DEV_1170` is on.

web/libs/editor/src/components/App/App.jsx
```diff
@@ -219,6 +226,7 @@ class App extends Component {
       </Block>
     );
 
+    const shouldShowPanels = isDefined(store) && (!isFF(FF_BULK_ANNOTATION) || !store.hasInterface("annotation:bulk"));
```
hlomzik: `store` is definitely defined at this point. and I'd suggest to rename it to `isBulkAnnotation` so later in the code it will be more clear when we display panels and when we don't

Gondragos: It kinda depends on what you want to make clear, but I'm ok with that change.

web/libs/editor/src/core/CustomTypes.ts
```diff
@@ -50,7 +50,30 @@ const CSSColor = types.custom<any, string>({
   },
 });
 
+// This type is using to store a raw callback function in mobx-state-tree model.
+// It won't be serialized right. However, it is useful to allow passing a callback function as a prop
+// and be able to get a reaction on its change
+//
+// /!\ Avoid using this type in case you need serialization of the model it contains
+const RawCallback = types.custom<Function, Function>({
```
hlomzik: I think we definitely need to add more explanations on how and why would we use this new type, including:
- what was the high-level reason of adding new uncommon type to MST
- how would we create modals with this types from outside, examples maybe?
- what other considerations were there and why this one is the best option

Why I think that these are important questions they can be answered later.

web/libs/editor/src/tags/visual/View.jsx
hlomzik, line 1: It's better to finally move this list of allowed children from this and other tags to a reasonable place (like in tags/children or some constants list), plus optionally add an ability to easily modify them if we don't accept one specific tag for example. But that's for later.

web/libs/editor/src/core/Tree.tsx
```diff
@@ -234,6 +234,10 @@ function renderItem(ref: IAnyStateTreeNode, annotation: IAnnotation, includeKey
   const typeName = type.name;
   const View = Registry.getViewByModel(typeName);
 
+  if (isFF(FF_BULK_ANNOTATION) && annotation?.store?.hasInterface("annotation:bulk") && el.isIndependent !== true) {
```
hlomzik: I'd split this too long line into `isBulk` const and short check

web/libs/editor/src/components/BottomBar/BottomBar.jsx
```diff
@@ -20,7 +20,11 @@ export const BottomBar = observer(({ store }) => {
       <Elem name="group">
         {store.hasInterface("controls") && (store.hasInterface("review") || !isPrediction) && (
           <Elem name="section" mod={{ flat: true }}>
-            <Controls annotation={entity} />
+            {isFF(FF_BULK_ANNOTATION) && store.hasInterface("controls:custom") ? (
```
hlomzik: Add this interface to the PR description

hlomzik (approved):

web/libs/editor/src/core/CustomTypes.ts
Gondragos: It was decided to take the solution of this issue beyond this PR

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10562522580)

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10617538191)
> Successfully merged:  387 files changed, 6483 insertions(+), 9039 deletions(-)

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10674028201)
> Successfully merged:  29 files changed, 94 insertions(+), 69 deletions(-)

robot-ci-heartex:

robot-ci-heartex:

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11975726292)
> Successfully merged:  29 files changed, 78 insertions(+), 117 deletions(-)

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/12011593003)
> Successfully merged:  12 files changed, 21 insertions(+), 40 deletions(-)

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/12033118881)
> Successfully merged:  delete mode 100644 Dockerfile.redhat

web/libs/editor/src/utils/feature-flags.ts
```diff
@@ -272,6 +272,10 @@ export const FF_REVIEWER_FLOW = "fflag_feat_all_leap_1081_reviewer_flow_updates"
 
 export const FF_CUSTOM_SCRIPT = "fflag_feat_all_leap_883_custom_script_270524_short";
 
+/**
+ * It adds functionality of bulk annotation
+ */
+export const FF_BULK_ANNOTATION = "fflag_feat_all_leap_1181_bulk_annotation_short";
```
hlomzik, line 278:
```suggestion
export const FF_BULK_ANNOTATION = "fflag_feat_all_leap_1181_bulk_annotation_short";

```

web/libs/editor/src/tags/visual/Collapse.jsx
```diff
@@ -22,76 +23,105 @@ const { Panel } = Collapse;
...
+  .views((self) => ({
+    // Indicates that it could exist without information about objects, taskData and regions
+    get isIndependent() {
+      // if value starts with $ it's related to the data but in case of Panel it's just affect title
+      // we may still want to show it even if there is no data
+      // if (self.value && self.value[0] === "$") return false;
 
-  _value: types.optional(types.string, ""),
-  value: types.optional(types.string, ""),
+      // In other cases Panel can exist independent if it hase some independent children
```
hlomzik:
```suggestion
      // In other cases Panel can exist independent if it has some independent children
```

hlomzik (approved):

yyassi-heartex (approved):
looks good, but as @nicholasrq pointed out unit tests need to be fixed - approving with the caveat that they be fixed

web/libs/editor/src/components/BottomBar/buttons.tsx
```diff
@@ -15,15 +15,15 @@ type MixedInParams = {
...
+export function controlsInjector<T extends {}>(fn: (props: T & MixedInParams) => ReactElement) {
   const wrapped = inject(({ store }) => {
     return {
       store,
       history: store?.annotationStore?.selected?.history,
     };
   })(fn);
   // inject type doesn't handle the injected props, so we have to force cast it
-  return wrapped as unknown as (props: T) => ReactNode;
+  return wrapped as unknown as (props: T) => ReactElement;
```
nick-skriabin, line 26: you don't need to specify return type in the function signature if you explicitly do that in the return statement

web/libs/editor/src/components/App/App.jsx
```diff
@@ -278,24 +280,33 @@ class App extends Component {
...
+                  >
+                    {mainContent}
+                    {store.hasInterface("topbar") && <BottomBar store={store} />}
+                  </SideTabsPanels>
+                ) : (
+                  <>
+                    {mainContent}
+                    {store.hasInterface("topbar") && <BottomBar store={store} />}
+                  </>
+                )
```
nick-skriabin: i would reverse this ternary so the bulk UI goes first. `!isBulk` type of notation are harder to read in such case

nick-skriabin (approved):

web/libs/editor/src/components/App/App.jsx
yyassi-heartex: thank you for pointing this out - im considering adding this lint rule https://biomejs.dev/linter/rules/no-negation-else/#_top to help catch this with the linter

web/libs/editor/src/components/TopBar/TopBar.jsx
```diff
@@ -77,8 +80,8 @@ export const TopBar = observer(({ store }) => {
       ) : (
         <>
           <Elem name="group">
-            <CurrentTask store={store} />
-            {!isViewAll && (
+            {!isBulkMode && <CurrentTask store={store} />}
+            {!isViewAll && !isBulkMode && (
```
bmartel, line 84: Is this even possible to reach here with `isBulkMode` true given the check at the top of the component returns null in that case? And if it cannot be true, it is presumed only false here.

web/libs/editor/src/tags/visual/Collapse.jsx
```diff
@@ -22,76 +23,105 @@ const { Panel } = Collapse;
...
+      "paragraphs",
+      "paragraphlabels",
+    ]),
+  })
+  .views((self) => ({
+    // Indicates that it could exist without information about objects, taskData and regions
+    get isIndependent() {
+      // if value starts with $ it's related to the data but in case of Panel it's just affect title
+      // we may still want to show it even if there is no data
+      // if (self.value && self.value[0] === "$") return false;
```
bmartel, line 79:
```
      // if (self.value && self.value[0] === "$") return false;
```

Is this meant to be commented out? I would remove this if unused.

bmartel (approved):
Couple tiny little things, really well done!

web/libs/editor/src/tags/visual/Collapse.jsx
Gondragos: Yeah... At first I thought that it was a good idea to hide this element of it depends on a data field. But then I figured out that it could be better to display it anyway. This line is  "ready to be uncomented" in case if my last assumption is not right. But I would like to keep it just as an option for a while.

Maybe there is a better way to describe it in the comment but I didn't find it yet. 😅

web/libs/editor/src/components/TopBar/TopBar.jsx
Gondragos: Yep, we can have feature flag `FF_DEV_3873` turned off.
I mean it shouldn't be a usual case but for some reason I thought it's better to cover.
Although i'm not sure right now we cover all other functionality with `FF_DEV_3873` turned off, so maybe it's better to remove these 2 conditions at all. 🤔

web/libs/editor/src/components/BottomBar/buttons.tsx
Gondragos: At first a thought that you are right but then I figured out that it's different variables.
image

web/libs/editor/src/components/TopBar/TopBar.jsx
Gondragos: I have checked that. This conditions are covered by the test https://github.com/HumanSignal/label-studio/pull/6203/files#diff-15e1d28fc5b0c407a694fe6996ee68a73faa93b0ea1aa5d47bec1dd585ef82c8R146-R158
So I would say it's better to keep them there untill we remove the feature flag about old ui.

hlomzik (approved):

PR #6208, opened 2024-08-14
fix: OPTIC-1011: Making OneToOne to ForeignKey migration better 

wesleylima:
…es concurrently and change constraints more succinctly

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


label_studio/io_storages/migrations/0017_auto_20240731_1638.py
```diff
@@ -1,40 +1,99 @@
...
+                                            related_name=table['table_name'],
+                                            to='tasks.annotation'),
+                ),
+            )
+            continue
+
+        operations.append(
+            migrations.RunSQL(
+                sql=index_sql,
+                reverse_sql=RunSQL.noop
```
jombooth: Is it worth implementing the drop index / alter table drop constraint to enable a proper reverse migration?

jombooth:
I think this migration still isn't truly asynchronous, in the sense that if the queries from the migration haven't run in five minutes, we'll kill the Django process that's running this migration. Would it be better to perform the SQL work on an RQWorker and use AsyncMigrationStatus to keep track of whether it's completed or not? This code looks like a good candidate for being handled that way to me.

label_studio/io_storages/migrations/0017_auto_20240731_1638.py
```diff
@@ -1,40 +1,99 @@
...
-            model_name='s3exportstoragelink',
-            name='annotation',
-            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='io_storages_s3exportstoragelink', to='tasks.annotation'),
-        ),
-    ]
+    operations = []
+
+    for table in tables:
+        index_sql = create_index_sql(table['table_name'], table['index_name'], table['column_name'])
+        fk_sql = create_fk_sql(table['table_name'], table['fk_constraint'], table['column_name'], "task_completion", "id")
```
bmartel:
```
        index_sql = create_index_sql(table['table_name'], table['index_name'], table['column_name'])
        fk_sql = create_fk_sql(table['table_name'], table['fk_constraint'], table['column_name'], "task_completion", "id")
```
This should be moved to after the IS_SQLITE check, as the usage is only when it's not Sqlite.

label_studio/io_storages/migrations/0017_auto_20240731_1638.py
wesleylima: I added a `DROP INDEX` but I left the `noop` for the fk constraint since the reverse operation would involve re-adding a uniqueness constraint on the `annotation_id` column which may not be valid on the reversal. I did verify that the `IF NOT EXISTS ` and `IF EXISTS` keywords allow the forward migrations to work after migrations are reversed.

jombooth: Great point about re-adding the constraint no longer being valid, SGTM

wesleylima:
@jombooth My impression was that the CONCURRENT create would return fast but I think you're right.  While the table won't lock with CONCURENT, this would block the python process and would time out in some deployment. I moved the Postgres migration to Rqworkers.

label_studio/io_storages/migrations/0017_auto_20240731_1638.py
```diff
@@ -1,40 +1,158 @@
...
+
+    migration.status = AsyncMigrationStatus.STATUS_FINISHED
+    migration.save()
+    logger.debug(
+        f'Async migration {migration_name} complete'
+    )
+
+def reverse_migration(migration_name):
+    migration = AsyncMigrationStatus.objects.create(
+        name=migration_name,
```
jombooth, line 100: should we start adding a suffix to these like `_rev`?

jombooth (approved):
LGTM

bmartel (approved):

PR #6215, opened 2024-08-15
docs: DOC-233: Doc updates for adding Azure API keys

caitlinwheeless:
Updating the Prompts docs to include information about adding Azure OpenAI API keys to Label Studio. 

This affects:
- [ ] Community docs
- [X] Enterprise docs


docs/source/guide/prompts_create.md
```diff
@@ -14,20 +14,71 @@ date: 2024-06-11 16:53:16
...
 * A project that meets the following criteria:
   * Text-based data set (meaning you are annotating text and not image or video files). 
   * The labeling configuration for the project must be set up to use single-class classification (`choice="single"`). 
   * (Optional, depending on your [use case](prompts_overview#Use-cases) and if you want to evaluate the accuracy of your prompt): At least one task with a [ground truth annotation](quality#Define-ground-truth-annotations-for-a-project). 
 
-## API key
+## Model provider API keys
 
-You can only specify one OpenAI API key per organization, and it only needs to be added once. 
+You can specify one OpenAI API key and multiple Azure OpenAI keys per organization. Keys only need to be added once.
```
pakelley: nit: and/or? Wdyt?

docs/source/guide/prompts_create.md
```diff
@@ -14,20 +14,71 @@ date: 2024-06-11 16:53:16
...
+
+You can find your OpenAI API key on the [API key page](https://platform.openai.com/api-keys). 
+
+Once added, all supported OpenAI models will appear in the base model options when you configure your prompt.
+
+{% enddetails %}
+
+{% details <b>Use an Azure OpenAI key</b> %}
+
+Each Azure OpenAI key is tied to a specific deployment (OpenAI model). So if you want to use multiple models through Azure, you will need to add multiple keys.
```
pakelley: while different OpenAI models is the obvious use case, I don't think it's necessarily 1:1. For example, I think you can have multiple deployments for a single OpenAI model. I can't think of a great reason for this (maybe there are deployment-level settings that could be different enough to have multiple deployments? Like different rate limiting for different consumers of the deployments? idk), but it may be worth reflecting this here. On the other hand, saying "OpenAI models" makes it a decent amount clearar by showing why you'd have this, but idk...

Ultimately, if we're choosing this phrasing for clarity and accepting that it's not strictly accurate, that's ok with me

docs/source/guide/prompts_draft.md
```diff
@@ -16,7 +16,9 @@ With your [Prompt created](prompts_create), you can begin drafting your prompt c
 ## Draft a prompt and generate predictions
 
 
-1. Select your base model. For a description of all OpenAI models, see [OpenAI's models overview](https://platform.openai.com/docs/models/models-overview).
+1. Select your base model. 
+
+    The models that appear depend on the [API keys](prompts_create#Model-provider-API-keys) that you have configured for your organization. For a description of all OpenAI models, see [OpenAI's models overview](https://platform.openai.com/docs/models/models-overview).
```
pakelley: Might be worth explaining here that there will be 1 Azure model listed per deployment you add

pakelley (approved):
Nothing blocking, looks great!

PR #6230, opened 2024-08-20
refactor: remove unused code

luarmr:
I ran a tool to detect unused code and found a fair amount. I decided to simplify by removing entire files and some potentially confusing elements. Trying to avoid code that may be on current progression or false positives.


luarmr:
close to generate LSE branch here:
https://github.com/HumanSignal/label-studio/pull/6232

PR #6232, opened 2024-08-20
refactor: remove unused code

luarmr:
I ran a tool to detect unused code and found a fair amount. I decided to simplify by removing entire files and some potentially confusing elements. Trying to avoid code that may be on current progression or false positives.


web/libs/editor/tests/integration/data/lmm/accept_suggestions.ts
```diff
@@ -1,64 +0,0 @@
-import { LabelStudio } from "@humansignal/frontend-test/helpers/LSF";
```
yyassi-heartex, line 1: strange, i thought this would get picked up by our tests being in the `tests/integration` directory.
weirder still non of the other ones got picked up for deletion

yyassi-heartex (approved):
we gotta double check whats up with that test and why its getting removed while the others aren't
and we should archive all the deleted Feature flags in LD but other wise approved!

web/libs/editor/tests/integration/data/lmm/accept_suggestions.ts
luarmr: The list I gather was lengthy. I removed these files because they contained constants that were not in use. Therefore, the entire file can be deleted.

luarmr:
Thanks @yyassi-heartex for the review!

Only two feature flags are still referenced in the LD. I expected them to show up as removable when I ran the script for the FF state. 

For the test, this file contains only constants that are not in use, which is why I removed the entire file. However, there is more unused code around that requires further confirmation before removal I just pick the easy wins here.

luarmr:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/10511339169)
> Successfully merged:  create mode 100644 label_studio/ml_model_providers/migrations/0003_modelproviderconnection_cached_available_models.py

Issue #6296, opened 2024-09-03
A bug with CSRF  after update label-studio


lgq-garbage:
![17253509171808](https://github.com/user-attachments/assets/3351cb23-7767-47b9-b600-c7bece2fcf3d)

I update label-studio to the newest version and run  docker-compose up -d --build.    

But when I open the 8080 port, the website shows these.(Include edge or chrome

lgq-garbage:
This problem appears again. Anyone meets this problem？  I don’t know how to adjust files.

june-oh:
Same problem

june-oh:
I found a temporary workaround. Add `USE_ENFORCE_CSRF_CHECKS=false` to the `docker-compose.yml` file as shown below:
```
app:  
   stdin_open: true  
   tty: true  
   build: .  
   image: heartexlabs/label-studio:latest  
   restart: unless-stopped  
   expose:  
     - "8000"  
   depends_on:  
     - db  
   environment:  
     - DJANGO_DB=default  
     - POSTGRE_NAME=postgres  
     - POSTGRE_USER=postgres  
     - POSTGRE_PASSWORD=  
     - POSTGRE_PORT=5432  
     - POSTGRE_HOST=db  
     - LABEL_STUDIO_HOST=${LABEL_STUDIO_HOST:-}  
     - JSON_LOG=1  
     - USE_ENFORCE_CSRF_CHECKS=false     #add this line 
```

lgq-garbage:
Thanks，I&nbsp; will&nbsp; add&nbsp; this option&nbsp; and rebuild.



---Original---
From: ***@***.***&gt;
Date: Wed, Sep 4, 2024 18:42 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [HumanSignal/label-studio] A bug with CSRF  after updatelabel-studio (Issue #6296)




 
I found a temporary workaround. Add USE_ENFORCE_CSRF_CHECKS=false to the docker-compose.yml file as shown below:
[...]

PR #6344, opened 2024-09-09
fix: LEAP-1456: CSRF - use $http_host instead of nginx $host so port will be included

jombooth:
`host` is just `localhost` when LSO is running in docker; `http_host` has the port number. If the port number isn't present in the result from `get_host`, this check will fail: https://github.com/django/django/blob/stable/4.2.x/django/middleware/csrf.py#L277-L288

### DIFF
--- a/deploy/default.conf
+++ b/deploy/default.conf
@@ -184,7 +184,7 @@ http {
         send_timeout 90;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


bmartel (approved):

TheAIWizard:
Hi,

I'm still encountering the 403 Forbidden error related to CSRF, mentioned in #6296, even after updating to version 1.14.0, which was supposed to address the CSRF errors when using Docker (#6344). 

Could you please assist with resolving this?

Thanks in advance!

PR #6406, opened 2024-09-18
fix: OPTIC-1120: Draft lead time incorrectly calculated which can cause annotation lead time to be wrong

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

When the current annotation would be a draft, based on the logic previously there it would assign total accumulated leadTime to both the current draft and any annotations. There was a point in time where the draft and annotation would be loaded into state, and the calculation made here would not properly factor in that drafts accumulate and eventually add the current running total of the draft over to the annotation once submitted. Because of this, we would end up doubling the value stored each time a draft would be saved, and once applied to the annotation.

An example I was able to reproduce was as follows:

1. Open a task in quick view with no current annotations from my user.
2. Annotate something and allow a draft to save
3. Let some time pass noting the approximate time elapsed.
4. Submit the annotation.
5. Annotate or change something to produce the next draft
6. Leave the task
7. Return to the task after some amount of time noting the approximate time elapsed.
8. Annotate or change something to produce the next draft save
9. And continue to cycle draft, draft update, annotation submit, leaving and returning to task to simulate normal usage.
10. Eventually the value of lead time would become extremely large, larger than even the total time from first time the task was opened and the first draft was produced.

The numbers I was able to find just infrequently working through a task with the above steps over the course of 4 hours, was:

- Reported LeadTime: **119 hours**
- Approximate LeadTime: **1.7 hours** (this is what I calculated it should have been)

It cannot be possible to have a lead time that is larger than the time from which the task was first taken and the first interaction made.

This is how it should be calculating:

Draft Created      → **0s**   DraftLeadTime
Elapsed                → 10s
Draft Save            → **10s** DraftLeadTime
Elapsed                → 5s
Draft Save            → **15s** DraftLeadTime
Draft Submitted   → 0 + 15 = **15s** LeadTime
Draft Created       → **0s**   DraftLeadTime
Elapsed                 → 20s
Draft Save             → **20s** DraftLeadTime
Elapsed                 → 25s
Draft Save             → **45s** DraftLeadTime
Draft Submitted   → 15 + 45 = **60s** LeadTime

The draft values should be rolling up to the annotation it would be converted to on submit, not incorrectly adding all previous annotation values on every draft save.

_(check only one)_
- [X] No

(Annotation, Draft, History) Lead Time


web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -915,7 +915,7 @@ export class LSFWrapper {
     const sessionTime = (new Date() - annotation.loadedDate) / 1000;
     const submittedTime = Number(annotation.leadTime ?? 0);
     const draftTime = Number(this.task.drafts[0]?.lead_time ?? 0);
-    const lead_time = sessionTime + submittedTime + draftTime;
+    const lead_time = draft ? draftTime + sessionTime : submittedTime + draftTime + sessionTime;
```
yyassi-heartex: this is absolutely fine but for readability, what do you think about 
```suggestion
    const lead_time = (draft ? 0 : submittedTime) + draftTime + sessionTime;
```

yyassi-heartex (approved):
minor tweak for readability otherwise approved

yyassi-heartex (approved):

hlomzik (approved):

AndrejOros (approved):

bmartel:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11032856079)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/linking_modes/to_comments.cy.ts

PR #6411, opened 2024-09-19
feat: LEAP-1454: Custom action buttons in LSF

hlomzik:
Introducing custom action buttons for LSF to allow additional actions, controlled from outside of editor. Provided buttons will completely replace the internal ones, but there is a way to reuse buttons you need, just add the button with internal name and no extra params.

TODO: Add example

Custom buttons need to behave similarly to internal ones, so we are using flags and additional checks. They will call the same event `handleCustomButton(button_name)` so external app will handle `customButton` event with `store, button_name, options` params.

Also new types for global MST stores are introduced in a draft state with intention to expand them with more props and methods so they can be used in other TypeScript code, which currently uses `any` for any MST stuff.

- [x] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

We need to introduce new buttons in LSE and instead of creating another "hidden" buttons in LSF we create a way to use any button you need.

Hardcode new buttons in LSF and display them by one more interface.

None. That's addition and will be used only if specific props are passed in.

- [x] No

- [ ] e2e
- [ ] integration
- [ ] unit


web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -80,7 +128,8 @@ export const Controls = controlsInjector(
       ],
     );
 
-    const RejectButton = useMemo(() => {
+    // @todo memo?
```
hlomzik:
```suggestion
```

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -80,7 +132,7 @@ export const Controls = controlsInjector(
       ],
     );
 
-    const RejectButton = useMemo(() => {
+    const RejectButton = memo(({ disabled, store }: { disabled: boolean, store: MSTStore }) => {
```
Gondragos: Maybe it's better to move elements outside the `Controls`, 'cause I guess in this case they'll be created on each render.

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -102,12 +154,10 @@ export const Controls = controlsInjector(
...
         </ButtonTooltip>
       );
-    }, [disabled, store]);
-
-    if (isReview) {
-      buttons.push(RejectButton);
+    });
 
-      buttons.push(
+    const AcceptButton = memo(({ disabled, history, store }: { disabled: boolean, history: any, store: MSTStore }) => {
```
Gondragos: The same as for `RejectButton` https://github.com/HumanSignal/label-studio/pull/6411/files#r1768908443

web/libs/editor/src/components/BottomBar/Controls.tsx
hlomzik: Only functions created at each render, which is usual sacrifice nowadays. These two buttons don't have external dependencies, but others do, so for now I'd leave them as is to be more bulletproof, but I have a draft of new Controls which will improve this situation of course

Gondragos: It's not about functions, it's about memo.
https://codesandbox.io/p/sandbox/react-memo-example-forked-f8yjgc?workspaceId=91ee8eb7-598d-4294-8b94-c2ca887b1c01

bmartel: I agree with what @Gondragos mentioned here. The `memo()` is for component declaration, as it memoizes it based on the props, such that the usage of it would not re-render unless the props had reason to change. But because this is defined not in the module space, but within another component, it will be the wrong layer of memoization. Because it is within another component, it would have to use a hook `useMemo` to memoize the function declaration, not the return value of calling the function itself. The difference here would mean that depending on how it is used, the value of the `RejectButton` reference when rendered would cause other components to also re-render if they are memoized, because the reference is recreated each time. Or of course if there are no scope dependencies the best way is to just lift this out to a module level component, which you can decide at that level if `memo()` is still needed or not.

Gondragos (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11013541980)
> Successfully merged:  create mode 100644 label_studio/ml_models/migrations/0012_alter_thirdpartymodelversion_provider.py

bmartel (approved):

PR #6417, opened 2024-09-20
fix: OPTIC-1125: Label config preview always showing a step behind when making updates

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

The change originally made to fix an issue with taxonomy loading after updating the Label Config made it such that the preview was always a step behind changes that were made by the user. This addresses this by moving all the associated logic of updating and initializing to the same useEffect.

The following now occurs:

- Only a single initialization takes place for the preview
- Updates apply consistently on change of either config or task data.
- Only a single destroy will take place when the preview unmounts (important fix to the original issue!)

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Label Config Preview


web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
```diff
@@ -9,15 +9,21 @@ import { useAPI } from "../../../providers/ApiProvider";
...
+let dependencies;
+const loadDependencies = async () => {
+  if (!dependencies) {
+    dependencies = import("@humansignal/editor");
+  }
+  return dependencies;
+};
 
 export const Preview = ({ config, data, error, loading, project }) => {
+  loadDependencies();
```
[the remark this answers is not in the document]
bmartel: @ellipsis-dev This is explicitly addressed at the module level to ensure we are loading this as early as the first render of the component. We purposely are making a check to ensure this only resolves once, and the same promise is awaited in a useEffect as well so this is inaccurate.

hlomzik, line 24: this way of calling it doesn't feel correct... besides of the comment from ellipsis it's also called synchronously.
or I'm just lacking some knowledge of dynamic `import()`. if you call it repeatedly and in parallel would the module be imported only once and the same instance will be returned in different places?

bmartel: >if you call it repeatedly and in parallel would the module be imported only once and the same instance will be returned in different places?

Technically there is no parallel capability but the dynamic import will happen only once, right on the first render cycle of the component. That starts the async load, which then is awaited in the useEffect to ensure it completes. It all references the same promise, and only assigns it once.

This is basically the loader pattern introduced with React 18, and Suspense, but without all that jazz.

bmartel: As mentioned in DM, we can go with a different approach if you are feeling this is not resonating.

bmartel: As discussed, added a comment here about what this is doing.

hlomzik (approved):

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
```diff
@@ -53,12 +62,14 @@ export const Preview = ({ config, data, error, loading, project }) => {
   }, [config]);
 
   const initLabelStudio = useCallback(async (config, task) => {
-    if (!task.data) return;
+    // wait for dependencies to load, the promise is resolved only once
+    // and is started when the component is mounted for the first time
+    await loadDependencies();
 
-    await resolvingEditor;
+    if (lsf.current || !task.data) return;
```
yyassi-heartex, line 70: `!task.data` im not sure this will calculate as expected.
`!{}` is false but so is `!{id:1}` is there a better way to test for this? does it start undefined and become an object after?

yyassi-heartex: image

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
```diff
@@ -126,18 +122,15 @@ export const Preview = ({ config, data, error, loading, project }) => {
       });
 
       store.annotationStore.selectAnnotation(c.id);
-      console.log("LSF task updated");
+      console.log("LSF updated");
```
yyassi-heartex: this is a good opportunity to remove this to reduce the console flood - alternatively maybe we can do something like 
```suggestion
      if(window.APP_SETTINGS.debug) {
            console.log("LSF updated");
      }
```

yyassi-heartex:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11018484288)
> Successfully merged:  create mode 100644 web/libs/editor/src/stores/Comment/Anchor.js

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
hlomzik: @yyassi-heartex we specifically check for missing data (e.g. not loaded yet), not empty one, which is actually a valid data.

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
hlomzik: good idea, but maybe not to do it via APP_SETTINGS, which is coming from a backend. we should be able to debug with clients so we have to be able to enable debug in runtime. we still can do `window.APP_SETTINGS.debug = true` though

PR #6426, opened 2024-09-23
fix: LEAP-1258: prevent dropdown from opening upward when viewport short

jombooth:
## before:
image
image

## after:
image
image

### DIFF
--- a/web/libs/datamanager/src/components/Common/Dropdown/DropdownComponent.jsx
+++ b/web/libs/datamanager/src/components/Common/Dropdown/DropdownComponent.jsx
@@ -17,7 +17,7 @@ export const Dropdown = React.forwardRef(({ animated = true, visible = false, ..
   const { triggerRef } = React.useContext(DropdownContext) ?? {};
   const isInline = triggerRef === undefined;
 
-  const { children, align } = props;
+  const { children, align, openUpwardForShortViewport } = props;
   const [currentVisible, setVisible] = React.useState(visible);
   const [offset, setOffset] = React.useState({});
   const [visibility, setVisibility] = React.useState(visible ? "visible" : null);
@@ -29,8 +29,8 @@ export const Dropdown = React.forwardRef(({ animated = true, visible = false, ..
       parent,
       dropdownEl,
       align ?? "bottom-left",
-      /* padding = */ 0,
-      /* openUpwardForShortViewport = */ false,
+      0,
+      openUpwardForShortViewport ?? true,
     );
 
     setOffset({ left, top });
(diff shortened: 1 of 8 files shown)


jombooth:
FYI this also applies to the other dropdowns (columns, filters, order)

web/libs/datamanager/src/utils/dom.js
```diff
@@ -70,7 +70,7 @@ export const alignElements = (elem, target, align, padding = 0) => {
   if (offsetTop < window.scrollX) {
     offsetTop = pos.bottom + padding;
     resultAlign[0] = "bottom";
-  } else if (offsetTop + pos.target.height > window.scrollX + window.innerHeight) {
+  } else if (openUpwardForShortViewport && offsetTop + pos.target.height > window.scrollX + window.innerHeight) {
```
hlomzik: I see a huge bug here with wrong dimensions in old code — "top" is compared to "x" instead of "y". The main reason I guess it didn't cause troubles is that dropdowns are used in a non-scrollable areas maybe. But it should be fixed.

web/libs/datamanager/src/components/Common/Dropdown/DropdownComponent.jsx
```diff
@@ -25,7 +25,13 @@ export const Dropdown = React.forwardRef(({ animated = true, visible = false, ..
   const calculatePosition = React.useCallback(() => {
     const dropdownEl = dropdown.current;
     const parent = triggerRef?.current ?? dropdownEl.parentNode;
-    const { left, top } = alignElements(parent, dropdownEl, align ?? "bottom-left");
+    const { left, top } = alignElements(
+      parent,
+      dropdownEl,
+      align ?? "bottom-left",
+      /* padding = */ 0,
+      /* openUpwardForShortViewport = */ false,
```
hlomzik: we don't normally do this in js — IDE (and even github now) helps with function defeinitions + for really big list of options we use `options` as an object.

hlomzik: but the main concern here is that this is the common component, so it should not force the behaviour, it should receive it from the consumer. common component can be used in other cases where this option should be true, like on the bottom of the page or in a scrollable areas.

web/libs/datamanager/src/utils/dom.js
```diff
@@ -70,7 +70,7 @@ export const alignElements = (elem, target, align, padding = 0) => {
   if (offsetTop < window.scrollX) {
     offsetTop = pos.bottom + padding;
     resultAlign[0] = "bottom";
-  } else if (offsetTop + pos.target.height > window.scrollX + window.innerHeight) {
+  } else if (openUpwardForShortViewport && offsetTop + pos.target.height > window.scrollY + window.innerHeight) {
```
hlomzik, line 73: `scrollX` should be fixed in a previous condition as well

web/libs/datamanager/src/utils/dom.js
```diff
@@ -70,7 +70,7 @@ export const alignElements = (elem, target, align, padding = 0) => {
   if (offsetTop < window.scrollX) {
```
hlomzik:
```suggestion
  if (offsetTop < window.scrollY) {
```

hlomzik (approved):

PR #6428, opened 2024-09-24
fix: DIA-1428: RunTime error occurs on trying to delete certain prompts

niklub:
Prior to this PR, deleting certain prompts would lead to a runtime error:

```
new row for relation "prediction_meta" violates check constraint "prediction_or_failed_prediction_not_null"
DETAIL: Failing row contains (104, null, 113, 6, 119, null, null, null, null, null, null)
```

Use `CASCADE` setting in `on_delete` instead of Django signals to delete `PredictionMeta` objects associated with `Prediction` and `FailedPrediction` objects that are themselves deleted.


label_studio/ml_models/models.py
```diff
@@ -191,27 +191,37 @@ def delete_predictions(self):
...
         # delete predictions meta
         PredictionMeta.objects.filter(prediction__in=prediction_ids).delete()
+        logger.debug(f'PredictionMeta for predictions {prediction_ids} have been deleted')
         PredictionMeta.objects.filter(failed_prediction__in=failed_predictions_ids).delete()
+        logger.debug(f'PredictionMeta for failed predictions {failed_predictions_ids} have been deleted')
 
         # remove predictions from db
-        predictions._raw_delete(predictions.db)
-        failed_predictions._raw_delete(failed_predictions.db)
+        predictions.delete()
```
hakan458: We were doing raw_delete for performance reasons, regular delete was taking a long time when we have lots of predictions. Deleting happens at the time of re-running a model version or updating a model version, and it was degrading UX so please test with a large project for example

hakan458 (approved):

jombooth (approved):

PR #6470, opened 2024-10-05
feat: LEAP-1474: Create and display comments with classifications

jombooth:
Implements classifications on comments, leveraging new-style taxonomy. Classifications are loaded from and persisted to the DB. Also allows making a comment that's classifications-only and has no text.

Classifications are **non-editable** for this PR.

### A couple examples of comments with associated regions, one with text and one without:
image

### Comments without associated regions with and without text:
image

### Comment interface when org has a classification config, with and without linking mode:
image
image

### Comment interface when org has no classification config, with and without linking mode:
image
image


web/libs/editor/src/components/Comments/Comment/CommentClassificationUtil.tsx
bmartel, line 1: Just a particularity, but this file is only housing classification utils and no actual jsx components, probably can be renamed to something like: `classificationUtils.ts`.

web/libs/editor/src/components/Comments/Comment/CommentForm.scss
```diff
@@ -56,6 +54,8 @@
     background-color: transparent;
     color: var(--secondary-action-color);
     border-radius: var(--primary-action-border-radius);
+    padding-left: 8px;
+    padding-right: 8px;
```
bmartel:
```suggestion
    padding-inline: 8px;
```

web/libs/editor/src/components/Comments/Comment/CommentForm.tsx
```diff
@@ -114,36 +127,69 @@ export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annot
...
+          <Elem name="action" tag="button" mod={{ highlight: linking }} onClick={linkToHandler}>
+            <IconCommentLinkTo />
+          </Elem>
+        </Tooltip>
+      )}
+      <Elem name="action" tag="button" type="submit">
+        <IconSend />
+      </Elem>
+    </Elem>
+  );
```
bmartel: Component defined in a component is usually a sign to lift it out, which ensures that all components remain easy to read through with regards to scope and intent.

web/libs/editor/src/components/Comments/Comment/CommentForm.tsx
```diff
@@ -114,36 +127,69 @@ export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annot
...
+                    ? {
+                        default: {
+                          type: "taxonomy",
+                          values,
+                        },
+                      }
+                    : null;
+                updateCommentClassifications(newClassifications);
+              }}
+              options={{ pathSeparator: "/", showFullPath: true }}
```
bmartel: Given that this has inline definitions for both options and onChange, if the underlying component uses any memoization for performance on lengthy lists or deep tree rendering, this can cause a lot of unnecessary re-renders for sibling or parent change which would not warrant this component to need to update.

One way to avoid that is to pull the defined onChange as a useCallback in the component with `updateCommentClassifications` as its only dependency:

```tsx
...
  const updateCommentClassifications = useCallback(
    (classifications: object | null) => {
      const currentComment = getCurrentComment();
      currentComment.setClassifications(classifications);
    },
    [commentStore, annotationStore],
  );

  const taxonomyOnChange = useCallback(async (_, values) => {
    const newClassifications =
      values.length > 0
        ? {
            default: {
              type: "taxonomy",
              values,
            },
          }
        : null;
    updateCommentClassifications(newClassifications);
  }, [updateCommentClassifications]);
```


Also lift the options `{ pathSeparator: "/", showFullPath: true }` to a module scope (outside the component) const definition:

```tsx
...
const TOOLTIP_DELAY = 0.8;
const TAXONOMY_OPTIONS = { pathSeparator: "/", showFullPath: true };

export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annotationStore, inline = true }) => {
...
```

```suggestion
            <Taxonomy
              selected={selections}
              items={classificationsItems}
              onChange={taxonomyOnChange}
              options={TAXONOMY_OPTIONS}
```

web/libs/editor/src/components/Comments/Comment/CommentItem.scss
```diff
@@ -151,4 +151,9 @@
     justify-content: space-between;
     padding: var(--comment-item-padding);
   }
+
+  &__classifications {
+    font-size: 13px;
+    letter-spacing: .05em;
```
bmartel: That is a very specific sizing, is it something pulled from the design file? I haven't seen us ever use `em` or a font size of 13px.

bmartel (approved):
Nice work 🔥

web/libs/editor/src/components/Comments/Comment/CommentForm.scss
hlomzik: wow! TIL, thank you! :)
I knew there should be something for this! and it's even more semantic

web/libs/editor/src/components/Comments/Comment/CommentForm.tsx
```diff
@@ -114,36 +127,69 @@ export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annot
 
   const currentLinkingComment = annotationStore.selected.currentLinkingMode?.comment;
   const currentComment = getCurrentComment();
-  const { text = "", regionRef } = currentComment || {};
+  const { text = "", regionRef, classifications } = currentComment || {};
   const { region, result } = regionRef || {};
   const linking = !!linkingComment && currentLinkingComment === linkingComment && globalLinking;
   const hasLinkState = linking || region;
+  const selections = taxonomyPathsToSelectedItems(classifications?.default?.values);
+  const classificationsItems = parseCommentClassificationConfig(commentStore.commentClassificationConfig);
```
hlomzik: we should not make calculations like this directly in render. At least `useMemo()` should be used, but in this case it should be moved to a getter in comment store, because it will be calculated only once.

web/libs/editor/src/components/Comments/Comment/CommentForm.tsx
```diff
@@ -114,36 +127,69 @@ export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annot
 
   const currentLinkingComment = annotationStore.selected.currentLinkingMode?.comment;
   const currentComment = getCurrentComment();
-  const { text = "", regionRef } = currentComment || {};
+  const { text = "", regionRef, classifications } = currentComment || {};
   const { region, result } = regionRef || {};
   const linking = !!linkingComment && currentLinkingComment === linkingComment && globalLinking;
   const hasLinkState = linking || region;
+  const selections = taxonomyPathsToSelectedItems(classifications?.default?.values);
```
hlomzik: and this should be transformed to `useMemo()`

web/libs/editor/src/components/Comments/Comment/CommentClassificationUtil.tsx
hlomzik: ... and moved to utils

web/libs/editor/src/components/Comments/Comment/CommentClassificationUtil.tsx
```diff
@@ -0,0 +1,39 @@
+import type { TaxonomyItem, SelectedItem, TaxonomyPath } from "../../../components/NewTaxonomy/NewTaxonomy";
+
+export const parseCommentClassificationConfig = (xmlString: string): TaxonomyItem[] => {
```
hlomzik: please, specify the expected xml format

web/libs/editor/src/components/Comments/Comment/CommentItem.scss
jombooth: I was emulating the look of these typography options in figma; know that it's a bit weird. Changing from em to px

image

jombooth: (feel free to reopen btw)

web/libs/editor/src/components/Comments/Comment/CommentForm.tsx
```diff
@@ -114,36 +118,69 @@ export const CommentForm: FC<CommentFormProps> = observer(({ commentStore, annot
...
   const hasLinkState = linking || region;
+  const selections = useMemo(() => taxonomyPathsToSelectedItems(classifications?.default?.values), [classifications]);
+  const classificationsItems = commentStore.commentClassificationsItems;
+
+  const updateCommentClassifications = useCallback(
+    (classifications: object | null) => {
+      const currentComment = getCurrentComment();
+      currentComment.setClassifications(classifications);
+    },
+    [commentStore, annotationStore],
```
bmartel:
```suggestion
    [getCurrentComment],
```
I didn't see this one earlier, but if the only reason this function has to change is if the other function changes, it is the only dependency required here. commentStore and annotationStore are not used within this function, only getCurrentComment.

web/libs/editor/src/components/Comments/Comment/CommentFormButtons.scss
bmartel, line 1: @luarmr does the `yarn lint:scss` work in this repo? Seems the linter should have caught the indentation?

@jombooth Can you try running from the `web/` directory:

```shell
yarn lint:scss
```

web/libs/editor/src/stores/Comment/Comment.js
```diff
@@ -174,6 +182,10 @@ export const Comment = CommentBase.named("Comment")
       self.update({ regionRef });
     }
 
+    function setClassifications(classifications) {
+      self.update({ classifications });
+    }
+
```
bmartel:
```suggestion
```
This one appears unused, only the CommentBase definition will be running.

web/libs/editor/src/utils/commentClassification.ts
```diff
@@ -0,0 +1,56 @@
+import type { TaxonomyItem, SelectedItem, TaxonomyPath } from "../components/NewTaxonomy/NewTaxonomy";
+
+export const parseCommentClassificationConfig = (config: string | null): TaxonomyItem[] => {
+  /**
+   * Assume that there is a single root Taxonomy element for now, which may have multiple
+   * TaxonomyItem childrenl, as in this simple example below.
```
hlomzik:
```suggestion
   * TaxonomyItem children, as in this simple example below.
```

hlomzik (approved):

web/libs/editor/src/components/Comments/Comment/CommentFormButtons.scss
jombooth:
```
~/Repos/label-studio/web on  fb-LEAP-1474/classifications-in-comments [$?] via  v18.16.1 on ☁️  jo@humansignal.com 
❯ yarn lint-scss
yarn run v1.22.19
$ yarn stylelint '**/*.scss' --fix
$ /home/jo/Repos/label-studio/web/node_modules/.bin/stylelint '**/*.scss' --fix
Done in 5.36s.
```

Doesn't seem to have had any effect

jombooth: ended up using copilot to reformat the file, which resulted in whitespace-only changes :)

PR #6489, opened 2024-10-08
refactor: OPTIC-1200: stale ff fflag_feat_all_optic_114_soft_delete_f…

luarmr:
Stale a feature flag that is already filtering in the cloud. We know that disablig this has a negative impact on performance and is an unsupported path, so it should remain enabled.


makseq (approved):

bmartel (approved):

luarmr:
@makseq @bmartel added extra code

makseq (approved):

bmartel (approved):

PR #6490, opened 2024-10-08
docs: DOC-254: Prompts updates

caitlinwheeless:
Updates to Prompts covering:

- Ability to save as a version
- Evaluation metrics section

- [ ] Community docs
- [X] Enterprise docs


pakelley (approved):
Lgtm, fwiw looks like we should probably record a new video for this page soon though (the current one only has the accuracy and outputs)

PR #6501, opened 2024-10-10
ci: LEAP-1553: Fix path-to-regexp CVE

hlomzik:
Fixes https://github.com/HumanSignal/label-studio/security/dependabot/313 and https://github.com/HumanSignal/label-studio/security/dependabot/314


jombooth (approved):
ty

PR #6520, opened 2024-10-15
fix: LEAP-1560: Add interface for requiring comments on rejection in QuickView

Gondragos:
This PR makes an option
image
work with QuickView when it is in a Review mode.

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

In the context of last changes we gain an ability to make a review from QuickView. But as QuickView and ReviewStream modes are separated in code, we missed some interfaces when moved functionality.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

 `QuickView`

### DIFF
--- a/web/libs/datamanager/src/sdk/lsf-sdk.js
+++ b/web/libs/datamanager/src/sdk/lsf-sdk.js
@@ -155,6 +155,10 @@ export class LSFWrapper {
       interfaces.push("comments:resolve-any");
     }
 
+    if (this.project.review_settings.require_comment_on_reject) {
+      interfaces.push("comments:reject");
+    }
+
     if (this.interfacesModifier) {
       interfaces = this.interfacesModifier(interfaces, this.labelStream);
     }


web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -155,6 +155,10 @@ export class LSFWrapper {
       interfaces.push("comments:resolve-any");
     }
 
+    if (this.project.review_settings.require_comment_on_reject) {
```
yyassi-heartex, line 158: i see this flag being used in another spot that is also checking for `ff_front_dev_2186_comments_for_update` do we need to encapsulate this in this FF too?

yyassi-heartex (approved):
looks good - just double check if we need to add FF

web/libs/datamanager/src/sdk/lsf-sdk.js
Gondragos: Great catch! For a second, I thought I needed to add it here, but then I saw how it had been implemented.

https://github.com/HumanSignal/label-studio/blob/develop/web/libs/datamanager/src/sdk/lsf-sdk.js#L124-L126
`FF_DEV_2186` is a string 🥲

yyassi-heartex: ahh, true thats not even checking it properly.
I also looked at who that FF is enabled for and it only raised more questions. seems to be on for a handful of users and off for every one else. need more info on what happened there. but as it stands - we should fix how this is being checked (probably another ticket)

hlomzik (approved):

PR #6532, opened 2024-10-16
feat: LEAP-1476: Add ability to edit classifications on existing comments

jombooth:
If there's no taxonomy defined or the FF is off, there are no changes. Otherwise, allow submitting a comment edit if it has either a category selected or a nonempty body (just like when creating a comment). Best viewed with `?w=1` because the formatter made a bunch of changes to whitespace. Simple UX has been cleared with @alecjonharris.

Screenshot from 2024-10-16 15-46-29

### DIFF
--- a/web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
+++ b/web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
@@ -13,9 +13,13 @@ import { Block, Elem } from "../../../utils/bem";
 import { humanDateDiff, userDisplayName } from "../../../utils/utilities";
 import { CommentFormBase } from "../CommentFormBase";
 import { CommentsContext } from "./CommentsList";
+import { NewTaxonomy as Taxonomy, type TaxonomyPath } from "../../../components/NewTaxonomy/NewTaxonomy";
+import { taxonomyPathsToSelectedItems } from "../../../utils/commentClassification";
 
 import "./CommentItem.scss";
 import { LinkState } from "./LinkState";
+
+const TAXONOMY_OPTIONS = { pathSeparator: "/", showFullPath: true };
 
 interface CommentItemProps {
   comment: {
@@ -33,6 +37,7 @@ interface CommentItemProps {
     updateComment: (comment: string, classifications?: any) => void;
     deleteComment: () => void;
     setConfirmMode: (confirmMode: boolean) => void;
+    setClassifications: (classifications: any) => void;
     setEditMode: (isGoingIntoEditMode: boolean) => void;
     toggleResolve: () => void;
     canResolveAny: boolean;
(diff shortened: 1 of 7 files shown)


web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
```diff
@@ -42,198 +45,227 @@ interface CommentItemProps {
...
+                          values.length > 0
+                            ? {
+                                default: {
+                                  type: "taxonomy",
+                                  values,
+                                },
+                              }
+                            : null;
+                        setClassifications(newClassifications);
+                      }}
```
bmartel: This should be extracted to a function with a useCallback wrapping it, otherwise there is a chance that on each keystroke of the form the Taxonomy would be re-rendered since this function will change each render of the parent, influenced by all other siblings as well.

web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
```diff
@@ -42,198 +45,227 @@ interface CommentItemProps {
...
+                    await updateComment(value, classifications);
+                    setText(value);
+                    await listComments({ suppressClearComments: true });
+                  }}
+                  classifications={classifications}
+                />
+                {classificationsItems.length > 0 && (
+                  <Elem name="classifications-row">
+                    <Taxonomy
+                      selected={taxonomyPathsToSelectedItems(classifications?.default?.values)}
```
bmartel: This should be lifted up to the hook definitions and wrapped in a useMemo otherwise the underlying map performed within the utility function will produce a new array each time the surrounding components would render, and force Taxonomy to also re-render.

web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
```diff
@@ -42,198 +45,227 @@ interface CommentItemProps {
...
+          <Elem name="text">
+            {isEditMode ? (
+              <>
+                <CommentFormBase
+                  value={text}
+                  onSubmit={async (value) => {
+                    await updateComment(value, classifications);
+                    setText(value);
+                    await listComments({ suppressClearComments: true });
+                  }}
```
bmartel: This should similarly to the other suggestions be extracted and wrapped to a useCallback with a dependency on classifications. Not sure what underpins the CommentFormBase but likely not as big of a problem for re-renders as Taxonomy could be.

web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
jombooth: Is it enough to add just `setClassifications` to the callback's depenendencies?

web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
jombooth: Here we don't need dependencies on `updateComment` or `listComments`?

bmartel: That is a great question. Only if the functions are useCallback based and require their own dependencies. If they are useCallback with no dependencies that is considered stable and not required. In general the best rule of thumb is add all dependencies except for refs (useRef() created values), and externally defined functions. Now the only question that remains is to ensure the functions that are added as dependencies are properly using useCallback, or considered stable.

bmartel: These look like props, would definitely affect the render of this component if not properly defined with useCallback in the parent context, but should be included in this useCallback based on the fact they are prop based (already influences the component itself). Otherwise the reference will have potential to go stale and cause bugs.

web/libs/editor/src/components/Comments/Comment/CommentItem.tsx
bmartel: Yes, that is the only function which is used within. I went into more depth above about what should be included, but a good default is to just include the things used, unless it is a ref or externally defined function that has no component scope dependencies.

bmartel (approved):
Nicely done 🔥

yyassi-heartex (approved):

PR #6534, opened 2024-10-17
chore: LEAP-1275: Bump sentry-sdk to fix vulnerability, py3.13 compatibility, remove py3.8 support

jombooth:
- Bump all packages that didn't support PEP 517 builds
- Minimum python version has to be 3.9 for `numpy`, but 3.8 was EOL'd 10/7 anyway
- sentry_sdk is getting a major version bump from 1 to 2, but we're already running 2.15 on some deployments without issues
- Adjust CI workflows to include 3.13 and disinclude 3.8
- Deprecate py3.8 specific code for an old windows DLL fix


label_studio/data_import/uploader.py
```diff
@@ -74,9 +74,9 @@ def create_file_upload(user, project, file):
     if settings.SVG_SECURITY_CLEANUP:
         content_type, encoding = mimetypes.guess_type(str(instance.file.name))
         if content_type in ['image/svg+xml']:
-            clean_xml = allowlist_svg(instance.file.read())
+            clean_xml = allowlist_svg(instance.file.read().decode())
             instance.file.seek(0)
-            instance.file.write(clean_xml)
+            instance.file.write(clean_xml.encode())
```
jombooth, line 79: These changes were needed because lxml's `clean_html` seems to no longer work on bytes-like objects in recent versions. I hope it's okay to decode and encode the svg string?

nikitabelonogov (approved):

makseq (approved):

label_studio/data_import/uploader.py
jombooth: tested locally and seems to work right - only real concern area I could think of would be giant SVG uploads that are expensive to decode, but that seems like it'd cause other problems anyway.

PR #6583, opened 2024-10-30
feat: LEAP-1600: Add created_at for TaskLock

triklozoid:
* add `created_at` field for TaskLock model and fill it with proper values in async migration

table state before rqworker job start: 
Screenshot_2024-11-04_10-40-36


label_studio/tasks/models.py
```diff
@@ -798,6 +798,7 @@ class TaskLock(models.Model):
         on_delete=models.CASCADE,
         help_text='User who locked this task',
     )
+    created_at = models.DateTimeField(_('created at'), auto_now_add=True, help_text='Creation time', null=True)
```
jombooth, line 801: I believe auto_now_add will set created_at to the current time for all existing rows, and it'll do so without using a db default, making this an unsafe migration (and also the null=True will have no effect)

triklozoid: Fixed.

jombooth (approved):

triklozoid:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11672828208)
> Successfully merged:  delete mode 100644 web/libs/editor/tests/integration/e2e/outliner/filter.cy.ts

PR #6585, opened 2024-10-31
ci: PLT-536: rework migration script to use pg_try_advisory_xact_lock

farioas:
Switch to pg_try_advisory_xact_lock for locked_migrations

### DIFF
--- a/label_studio/core/management/commands/locked_migrate.py
+++ b/label_studio/core/management/commands/locked_migrate.py
@@ -1,5 +1,5 @@
-import time
 import logging
+import time
 
 from django.conf import settings
 from django.core.management.commands.migrate import Command as MigrateCommand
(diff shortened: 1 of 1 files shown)


label_studio/core/management/commands/locked_migrate.py
```diff
@@ -33,10 +32,17 @@ def handle(self, *args, **options):
...
+            cursor.execute(f'SELECT pg_advisory_lock({lock_id})')
+            logger.info('Successfully acquired the lock. Starting migrations...')
             try:
-                logger.info(f'Attempting to acquire a migration lock: {lock_id}')
-                cursor.execute(f'SELECT pg_advisory_xact_lock({lock_id})')
-                logger.info('Successfully acquired the lock. Starting migrations...')
                 MigrateCommand.handle(self, *args, **options)
+            except DatabaseError as migration_error:
+                logger.error(f'Error during migration execution: {migration_error}')
+                raise
```
jombooth: ah I realize I misunderstood the `try...finally` approach in the original version, I don't think this is needed and `try...finally` without any `except` block here or below is fine.

triklozoid (approved):

jombooth (approved):
LGTM, please update the description to describe the solution in greater detail and correct the pg lock function used (should be `pg_try_advisory_xact_lock`)

PR #6613, opened 2024-11-06
fix: OPTIC-1287: Project config validation should ignore xml comments

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [X] Backend (API)
- [ ] Frontend

When using xml comments which may contain code samples that have a name property like`<!--document.querySelector('[name="field"]')-->` that also occurs in the label config xml `<Taxonomy name="field">` it would fail to validate stating that Config cannot contain non-unique names. This is not correct as the values do not occur within the Label config portion of the xml.

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [X] unit

Project Config Validation

### DIFF
--- a/label_studio/core/label_config.py
+++ b/label_studio/core/label_config.py
@@ -76,7 +76,9 @@ def _fix_choices(config):
     return config
 
 
-def remove_xml_comments(config_string):
+def remove_xml_comments(config_string: str | None) -> str | None:
+    if config_string is None:
+        return None
     return re.sub(r'<!--[\s\S]*?-->', '', config_string)
 
 
@@ -92,7 +94,7 @@ def parse_config_to_json(config_string):
     return config
 
 
-def validate_label_config(config_string):
+def validate_label_config(config_string: str | None) -> None:
     # xml and schema
     config_string = remove_xml_comments(config_string)
     try:
--- a/label_studio/core/label_config.py
+++ b/label_studio/core/label_config.py
@@ -101,7 +101,7 @@ def parse_config_to_json(config_string: Union[str, None]) -> Tuple[Union[Ordered
         raise etree.ParseError('xml is empty or incorrect')
     config = xmljson.badgerfish.data(xml)
     config = _fix_choices(config)
-    return config, etree.tostring(xml, encoding='unicode')
+    return config, etree.tostring(xml)
 
 
 def validate_label_config(config_string: Union[str, None]) -> None:
--- a/label_studio/core/label_config.py
+++ b/label_studio/core/label_config.py
@@ -4,11 +4,11 @@ import json
 import logging
 import re
 from collections import OrderedDict, defaultdict
-from typing import Union
+from typing import Tuple, Union
 from urllib.parse import urlencode
 
-import defusedxml.ElementTree as etree
 import jsonschema
+import lxml.etree as etree
 import numpy as np
 import pandas as pd
 import xmljson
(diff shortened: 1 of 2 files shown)


label_studio/core/label_config.py
```diff
@@ -76,7 +77,13 @@ def _fix_choices(config):
     return config
 
 
-def parse_config_to_json(config_string):
+def remove_xml_comments(config_string: Union[str, None]) -> Union[str, None]:
+    if config_string is None:
+        return None
+    return re.sub(r'<!--[\s\S]*?-->', '', config_string)
```
makseq: I am not sure regex is good here and copilot confirms it: 

"Using regex to remove comments from XML configuration files is not recommended because regex is not well-suited for parsing nested or complex structures like XML. Instead, you should use an XML parser that can accurately handle the structure and syntax of XML."

It suggests using this and I thought about the same: 

```suggestion
def remove_xml_comments(config_string: Union[str, None]) -> Union[str, None]:
    if config_string is None:
        return None
    parser = etree.XMLParser(remove_comments=True)
    xml = etree.fromstring(config_string, parser=parser)
    return etree.tostring(xml, encoding='unicode')
```

makseq (changes requested):

label_studio/core/label_config.py
bmartel: Good point, I wasn't aware this library had the means to remove comments like this, lxml variant normally has a keyword arg for it directly on fromstring which this lib did not.

label_studio/core/label_config.py
```diff
@@ -76,22 +77,37 @@ def _fix_choices(config):
...
-        xml = etree.fromstring(config_string, forbid_dtd=False)
+        xml = parse_config_to_xml(config_string, raise_on_empty=True)
     except TypeError:
         raise etree.ParseError('can only parse strings')
     if xml is None:
         raise etree.ParseError('xml is empty or incorrect')
     config = xmljson.badgerfish.data(xml)
     config = _fix_choices(config)
-    return config
+    return config, etree.tostring(xml, encoding='unicode')
```
makseq, line 104: why do we need to use unicode encoding here?

bmartel: This is required because the default of the lib is to output a bytestring which then fails all of our tests later when the string is used elsewhere. Providing the encoding properly decodes the string.

bmartel, line 104:
```suggestion
    return config, etree.tostring(xml)
```

label_studio/core/label_config.py
```diff
@@ -76,22 +77,37 @@ def _fix_choices(config):
...
-        xml = etree.fromstring(config_string, forbid_dtd=False)
+        xml = parse_config_to_xml(config_string, raise_on_empty=True)
     except TypeError:
         raise etree.ParseError('can only parse strings')
     if xml is None:
         raise etree.ParseError('xml is empty or incorrect')
     config = xmljson.badgerfish.data(xml)
     config = _fix_choices(config)
-    return config
+    return config, etree.tostring(xml)
```
bmartel: This is still required, otherwise the encoding used by default in the lib causes:

`cannot use a string pattern on a bytes-like object` as the value returned is a bytestring.

```suggestion
    return config, etree.tostring(xml, encoding='unicode')
```

makseq (approved):

AndrejOros (approved):

PR #6618, opened 2024-11-08
fix: LEAP-1641: Migration for fixing ProjectMember duplicates

triklozoid:
* in async job query all projects with duplicates, iterate through and remove duplicates to have only unique members


label_studio/projects/migrations/0028_auto_20241107_1031.py
```diff
@@ -0,0 +1,86 @@
...
+                entries_to_delete = (
+                    ProjectMember.objects
+                    .filter(user_id=user_id, project_id=project_id)
+                    .exclude(id=min_id)
+                    .only('id')  # Minimizing fields selected
+                )
+                deleted_count, _ = entries_to_delete.delete()
+                total_deleted += deleted_count
+            logger.info(f'Deleted {total_deleted} duplicate ProjectMember entries for project ID {project_id}.')
+
```
makseq, line 55: maybe we can make as atomic transaction? I will be easy to restart this process if it fails on some project

makseq (approved):

label_studio/projects/migrations/0028_auto_20241107_1031.py
triklozoid: I don't see a big difference with `atomic` because we have only "remove all duplicates for one member" updating query, if it would stop at any moment, we can restart it, it will select again all projects with duplicates and iterate though them again. I some project don't have duplicates anymore, it will not be includes into queryset.

makseq (approved):

label_studio/projects/migrations/0028_auto_20241107_1031.py
```diff
@@ -0,0 +1,86 @@
...
+from core.models import AsyncMigrationStatus
+from core.redis import start_job_async_or_sync
+
+logger = logging.getLogger(__name__)
+migration_name = '0028_auto_20241107_1031'
+
+
+def forward_migration(migration_name):
+    from django.db.models import Count, Min
```
jombooth: seems a bit strange to import these here? not sure also if the import below needs to be done when the function runs

label_studio/projects/migrations/0028_auto_20241107_1031.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+            total_deleted = 0
+            for dup in duplicates:
+                user_id = dup['user_id']
+                min_id = dup['min_id']
+                entries_to_delete = (
+                    ProjectMember.objects
+                    .filter(user_id=user_id, project_id=project_id)
+                    .exclude(id=min_id)
+                    .only('id')  # Minimizing fields selected
```
jombooth: I don't think you need to do this. If you don't iterate over the queryset I don't think any fields will be loaded.

jombooth (approved):

label_studio/projects/migrations/0028_auto_20241107_1031.py
triklozoid: Fixed. By the way, do you think this is a critical issue? I don't believe there's any significant difference in the migration run result.

label_studio/projects/migrations/0028_auto_20241107_1031.py
triklozoid: Fixed.

triklozoid:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11838674948)
> Successfully merged:  create mode 100644 web/apps/labelstudio/src/components/HeidiTips/liveContent.json

PR #6623, opened 2024-11-08
feat: OPTIC-1245: Allow HeidiTips to pull content updates live

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

To ensure HeidiTips can be up to date with the code over time we are going to host the contents in a raw json file within the repository similar to docs. That way just like docs we can align and amend without users having to update their LabelStudio version.

Video showcasing the loading of live content from Github file, using hard coded prefixes in all tip titles to show that the content is swapping upon fetching in subsequent tooltip usages:

[Kapture 2024-11-13 at 09.17.05.webm](https://github.com/user-attachments/assets/bd210378-84fa-4a21-b9eb-6c2c3310d57c)

HeidiTips


label_studio/core/views.py
```diff
@@ -169,6 +170,21 @@ def samples_paragraphs(request):
     return HttpResponse(json.dumps(result), content_type='application/json')
 
 
+def heidi_tips(request):
```
yyassi-heartex, line 173: any way we can make this optional? I'm concerned if an outage happens this will block pytest checks

yyassi-heartex (approved):

label_studio/core/views.py
bmartel: Great point, I'm going to look at that because honestly we don't want this in any pytests as there is nothing to test but everything to possibly fail randomly which is not helpful.

bmartel: At the very least we'd want to mock the requests call.

label_studio/core/views.py
```diff
@@ -169,6 +170,21 @@ def samples_paragraphs(request):
...
+    """Fetch live tips from github raw liveContent.json to avoid caching and client side CORS issues"""
+    url = 'https://raw.githubusercontent.com/HumanSignal/label-studio/refs/heads/develop/web/apps/labelstudio/src/components/HeidiTips/liveContent.json'
```
hlomzik, line 175: are we sure that this domain have wide open CORS policy? asking just in case

label_studio/core/views.py
```diff
@@ -169,6 +170,21 @@ def samples_paragraphs(request):
...
+    response = requests.get(
+        url,
+        headers={'Cache-Control': 'no-cache', 'Content-Type': 'application/json', 'Accept': 'application/json'},
+        timeout=5,
+    )
+
+    # Raise an exception for bad status codes to avoid caching
+    response.raise_for_status()
```
hlomzik: what would happen if we can't load tips? this should not be a blocker for the app

bmartel: In that case it just uses whatever loaded most recent in localStorage, and if never loaded from the endpoint it goes to the default which is what we were already using.

bmartel: We won't block for errors, or even timeouts. I put a strict timeout on the call as I didn't want this loading slow or showing nothing ever. I wanted it to just get and update the data as naturally as possible where the user doesn't notice it's happening at all. As well, for airgapped environments we cannot make that call anyways so it will fail fast and the defaults are all that will be provided.

label_studio/core/views.py
bmartel: There is no CORS in backend code, CORS is a browser security policy to protect the user. This is why it must be proxied from the backend, otherwise this would only fail CORS in the browser all the time as GH does not set the proper headers to allow the call to succeed directly from the browser.

web/apps/labelstudio/src/components/HeidiTips/utils.ts
```diff
@@ -1,16 +1,74 @@
...
+  const fetchedAt = localStorage.getItem(CACHE_FETCHED_AT_KEY);
+
+  // Read from local storage if the collection is less than CACHE_STALE_TIME milliseconds old
+  if (cachedData && fetchedAt && Date.now() - Number.parseInt(fetchedAt) < CACHE_STALE_TIME) {
+    return JSON.parse(cachedData);
+  }
+
+  const abortController = new AbortController();
+
+  // Abort the request after MAX_TIMEOUT milliseconds to ensure we don't block for too long, something might be wrong with the network
```
bmartel:
```suggestion
  // Abort the request after MAX_TIMEOUT milliseconds to ensure we won't wait for too long, something might be wrong with the network or it could be an air-gapped instance
```

web/apps/labelstudio/src/components/HeidiTips/utils.ts
```diff
@@ -1,16 +1,74 @@
...
   return `${STORE_KEY}:${collection}`;
 }
 
-export function getRandomTip(collection: keyof typeof TipsCollection): Tip | null {
+export const loadLiveTipsCollection = () => {
+  // stale while revalidate
+  const cachedData = localStorage.getItem(CACHE_KEY);
+  const fetchedAt = localStorage.getItem(CACHE_FETCHED_AT_KEY);
+
+  // Read from local storage if the collection is less than CACHE_STALE_TIME milliseconds old
```
bmartel:
```suggestion
  // Read from local storage if the cachedData is less than CACHE_STALE_TIME milliseconds old
```

web/apps/labelstudio/src/components/HeidiTips/utils.ts
```diff
@@ -1,16 +1,74 @@
...
+const CACHE_STALE_TIME = 1000 * 60 * 60; // 1 hour
+const MAX_TIMEOUT = 5000; // 5 seconds
 
 function getKey(collection: string) {
   return `${STORE_KEY}:${collection}`;
 }
 
-export function getRandomTip(collection: keyof typeof TipsCollection): Tip | null {
+export const loadLiveTipsCollection = () => {
+  // stale while revalidate
```
bmartel:
```suggestion
  // stale while revalidate - we will return the data present in the cache or the default data and fetch updated data to be put into the cache for the next time this function is called.
```

web/apps/labelstudio/src/components/HeidiTips/utils.ts
```diff
@@ -1,16 +1,74 @@
...
+const CACHE_STALE_TIME = 1000 * 60 * 60; // 1 hour
+const MAX_TIMEOUT = 5000; // 5 seconds
 
 function getKey(collection: string) {
   return `${STORE_KEY}:${collection}`;
 }
 
-export function getRandomTip(collection: keyof typeof TipsCollection): Tip | null {
+export const loadLiveTipsCollection = () => {
+  // stale while revalidate - we will return the data present in the cache or the default data and fetch updated data to be put into the cache for the next time this function is called.
```
bmartel:
```suggestion
  // stale while revalidate - we will return the data present in the cache or the default data and fetch updated data to be put into the cache for the next time this function is called without waiting for the promise.
```

label_studio/core/views.py
bmartel: Double checked it, we don't run these urls in tests, so this is not being executed in the tests currently. Also, the handling of this is such that there is a max timeout of 5 seconds and the call made for this file will also gracefully degrade for air-gapped environments with no internet access. So in all cases this will never block and either return the response code from the error thrown from fetching the file, or 404 in the case the file cannot be reached or there are network issues.

web/apps/labelstudio/src/components/HeidiTips/utils.ts
```diff
@@ -1,16 +1,72 @@
...
+    return JSON.parse(cachedData);
+  }
+
+  const abortController = new AbortController();
+
+  // Abort the request after MAX_TIMEOUT milliseconds to ensure we won't wait for too long, something might be wrong with the network or it could be an air-gapped instance
+  const abortTimeout = setTimeout(abortController.abort, MAX_TIMEOUT);
+
+  // Fetch from github raw liveContent.json proxied through the server
+  fetch("/heidi-tips/", {
```
bmartel:
```suggestion
  fetch("/heidi-tips", {
```

label_studio/core/views.py
```diff
@@ -169,6 +170,33 @@ def samples_paragraphs(request):
...
+    response = None
+    try:
+        response = requests.get(
+            url,
+            headers={'Cache-Control': 'no-cache', 'Content-Type': 'application/json', 'Accept': 'application/json'},
+            timeout=5,
+        )
+        # Raise an exception for bad status codes to avoid caching
+        response.raise_for_status()
+    # Catch all statuses and return either the status code if there was a response, or default to 404 if there is network issues
```
bmartel:
```suggestion
    # Catch all exceptions and return either the status code if there was a response, or default to 404 if there are network issues
```

wesleylima (approved):

AndrejOros (approved):

PR #6630, opened 2024-11-11
fix: LEAP-1615: LEAP-456: Fix tool unselect

hlomzik:
Pan was blocking the app upon unselection, because there were no tools selected after this. When drawing tools are not displayed (like in usual RectangleLabels case) then it's hard to start drawing again.
Also Selection tool (aka Move tool) was switched to the same behavior, allowing users to unselect it. The default tool will be selected then.
Also Smart tools selection is fixed — when you click on exact tool it's selected without weird jumps and rotation through other tools.

https://github.com/user-attachments/assets/a25ad3b2-a188-4304-b51f-49db9239f2ed

- [x] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

Users can be stuck after using Pan tool and unselecting it. Also that's inconsistent with Move tool.

nope

nope

Make Pan tool also not unselectable like Move tool. But for cases without drawing tools visible that would make it worse.

- [x] No


yyassi-heartex (approved):
approved! one point - isSeparated and iconComponent are both removed - are they no longer used anywhere?

as an aside, its been a long time since i touched anything editor and I dont remember off hand if we have tests specifically for the different tools and tags - if not would be great to see more tests for that in a seperate ticket

bmartel (approved):

web/libs/editor/src/tools/Manager.js
```diff
@@ -123,12 +123,12 @@ class ToolsManager {
...
     if (selected) {
       this.unselectAll();
-      if (tool.setSelected) tool.setSelected(true);
+      tool.setSelected?.(true);
     } else {
       const drawingTool = this.findDrawingTool();
 
       if (drawingTool) return this.selectTool(drawingTool, true);
-      if (tool.setSelected) tool.setSelected(false);
+      this.selectTool(this._default_tool, true);
```
bmartel: This could be cleaned up logically, at first glance I thought it was 2 different calls made but realized that it is just trying to select the most relevant tool. Really minor, just mentioning because the way it read made me double take.

MihajloHoma:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11814123871)
> Successfully merged:  create mode 100644 label_studio/tests/data_manager/actions/test_cache_labels.py

MihajloHoma (approved):

hlomzik:
> isSeparated and iconComponent are both removed - are they no longer used anywhere?

@yyassi-heartex they are used when we render base tool via `ToolView` in `tools/Base.jsx`. and I switched Move tool to render its custom tag like we do in some other tools, so we don't need these params

web/libs/editor/src/tools/Manager.js
hlomzik: thank you, improved this moment!

PR #6638, opened 2024-11-12
fix: LEAP-1650: pkg_resources deprecation problems for 3.12,3.13

jombooth:
pkg_resources was deprecated in python 3.12; replace usages with importlib and address imported packages that use pkg_resources as well. `boxing` has been replaced with the very similar `pyboxen`.


farioas (approved):

farioas:
Best change, been waiting for it for two years
LGTM 🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

PR #6662, opened 2024-11-15
fix: LEAP-1508: Prevent editing non-owned draft annotations

Gondragos:
Added `isNonEditableDraft` check to disable editing for drafts not created by the current user. Updated read-only conditions and related controls in the bottom bar to respect this new state.

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

There were some concerns, that user being an administrator (or an owner or a manager) can meet the situation when the other's person non commited draft (without related annotation yet) is visible and editable. Comment notifications imply this possibility.
In this case there shouldn't be an ability to edit and to submit annotation as well as accepting/rejecting them, as it will cause to breaking the data.

Not owned drafts are readonly.

You can change someone else draft and get error from the server on accepting/rejecting as there is no annotation yet.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

`Notifications`, `Quick View`, `Bottom Bar`


yyassi-heartex (approved):

web/libs/editor/src/stores/Annotation/Annotation.js
```diff
@@ -308,8 +308,22 @@ const _Annotation = types
       });
     },
 
+    get isNonEditableDraft() {
+      const isKnownUsers = !!self.user && !!self.store.user;
+      // If we do not know what user created draft
+      // and who we are, then, we shouldn't prevent the ability to edit annotation
+      // because we can't predict is it our draft or not.
+      // It most probably could be relevant for standalone `lsf`
+      if (!isKnownUsers) return false;
```
yyassi-heartex, line 317: while i approved this - im wondering if we should override this in LSE to not allow it if user is unknown

bmartel (approved):

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -108,6 +109,8 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
       ],
     );
 
+    if (isNonEditableDraft) return null;
```
hlomzik: given the lengthy function, simple access to the value and the fact it's used only here, I'd suggest to not use a var:
```suggestion
    if (annotation.isNonEditableDraft) return null;
```

web/libs/editor/src/stores/Annotation/Annotation.js
hlomzik: this condition is just in case because of possible cases of LSF usage out there. on LSE we should always have a user. if not that's the bug.

Gondragos: From my understanding in LSE we always should have an information about users. It's a part under our control.

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11895731963)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/Toast/Toast.tsx

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -60,6 +60,7 @@ const ControlButton = observer(({ button, disabled, onClick }: ControlButtonProp
 
 export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
   observer(({ store, history, annotation }) => {
+    const isNonEditableDraft = annotation.isNonEditableDraft;
```
hlomzik: I'm surprised biome is not triggered here...
```suggestion
```

hlomzik (approved):

MihajloHoma:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/11910868164)
> Successfully merged:  create mode 100644 docs/scripts/tableOfContent.js

PR #6682, opened 2024-11-19
fix: Text for import dialog about uploads

makseq:
We constantly face questions about direct uploads and this PR is addressed to fix them. 

This text will make users more attentive to downloading files directly and more active in using cloud storage.


caitlinwheeless (approved):

hlomzik (commented):
Please, describe, how and why did you change texts here

PR #6731, opened 2024-11-27
fix: DIA-1689: Fix empty parsed_label_config in projects

niklub:
Users see 500 errors when interacting with subsets API: some of the project.parsed_label_configs are empty

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -889,8 +889,12 @@ class Project(ProjectMixin, models.Model):
 
     def get_parsed_config(self, autosave_cache=True):
         if self.parsed_label_config is None:
-            self.parsed_label_config = parse_config(self.label_config)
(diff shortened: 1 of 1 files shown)


pakelley (approved):

label_studio/projects/models.py
```diff
@@ -890,9 +890,7 @@ def max_tasks_file_size():
     def get_parsed_config(self, autosave_cache=True):
```
bmartel: Since this appears to be having no effect right now with `autosave_cache` do we want to remove this parameter and remove the only False case used:

label_studio/projects/models.py|699 col 24| outputs = self.get_parsed_config(autosave_cache=False)

bmartel (approved):

label_studio/projects/models.py
niklub: done

PR #6735, opened 2024-11-28
fix: OPTIC-1407: Optimize the tasks API paginated prediction and annotation totals

bmartel:

_(check all that apply)_
- [ ] Product design
- [X] Backend (Database)
- [X] Backend (API)
- [ ] Frontend

At larger counts of tasks and annotations, these 2 queries would become incredibly inefficient and cause slowdowns. This PR instead uses the task column values to do aggregation and simple SUM's on the values to produce the same output.

Improves performance about 15-20x when looking at volumes of data where there are greater than 500K tasks, and 500K annotations per project.

- Moving the original subquery to a JOIN, which is difficult to do without dropping down to raw SQL.
- Similarily this query while an improvement could possibly benefit further from moving this subquery to a JOIN. Same point of why not, it would likely require dropping out of the ORM to achieve as it really likes to use subqueries (which do tend to be faster in most cases, just not at certain scale)

- `fflag_fix_back_optic_1407_optimize_tasks_api_pagination_counts`

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

Tasks List API

### DIFF
--- a/label_studio/data_manager/api.py
+++ b/label_studio/data_manager/api.py
@@ -19,6 +19,7 @@ from data_manager.serializers import (
     ViewSerializer,
 )
 from django.conf import settings
+from django.db.models import Sum
 from django.utils.decorators import method_decorator
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_yasg import openapi
@@ -223,9 +224,20 @@ class TaskPagination(PageNumberPagination):
         self.total_predictions = Prediction.objects.filter(task_id__in=queryset).count()
         self.total_annotations = Annotation.objects.filter(task_id__in=queryset, was_cancelled=False).count()
         return super().paginate_queryset(queryset, request, view)
+    
+    def paginate_totals_queryset(self, queryset, request, view=None):
+        totals = queryset.values('id', 'total_predictions', 'total_annotations', 'cancelled_annotations').aggregate(
+            total_annotations=Sum('total_annotations') - Sum('cancelled_annotations'),
+            total_predictions=Sum('total_predictions'),
+        )
+        self.total_annotations = totals['total_annotations']
+        self.total_predictions = totals['total_predictions']
+        return super().paginate_queryset(queryset, request, view)
 
     def paginate_queryset(self, queryset, request, view=None):
-        if flag_set('fflag_fix_back_leap_24_tasks_api_optimization_05092023_short'):
+        if flag_set('fflag_fix_back_optic_1407_optimize_tasks_api_pagination_counts'):
+            return self.paginate_totals_queryset(queryset, request, view)
+        elif flag_set('fflag_fix_back_leap_24_tasks_api_optimization_05092023_short'):
             return self.async_paginate_queryset(queryset, request, view)
         else:
             return self.sync_paginate_queryset(queryset, request, view)
--- a/label_studio/data_manager/api.py
+++ b/label_studio/data_manager/api.py
@@ -230,8 +230,8 @@ class TaskPagination(PageNumberPagination):
             total_annotations=Sum('total_annotations') - Sum('cancelled_annotations'),
             total_predictions=Sum('total_predictions'),
         )
-        self.total_annotations = totals['total_annotations']
-        self.total_predictions = totals['total_predictions']
+        self.total_annotations = totals.get('total_annotations', 0)
+        self.total_predictions = totals.get('total_predictions', 0)
         return super().paginate_queryset(queryset, request, view)
 
     def paginate_queryset(self, queryset, request, view=None):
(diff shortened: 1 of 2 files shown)


label_studio/data_manager/api.py
```diff
@@ -224,11 +225,23 @@ def sync_paginate_queryset(self, queryset, request, view=None):
         self.total_annotations = Annotation.objects.filter(task_id__in=queryset, was_cancelled=False).count()
         return super().paginate_queryset(queryset, request, view)
 
+    def paginate_totals_queryset(self, queryset, request, view=None):
+        totals = queryset.values('id', 'total_predictions', 'total_annotations', 'cancelled_annotations').aggregate(
+            total_annotations=Sum('total_annotations') - Sum('cancelled_annotations'),
```
triklozoid: It looks like `total_annotations` doesn't include `cancelled_annotations`, so we don't need to subtract them.

```
    total_annotations = models.IntegerField(
        _('total_annotations'),
        default=0,
        db_index=True,
        help_text='Number of total annotations for the current task except cancelled annotations',
    )
```

label_studio/data_manager/api.py
```diff
@@ -224,11 +225,23 @@ def sync_paginate_queryset(self, queryset, request, view=None):
...
+            total_predictions=Sum('total_predictions'),
+        )
+        # if the total is None, set it to 0
+        # the default value of the aggregate function is None
+        self.total_annotations = totals['total_annotations'] or 0
+        self.total_predictions = totals['total_predictions'] or 0
```
mcanu: [Coalesce](https://docs.djangoproject.com/en/5.1/ref/models/database-functions/#coalesce) can be used here instead to defer this to the db.

label_studio/data_manager/api.py
bmartel: total_annotations=Sum('total_annotations') - Sum('cancelled_annotations')

label_studio/data_manager/api.py
bmartel: Good point, though in this case it is not a heavy operation and is only a default of this query. It is solely the default of the final aggregate, not the default of each intermediary value.

label_studio/data_manager/api.py
```diff
@@ -224,11 +225,23 @@ def sync_paginate_queryset(self, queryset, request, view=None):
         self.total_annotations = Annotation.objects.filter(task_id__in=queryset, was_cancelled=False).count()
         return super().paginate_queryset(queryset, request, view)
 
+    def paginate_totals_queryset(self, queryset, request, view=None):
+        totals = queryset.values('id', 'total_predictions', 'total_annotations', 'cancelled_annotations').aggregate(
```
mcanu: Is `total_predictions`, `total_annotations` and `cancelled_annotations` needed in `values` if we are aggregating them afterwards?

label_studio/data_manager/api.py
mcanu: Agreed, just a nitpick

label_studio/data_manager/api.py
bmartel: If not present in values the SELECT was missing these fields and failed to aggregate, I can try it again though just in case I missed something.

label_studio/data_manager/api.py
bmartel: I saw that, but everywhere I saw it appeared to not populate with that notion. Also every single test we have passed with this change 🤔

bmartel: Nope you are right:

```python
def update_tasks_counters(queryset, from_scratch=True):
    """
    Update tasks counters for the passed queryset of Tasks
    :param queryset: Tasks to update queryset
    :param from_scratch: Skip calculated tasks
    :return: Count of updated tasks
    """
    objs = []

    total_annotations = Count('annotations', distinct=True, filter=Q(annotations__was_cancelled=False))
    cancelled_annotations = Count('annotations', distinct=True, filter=Q(annotations__was_cancelled=True))
    total_predictions = Count('predictions', distinct=True)
    ```

Comment (triklozoid) [reply label_studio/data_manager/api.py]:
```
>>> queryset = project.tasks.all()
>>> totals = queryset.values('id', 'total_predictions', 'total_annotations', 'cancelled_annotations').aggregate(
... total_annotations=Sum('total_annotations') - Sum('cancelled_annotations'),
... total_predictions=Sum('total_predictions'),
... )
>>> totals['total_annotations']
1
>>> project.annotations.filter(was_cancelled=False).count()
2
```

bmartel: Going to amend this 👍

label_studio/data_manager/api.py
bmartel: Took it out, only need the id for the `SELECT DISTINCT`

triklozoid (approved):

mcanu (approved):

AndrejOros (approved):

PR #6780, opened 2024-12-11
fix: OPTIC-1418: Improve Organization Membership API usage to reduce latency

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [X] Frontend

Upstream PR has the details, but this allows for external queryCache to act on the calls from DataManager.

#### What does this fix?
_(if this is a bug fix)_

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


web/libs/datamanager/src/utils/api-proxy/index.js
```diff
@@ -268,6 +287,10 @@ export class APIProxy {
         writable: false,
       });
 
+      if (shouldUseQueryCache) {
+        methodSettings.queryCache(finalParams, responseResult);
+      }
```
yyassi-heartex, line 292: this is fine for now but down the line if we need to add soemthing to this we should turn this into a function and use it in both spots we have this copied in

yyassi-heartex (approved):
minor note but looks good

web/libs/datamanager/src/utils/api-proxy/index.js
bmartel: Not sure I quite follow, it is a function currently, triggered only when present and special query params are present (which get removed before sending).

yyassi-heartex: i meant including the if statement - but i maybe overthinking it

bmartel: ohh I see, well at this point I don't see it as an issue mostly because the only duplication is the if statement being based on the single boolean. I wasn't at this time looking to abstract that any further as it felt premature. I think the part that has me feeling this way is the fact that I think if this works well enough, there is a future where a lot of the code in these api proxies gets refactored to a single api proxy in core, and the functions that power it would all be refactored at that time.

yyassi-heartex: yeah makes sense - i was thinking down the line if the need arose where we need to add more stuff with in the if statement but we could just as easily put it in the function with in - no biggie

AndrejOros (approved):

PR #6782, opened 2024-12-12
fix: LEAP-1409: Add preserving hidden filters during filter operations

Gondragos:
Ensure hidden filters are retained and correctly processed in filter-related operations. Introduced `_originalFilters` to store and manage filters not editable by the user, preventing their loss during requests.

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

We have columns in Data Manager. For some user roles the list of columns is reduced. We can have filters over these columns set by other users. But we cannot do anything with these filters as they are dependent on  the non existing for the current user columns.

The previous solution was just filter this filters out of the models. But this approach causes the problem: we do not save this filtered out filters when we adding / deleting / editing other filters an columns.

Storing these filtered out filters somewhere and restoring them on creating a request should help with that.

1. Solve this on the backend. But we may lose consistency and orderliness. There is not enough control.
2. We could keep these filters in the model but just hide them. In that case we have problem with fixing current logic, that was written with assuming that we do not need these filters and everything else, related to them (f.e. colums in the model). It's hard to fixx all places like that. 

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

`Data Manager`, 'Filers'

### DIFF
--- a/web/libs/datamanager/src/stores/Tabs/store.js
+++ b/web/libs/datamanager/src/stores/Tabs/store.js
@@ -27,7 +27,10 @@ const dataCleanup = (tab, columnIds) => {
   if (!data) return { ...tab };
 
   if (data.filters) {
-    data.filters.items = data.filters.items.filter(({ filter }) => {
+    data._originalFilters = [];
+    data.filters.items = data.filters.items.filter((item) => {
+      const { filter } = item;
+      data._originalFilters.push(item);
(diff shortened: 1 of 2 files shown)


robot-ci-heartex:

PR #6856, opened 2025-01-06
fix: OPTIC-1490: Safely reference content_type of the response so it does not error

bmartel:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [X] Backend (API)
- [ ] Frontend

There will be cases where the response has no content_type, which is the case of NotFound response objects. This just makes sure it is present before reading it.

ContextLog

### DIFF
--- a/label_studio/core/utils/contextlog.py
+++ b/label_studio/core/utils/contextlog.py
@@ -271,7 +271,7 @@ class ContextLog(object):
         else:
             namespace = request.resolver_match.namespace if request.resolver_match else None
             status_code = response.status_code
-            content_type = response.content_type if 'content_type' in response else None
+            content_type = response.content_type if hasattr(response, 'content_type') else None
             response_content = self._get_response_content(response)
 
         payload = {
--- a/label_studio/core/utils/contextlog.py
+++ b/label_studio/core/utils/contextlog.py
@@ -271,7 +271,7 @@ class ContextLog(object):
         else:
             namespace = request.resolver_match.namespace if request.resolver_match else None
             status_code = response.status_code
-            content_type = response.content_type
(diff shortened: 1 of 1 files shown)


label_studio/core/utils/contextlog.py
```diff
@@ -271,7 +271,7 @@ def create_payload(self, request, response, body):
         else:
             namespace = request.resolver_match.namespace if request.resolver_match else None
             status_code = response.status_code
-            content_type = response.content_type
+            content_type = response.content_type if 'content_type' in response else None
```
makseq: does response support this check `'content_type' in response`? 

I would do something like this: 
```suggestion
            content_type = response.content_type if hasattr(response, 'content_type') else None
```

makseq (approved):

label_studio/core/utils/contextlog.py
bmartel: Good point, I am not sure what methods the base classes have, better to use a more catch-all solution

makseq (approved):

label_studio/core/utils/contextlog.py
```diff
@@ -271,7 +271,7 @@ def create_payload(self, request, response, body):
         else:
             namespace = request.resolver_match.namespace if request.resolver_match else None
             status_code = response.status_code
-            content_type = response.content_type
+            content_type = response.content_type if hasattr(response, 'content_type') else None
```
mcanu: nitpick (you can dismiss this):

```suggestion
            content_type = getattr(response, 'content_type', None)
```

mcanu (approved):

AndrejOros (approved):

PR #6867, opened 2025-01-08
fix: LEAP-1731: Bulk annotation follow-ups

hlomzik:
UX fixes:
- continue to pan when cursor is out of the image
- don't close modal if user released cursor out of the image
- `user-select: none` to always prevent from selection
- info about zoom and pan to tooltip

Image Preview is rewritten to be a class component, it's more readable and fixes some issues.

All modals in DM now fixes the window to be non-scrollable — that's usual behaviour for modals and we just missed this.

- [x] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend

`fflag_feat_front_leap_1424_grid_preview_short`

- [ ] No

All modals in DM are affected, but the fix is very expected.

- [ ] e2e
- [ ] integration
- [ ] unit


bmartel:
> it's more readable and fixes some issues.

Could you elaborate more about what issues it fixed?

The problem I have with introducing class components is that React going forward is moving further and further from this, and the mindshare is already entirely shifted into hooks being the canonical way to develop with React. As a result this means all main libraries are hooks only, including state management approaches. I would not want to have to be maintaining 2 methods of developing components, so it would have to be a definitive reason to consciously move towards.

As for the more readable, that is a subjective take, as I actually see the opposite. What I do agree with is that there are far too many cases of components which happened to use hooks incorrectly which clouds things quite a bit. That is why I framed it more of the mindset that we need to tighten the belt on component best practices, including (must include this) guidance on state management from top to bottom so we don't continue to perpetuate what I consider `useEffect` / `useState` hell. No one can reason about that type of state mismanagement, but I don't see that as a React hooks problem, more just a miss entirely on needed refactors when components graduate from a couple pieces of state with a simple interaction, to something greater.

web/libs/datamanager/src/components/MainView/GridView/ImagePreview.tsx
```diff
@@ -81,9 +86,8 @@ const ImagePreview = observer(({ task, field }: ImagePreviewProps) => {
       const scaledWidth = img.naturalWidth * imageScale;
       const scaledHeight = img.naturalHeight * imageScale;
       // how much should we zoom image in to cover container
-      const coverScale = Math.max(containerRect.width / scaledWidth, containerRect.height / scaledHeight);
+      // const coverScale = Math.max(containerRect.width / scaledWidth, containerRect.height / scaledHeight);
```
yyassi-heartex, line 89: are you looking to put this back at some point? or should this be removed?
```suggestion
```

web/libs/datamanager/src/components/MainView/GridView/ImagePreview.tsx
```diff
@@ -96,62 +100,84 @@ const ImagePreview = observer(({ task, field }: ImagePreviewProps) => {
...
-    setStartOffset({ x: offset.x, y: offset.y });
-  };
+      const { x: oldX, y: oldY } = dragParams.current.dragAnchor;
+      const { x: offsetX, y: offsetY } = dragParams.current.startOffset;
+      const newX = e.clientX - oldX;
+      const newY = e.clientY - oldY;
 
-  const handleMouseMove = (e: React.MouseEvent) => {
-    if (!isDragging || !containerRef.current || !imageRef.current) return;
+      setOffset(constrainOffset({ x: offsetX + newX, y: offsetY + newY }));
```
yyassi-heartex, line 147: I see the logic is changing here - and it seems reasonable but feels like we should have a test added to make sure the behaviour is consistent with expectations

yyassi-heartex: oh i see - i think the change in order for the functions threw me off a little but it looks like a refactor - not a logic change - disregard by previous comment

bmartel (approved):

yyassi-heartex (approved):
looks good!

web/libs/datamanager/src/components/MainView/GridView/ImagePreview.tsx
hlomzik: Yeah, this might be helpful, so I'll leave it here for now

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/12759100102)
> Successfully merged:  create mode 100644 web/libs/editor/src/components/InteractiveOverlays/RelationsOverlay.module.scss

MihajloHoma:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/12762819542)
> Successfully merged:  create mode 100644 docs/themes/v2/source/images/project/video_sync.png

Gondragos:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/12791922964)
> Successfully merged:  create mode 100644 label_studio/core/static/templates/medical-image-classification.png

robot-ci-heartex:
Follow Merge downstream workflow has been failed.
> [Workflow run](https://github.com/HumanSignal/label-studio-enterprise/actions/runs/12791962181)

PR #6878, opened 2025-01-08
refactor: OPTIC-1506: Remove Stale Feature Flag - fflag_fix_front_dev_3350_restrict_drawing_area_short

luarmr:
Removing a long time stale feature flag, 
value true.


web/libs/editor/src/core/settings/videosettings.ts
```diff
@@ -1,12 +1,10 @@
-import { FF_DEV_3350 } from "../../utils/feature-flags";
 import type { SettingsProperties } from "./types";
 
 export default {
   videoDrawOutside: {
     description: "Allow drawing outside of video boundaries",
     defaultValue: false,
     type: "boolean",
-    ff: FF_DEV_3350,
```
luarmr, line 9: I can't find were ff property is in use, can you @bmartel help me with that?

bmartel: https://github.com/HumanSignal/label-studio/blob/0f5a47873a24080f74e3edf50cf3cbee96ef98a0/web/libs/editor/src/components/Settings/TagSettings/SettingsRenderer.tsx#L13

bmartel: And passed here: https://github.com/HumanSignal/label-studio/blob/2660ec9857597f35130ce0e1e7e2ccde225cda4f/web/libs/editor/src/components/Settings/TagSettings/VideoSettings.tsx#L7

luarmr: I was tented to leave it like that, but I think is better if we:
```
      {Object.entries(settings).map(([key, value]) => (
        <SettingsField key={key} name={key} store={store} value={value} />
      ))}
```

yyassi-heartex: ahh i was confused about the changes in web/libs/editor/src/components/Settings/TagSettings/SettingsRenderer.tsx until i saw this comment thread lol very nice 👍

yyassi-heartex (approved):

bmartel (approved):

web/libs/editor/src/core/settings/videosettings.ts
luarmr: I am reverting this change for a comment from Andrew in slack

Issue #6567, opened 2024-10-29
yarn dev: Error: Must provide a proper URL as target     at ProxyServer.<anonymous>


lincolnneu:
**Describe the bug**
In `yarn dev` mode, opening `http://localhost:8010/` run into 

```
Error: Must provide a proper URL as target
    at ProxyServer.<anonymous> (/label-studio/web/node_modules/http-proxy/lib/http-proxy/index.js:69:35)
    at HttpProxyMiddleware.middleware (/label-studio/web/node_modules/http-proxy-middleware/dist/http-proxy-middleware.js:22:32)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

**To Reproduce**
1. Install for local development

Download latest dev branch. Following the READM.md in under `web` folder https://github.com/HumanSignal/label-studio/blob/0f5a47873a24080f74e3edf50cf3cbee96ef98a0/web/README.md#installation-instructions
```
yarn install --frozen-lockfile
```
create the .env (with the same content as .env.development
From the `web` directory: Run `yarn dev` and wait till 

```
webpack compiled successfully
```

2. Go to http://localhost:8010
3. The yarn session will show the error msg reported above, and the browser also shows the same error msg.

**Expected behavior**
yarn dev should result in working frontend session, and `http://localhost:8010` should show the label-studio UI

**Environment (please complete the following information):**
 - Ubuntu 22.04.1 LTS
 - Browser: Chrome
 - Label Studio Version origin/develop e4c2721131e1d4b62319b42410d0a8c06ec533fd

heidi-humansignal:
Hello,

The error you're encountering indicates that the frontend development server's proxy configuration lacks a proper target URL. This usually happens when the backend server is not running or the proxy settings are incorrect.
To resolve this issue, please follow these steps:

### **Step 1: Start the Backend Server**

1. Navigate to the root directory of Label Studio where `manage.py` is located.
2. Start the backend server by running:

   ` python label_studio/manage.py runserver`

By default, this will start the server at `http://localhost:8080`.

### **Step 2: Verify the Backend Server is Running**

- Open your browser and navigate to `http://localhost:8080` to confirm that the backend server is running.

### **Step 3: Check Frontend Proxy Configuration**

1. In the `web` directory, ensure you have a `.env` file. If not, create one and copy the contents from `.env.development`.
2. Open the `.env` file and verify that the `VITE_BACKEND_HOST` variable is set correctly:

[...]

PR #6880, opened 2025-01-09
feat: OPTIC-1553: Add URL-based region visibility, hiding all but the specified region on load

ddishi:

_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [x] Frontend
 

Couldn't provide a URL to a specific region while maintaining context.

#### What does this fix?
_(if this is a bug fix)_

The ability to read URL parameters in order open the task\annotation with only the specified region presented while all other regions are hidden.
image

`fflag_feat_front_optic_1553_url_based_region_visibility_short `

_(check only one)_
- [x] No

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit


ddishi:
Hi @makseq,
Just following up as you suggested in the Slack channel. Could you please bump this PR to the relevant team?
Also, I ran into a merge conflict regarding the proxy config, where the version in main didn’t seem to work for me. Please see https://github.com/HumanSignal/label-studio/issues/6567, Any guidance there would be great.

Thanks! 🙏

makseq:
@ddishi could you please rename your feature flag to this one:
`fflag_feat_front_optic_1553_url_based_region_visibility_short`
?

bmartel:
@ddishi Would you be able to add some tests for this change? Test case(s) for this can be added here: https://github.com/HumanSignal/label-studio/blob/2660ec9857597f35130ce0e1e7e2ccde225cda4f/web/libs/editor/tests/integration/e2e/outliner/hide-all.cy.ts#L46-L47

bmartel:
Thanks a bunch @ddishi 🙏

web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -199,7 +199,22 @@ export const AppStore = types
...
     setTask: flow(function* ({ taskID, annotationID, pushState }) {
       if (pushState !== false) {
-        History.navigate({ task: taskID, annotation: annotationID ?? null, interaction: null });
+        History.navigate({
+          task: taskID,
+          annotation: annotationID ?? null,
+          interaction: null,
+          region: null,
+        });
+      } else {
```
bmartel:
```suggestion
      } else if (isFF(FF_REGION_VISIBILITY_FROM_URL)) {
```

web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -214,22 +229,39 @@ export const AppStore = types
 
       if (annotationID !== undefined) {
         self.annotationStore.setSelected(annotationID);
-      } else {
-        self.taskStore.setSelected(taskID);
+      }
```
bmartel, line 234: Can you explain why the else case was removed?

ddishi: Because we need to load the task even if an `annotationID` is provided for the FF to work properly. We need to be able to choose both `annotationID` and `taskID`. 
Also we always have `taskID` defined since we return otherwise.

bmartel: But in the case of a defined `annotationID` we already have the task loaded and selected. We don't need to reload the task to switch annotations, that would incur an overhead which seems unnecessary.

benglewis: @bmartel I believe that this is specifically not the case with the URL argument. Maybe there is some kind of race condition or something. We'd appreciate your input

bmartel: I'll have to look closer, but the reason I flagged this is because there are two cases that were originally handled in this function.

1. Switching tasks
2. Switching annotations within a task

The removal of the else case means that in the case we are only switching an annotation, we now would be always loading the task from the API even though it has not changed. There seems like there should be a way we can position this to avoid introducing an unnecessary API call for a task we already have.

@hlomzik does this make sense?

ddishi: We checked it again and the feature seems to be working with the else block after all... I'll revert it. 
Thanks for pointing that up 🙏🏼

web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -216,20 +231,38 @@ export const AppStore = types
...
+            });
+
+            if (lsfAnnotation) {
+              const annID = lsfAnnotation.pk ?? lsfAnnotation.id;
+              self.LSF?.setLSFTask(self.taskStore.selected, annID);
+            }
+          }
+          if (regionIDFromUrl) {
+            const currentAnn = self.LSF?.currentAnnotation;
+            currentAnn?.regionStore?.setRegionVisible(regionIDFromUrl);
```
bmartel, line 261: @ddishi There was one more thing I wanted to ask, which was whether it would be ok if we in addition to setting the region visible, actually selected the region as well. The reason being is that while visibility focuses a user, selection gives the context and information directly in the outliner panel. I discussed this internally with my product team, they see further usability of this concept within the broader community. Would that work with your current workflows?

bmartel: @ddishi I went ahead and implemented it, everything appears to work well in what I was expecting. Can you please confirm this on your end? I'm prepared to get this to our QA team today, putting this to merge by early next week.

ddishi: @bmartel Everything looks good on our end—thanks for the cooperation!

ddishi: @bmartel I also added a small test case for the selection method. Let me know if anything needs adjustment.

bmartel: Thank you @ddishi! This looks great!

bmartel (approved):

bmartel:
@ddishi This change was reviewed internally and accepted and is now merged back into the mainline branch! Just wanted to thank you on behalf of the Label Studio team for your contribution 🙏.

benglewis:
@bmartel Thank you so much for merging this PR :)
Do you have any ideas when the plan is for the next release of Label Studio?

I ask since we would prefer to use an official release and to keep this feature from the mainline than to build our own custom build to get the feature early, but we need to understand how long that timeline is to be able to make that decision

bmartel:
@benglewis Absolutely! On calendar we are targeting the next Label Studio release for end of March.

PR #6882, opened 2025-01-09
refactor: OPTIC-1510: remove a batch of ff

luarmr:
Review commit by commit
A group of simple FF that were stale an we dice to remove them in the same pull request to avoid the conflict when merge.

`fflag_fix_front_dev_4075_taxonomy_overlap_281222_short` as  True,

`fflag_fix_back_lsdv_1044_check_annotations_24012023_short` as False,

`fflag_fix_font_lsdv_1148_hotkeys_namespaces_01022023_short` as True,

`fflag_feat_front_lsdv_4659_skipduplicates_060323_short` as True
 (skipduplicates_editing is referenced somehow here, so I take it down also in the last commit)

`fflag_feat_front_lsdv_4661_full_uri_resolve_15032023_short` as  True
(white spaces here are relevant)

`fflag_feat_front_lsdv_4712_skipduplicates_editing_110423_short` as True

All of this are LSO only.

You may want to review commit by commit and with whitespace option enabled.
First commit is empty since I create a ticket to group this work

### DIFF
--- a/web/libs/editor/src/components/Taxonomy/Taxonomy.tsx
+++ b/web/libs/editor/src/components/Taxonomy/Taxonomy.tsx
@@ -5,7 +5,6 @@ import { LsChevron } from "../../assets/icons";
 import { Tooltip } from "../../common/Tooltip/Tooltip";
 import { useToggle } from "../../hooks/useToggle";
 import type { CNTagName } from "../../utils/bem";
-import { FF_DEV_4075, isFF } from "../../utils/feature-flags";
 import { isArraysEqual } from "../../utils/utilities";
 import TreeStructure from "../TreeStructure/TreeStructure";
 
@@ -237,13 +236,9 @@ const Item: React.FC<RowProps> = ({ style, item, dimensionCallback, maxWidth, is
     <div ref={itemContainer} style={{ paddingLeft: padding, maxWidth, ...style, width: "fit-content" }}>
       {!isAddingItem ? (
         <>
-          <div
-            className={[styles.taxonomy__measure, isFF(FF_DEV_4075) ? styles.taxonomy__measure_ff_dev4075 : false]
-              .filter(Boolean)
-              .join(" ")}
-          >
+          <div className={styles.taxonomy__measure}>
             <label>{name}</label>
-            {isFF(FF_DEV_4075) && !isFiltering && (
+            {!isFiltering && (
               <div className={styles.taxonomy__extra}>
                 <span className={styles.taxonomy__extra_count}>{childCount}</span>
               </div>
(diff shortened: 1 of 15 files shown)


web/libs/editor/src/utils/feature-flags.ts
```diff
@@ -116,7 +110,7 @@ export const FF_LSDV_4583_6 = "fflag_feat_front_lsdv_4583_6_images_preloading_sh
  * Removing interrupting from the draft saving
  *
  * Without this flag we have a situation when changes in history leading to the empty results break functionality of adding comments and make the draft saving process indicator stay forever.
- * @link https://app.launchdarkly.com/default/production/features/fflag_fix_font_lsdv_1148_hotkeys_namespaces_01022023_short
+ * @link https://app.launchdarkly.com/default/production/features/fflag_fix_font_lsdv_3009_draft_saving_stuck_130223_short
  */
```
luarmr, line 107: Looks like a copy paste here :)

web/libs/editor/src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -237,13 +236,9 @@ const Item: React.FC<RowProps> = ({ style, item, dimensionCallback, maxWidth, is
     <div ref={itemContainer} style={{ paddingLeft: padding, maxWidth, ...style, width: "fit-content" }}>
       {!isAddingItem ? (
         <>
-          <div
-            className={[styles.taxonomy__measure, isFF(FF_DEV_4075) ? styles.taxonomy__measure_ff_dev4075 : false]
-              .filter(Boolean)
-              .join(" ")}
-          >
+          <div className={[styles.taxonomy__measure, styles.taxonomy__measure_ff_dev4075].filter(Boolean).join(" ")}>
```
bmartel: `taxonomy__measure_ff_dev4075`

This seems like it should also get removed, and the css updated.

luarmr: **Great catch**, I update the commit so the change of the feature flag still atomic.

web/libs/editor/src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -237,13 +236,9 @@ const Item: React.FC<RowProps> = ({ style, item, dimensionCallback, maxWidth, is
     <div ref={itemContainer} style={{ paddingLeft: padding, maxWidth, ...style, width: "fit-content" }}>
       {!isAddingItem ? (
         <>
-          <div
-            className={[styles.taxonomy__measure, isFF(FF_DEV_4075) ? styles.taxonomy__measure_ff_dev4075 : false]
-              .filter(Boolean)
-              .join(" ")}
-          >
+          <div className={styles.taxonomy__measure}>
```
luarmr, line 239: @bmartel here is the implementation of your great catch!

bmartel (approved):

hlomzik (approved):

PR #6918, opened 2025-01-16
feat: DIA-1815: Add conditional dependencies in product tour

niklub:
Ability to trigger product tours only if other product tours are completed using:
```yaml
dependencies:
  - another_tour

steps:
  ...
```

### DIFF
--- a/label_studio/users/product_tours/serializers.py
+++ b/label_studio/users/product_tours/serializers.py
@@ -1,10 +1,14 @@
+import logging
 import pathlib
-import yaml
 from functools import cached_property
+
+import yaml
 from core.utils.db import fast_first
 from rest_framework import serializers
 
-from .models import ProductTourInteractionData, UserProductTour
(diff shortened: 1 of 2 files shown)


label_studio/users/product_tours/serializers.py
```diff
@@ -29,14 +36,25 @@ def validate_name(self, value):
 
         return value
 
+    @cached_property
```
hakan458, line 39: what if we decide to change the config?

niklub: It'd go with redeployment and reinstantiating these objects, no?

matt-bernstein: True, but I think this isn't necessarily a great use case for cached_property anyway, loading this yaml file is cheap and doesn't happen often

matt-bernstein (approved):

hakan458 (approved):

yyassi-heartex (approved):

PR #6919, opened 2025-01-16
feat: LEAP-1774: exclude starter cloud accounts from bulk labeling feature

jombooth:
Basically just check isSelfServe wherever the bulk annotation FF is checked


yyassi-heartex (approved):

web/libs/editor/src/utils/billing.ts
```diff
@@ -0,0 +1,3 @@
+import { isFF, FF_SELF_SERVE } from "./feature-flags";
```
yyassi-heartex, line 1: this may be better positioned in the `core` library undersomething like `web/libs/core/src/lib/utils/billing.ts`
this way it can be reused in LSO and LSE with out redefining the function in both as i see the definition in LSE already exists but not in LSO

yyassi-heartex: this is a minor nitpick - i wouldnt hold up the PR over this

yyassi-heartex (approved):

web/libs/editor/src/utils/billing.ts
jombooth: I'll attempt this refactor in a follow up! Agree it would be better but I verified the code pretty carefully as is, so would rather push this to a future diff :)

bmartel (approved):

PR #6975, opened 2025-01-27
fix: DIA-1783: Fixing UI regressions

nick-skriabin:
Fixing various regressions caused by the Tailwind integration.


yyassi-heartex (approved):

web/apps/labelstudio/src/pages/CreateProject/Import/Import.jsx
```diff
@@ -73,7 +75,7 @@ function getFiles(files) {
 
 const Footer = () => {
   return (
-    <Modal.Footer>
+    <Modal.Footer style={{ display: "flex", alignItems: "center", justifyContent: "center", gap: "0.5rem" }}>
```
bmartel:
```
    <Modal.Footer style={{ display: "flex", alignItems: "center", justifyContent: "center", gap: "0.5rem" }}>
```
Why not encode this into the Footer styles directly?

bmartel: Or if required to be different here in particular, just use tailwind classes.

yyassi-heartex: yes i suspect this is a specific situation but i do like the tailwind class approach, +1 to this suggestion

web/apps/labelstudio/src/pages/CreateProject/Import/Import.scss
```diff
@@ -253,6 +252,13 @@
   margin-bottom: 24px;
 }
 
+.import-footer-ls {
```
bmartel:
```suggestion
.import-footer {
```
Won't BEM postfix `-ls` by default?

bmartel (approved):

web/apps/labelstudio/src/pages/CreateProject/Import/Import.scss
yyassi-heartex: no BEM doesnt apply the suffix automatically. only the prefix

borisheartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13006276713)
> Successfully merged:  delete mode 100644 docs/themes/v2/source/images/prompts/example_subset2.png

nick-skriabin:
All work is moved to #6989

PR #6885, opened 2025-01-09
feat: DIA-1794: [FE] Implement home page in LSO and LSE

nick-skriabin:
### PR fulfills these requirements
- [ ] Commit message(s) and PR title follows the format `[fix|feat|ci|chore|doc]: TICKET-ID: Short description of change made` ex. `fix: DEV-XXXX: Removed inconsistent code usage causing intermittent errors`
- [ ] Tests for the changes have been added/updated (for bug fixes/features)
- [ ] Docs have been added/updated (for bug fixes/features)
- [ ] Best efforts were made to ensure docs/code are concise and coherent (checked for spelling/grammatical errors, commented out code, debug logs etc.)
- [ ] Self-reviewed and ran all changes on a local instance (for bug fixes/features)

#### Change has impacts in these area(s)
_(check all that apply)_
- [ ] Product design
- [ ] Backend (Database)
- [ ] Backend (API)
- [ ] Frontend

### Describe the reason for change
_(link to issue, supportive screenshots etc.)_

#### What does this fix?
_(if this is a bug fix)_

#### What is the new behavior?
_(if this is a breaking or feature change)_

#### What is the current behavior?
_(if this is a breaking or feature change)_

#### What libraries were added/updated?
_(list all with version changes)_

#### Does this change affect performance?
_(if so describe the impacts positive or negative)_

#### Does this change affect security?
_(if so describe the impacts positive or negative)_

#### What alternative approaches were there?
_(briefly list any if applicable)_

#### What feature flags were used to cover this change?
_(briefly list any if applicable)_

### Does this PR introduce a breaking change?
_(check only one)_
- [ ] Yes, and covered entirely by feature flag(s)
- [ ] Yes, and covered partially by feature flag(s)
- [ ] No
- [ ] Not sure (briefly explain the situation below)

### What level of testing was included in the change?
_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### Which logical domain(s) does this change affect?
_(for bug fixes/features, be as precise as possible. ex. Authentication, Annotation History, Review Stream etc.)_

PR #7005, opened 2025-01-31
feat: DIA-1794: Revert "[FE] Implement home page in LSO and LSE"

nehalecky:
Reverts HumanSignal/label-studio#6885


luarmr:
we need to change the branch name

PR #7106, opened 2025-02-18
fix: DIA-1916: [FE] User is NOT able to upload profile image on new account page

nick-skriabin:
Refactoring the user account page to be more concise. Fixing avatar upload in LSO and LSE.

Along with that introducing new `useCurrentUser` hook that should be preferred over the existing one. New hook uses Jotai and tanstack-query under the hood and allows better cacheing without usage of Context API. This hook is a drop-in replacement for the old one.

### DIFF
--- a/web/libs/core/src/lib/hooks/useCurrentUser.ts
+++ b/web/libs/core/src/lib/hooks/useCurrentUser.ts
@@ -1,0 +1,34 @@
+import type { APIUser } from "@humansignal/core/types/user";
+import { API } from "apps/labelstudio/src/providers/ApiProvider";
+import { useAtomValue } from "jotai";
+import { atomWithQuery, queryClientAtom } from "jotai-tanstack-query";
+import { useCallback } from "react";
+
+const currentUserAtom = atomWithQuery(() => ({
+  queryKey: ["current-user"],
+  async queryFn() {
+    return await API.invoke<APIUser>("me");
+  },
+}));
+
+export function useCurrentUser() {
+  const user = useAtomValue(currentUserAtom);
+  const queryClient = useAtomValue(queryClientAtom);
+  const refetch = useCallback(() => queryClient.invalidateQueries({ queryKey: ["current-user"] }), []);
+
+  return user.isSuccess
+    ? ({
+        user: user.data,
+        isInProgress: user.isFetching,
+        loaded: user.isSuccess,
+        error: null,
+        fetch: refetch,
+      } as const)
+    : ({
+        user: null,
+        isInProgress: user.isFetching,
+        loaded: user.isSuccess,
+        error: user.error,
+        fetch: refetch,
+      } as const);
+}
--- a/web/libs/core/src/lib/hooks/useCurrentUser.ts
+++ b/web/libs/core/src/lib/hooks/useCurrentUser.ts
@@ -14,7 +14,7 @@ const currentUserAtom = atomWithQuery(() => ({
 const currentUserUpdateAtom = atomWithMutation((get) => ({
   mutationKey: ["update-current-user"],
   async mutationFn({ pk, user }: { pk: number; user: Partial<APIUser> }) {
-    return await API.invoke("updateUser", { pk }, { body: user });
+    return await API.invoke<APIUser>("updateUser", { pk }, { body: user });
   },
 
   onSettled() {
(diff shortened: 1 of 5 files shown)


web/libs/core/src/pages/AccountSettings/sections/PersonalInfo.tsx
```diff
@@ -1,90 +1,140 @@
...
-        },
-        body: {
-          avatar: avatarRef.current.files[0],
-        },
-        headers: {
-          "Content-Type": "multipart/form-data",
-        },
-        errorFilter: () => true,
+      e.stopPropagation();
+      console.log("submitting");
```
yyassi-heartex:
```suggestion
```

yyassi-heartex (approved):

web/libs/core/src/lib/hooks/useCurrentUser.ts
```diff
@@ -0,0 +1,34 @@
...
+import { useCallback } from "react";
+
+const currentUserAtom = atomWithQuery(() => ({
+  queryKey: ["current-user"],
+  async queryFn() {
+    return await API.invoke<APIUser>("me");
+  },
+}));
+
+export function useCurrentUser() {
```
hlomzik: we already have a [hook with this name](https://github.com/HumanSignal/label-studio/blob/develop/web/apps/labelstudio/src/providers/CurrentUser.jsx#L26), this would be very confusing to use and can create inconsistent states

hlomzik (changes requested):

web/libs/core/src/lib/hooks/useCurrentUser.ts
nick-skriabin: Explained in the description.

web/libs/core/src/pages/AccountSettings/sections/PersonalInfo.tsx
```diff
@@ -1,90 +1,139 @@
...
-      if (!isDelete && response?.status) {
+
+      if (!isDelete && !response.$meta.ok) {
         toast.show({ message: response?.response?.detail ?? "Error updating avatar", type: "error" });
       } else {
-        fetch();
+        refetchUser();
       }
-      userAvatarForm.current.reset();
+      userAvatarForm.current?.reset();
```
hlomzik: should not we also use it in `fileChangeHandler()`?

hlomzik: they both look too similar. is there a way to merge them?

nick-skriabin: that's actually an unused function, removed it

hlomzik (approved):

web/libs/core/src/lib/hooks/useCurrentUser.ts
```diff
@@ -0,0 +1,61 @@
+import type { APIUser } from "@humansignal/core/types/user";
+import { API } from "apps/labelstudio/src/providers/ApiProvider";
```
bmartel, line 2:
```
import { API } from "apps/labelstudio/src/providers/ApiProvider";
```

How does this import even work, we are importing ApiProvider from the application into a lib, then using this lib in the application?

web/libs/core/src/pages/AccountSettings/sections/PersonalInfo.tsx
```diff
@@ -1,90 +1,121 @@
...
-    async (e, isDelete = false) => {
-      e.preventDefault();
-      const response = await api.callApi(isDelete ? "deleteUserAvatar" : "updateUserAvatar", {
-        params: {
-          pk: user?.id,
-        },
-        body: {
-          avatar: avatarRef.current.files[0],
-        },
+const updateUserAtom = atomWithMutation(() => ({
```
bmartel: What is the difference between this atom and the one over here: https://github.com/HumanSignal/label-studio/pull/7106/files#diff-4687f24e5860c45a41d1491638ed99b13d9308325f8409cde26c2ff1b53a2601R14

web/libs/core/src/lib/hooks/useCurrentUser.ts
nick-skriabin: yeah, kinda like that :D the thing is that `ApiProvider` is where we initialize the APIProxy instance. it has all the necessary configuration and whatnot. while I believe this should also be a part of the core, I decided not to do that in this PR as it brings a lot of changes already and I didn't want to make it bigger

the plan is to have a core version of APIProxy instance that both LSO and LSE can extend on. unfortunately, APIProxy currently doesn't allow extending configuration on an already initialized instance, so that's something we need to do prior to moving API to core

for now unfortunately that's the only way to have access to API in the core lib

vladimirheartex (approved):

PR #7117, opened 2025-02-20
docs: DOC-278: Add Anthropic to Prompts model list

caitlinwheeless:
Add Anthropic to Prompts supported models


matt-bernstein (approved):
LGTM! For Anthropic models, we could technically include Claude Opus but found that its error rate was significantly higher than other models in our constrained generation setup, so wasn't worth including - but will try again when Anthropic releases a new checkpoint. This doesn't necessarily need to be added to docs, just some color in case anyone asks.

PR #7138, opened 2025-02-24
fix: DIA-1972: [FE] Updating TTL for an org doesn't seem to update the org settings

nick-skriabin:
Fix the issue when the user is not able to update Access Token TTL settings.

Moving from autosave to explicit submit.


hlomzik:
So you are switching from autosave approach to an explicit Save button? Can you add this to description?

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
```diff
@@ -59,6 +71,16 @@ export const TokenSettingsModal = ({
           />
         </Form.Row>
       )}
+      <Form.Actions>
+        {saved && (
+          <div className={styles.success}>
+            <IconCheck /> Saved!
+          </div>
+        )}
+        <Button type="button" onClick={onSubmit} disabled={isPending || saved}>
```
hlomzik: can we use `action` in `Form` instead of this inconsistent approach?

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
```diff
@@ -13,18 +16,27 @@ export const TokenSettingsModal = ({
...
       legacy_api_tokens_enabled: data.legacy_api_tokens_enabled === "on",
     } satisfies Partial<AuthTokenSettings>;
     saveSettings(body);
-  });
+    setSaved(true);
+  };
+
+  useEffect(() => {
+    if (saved) {
+      setTimeout(() => setSaved(false), 2000);
```
hlomzik: and `Form.Indicator` instead of this?

hlomzik (changes requested):
Can we move this modal to app? That's definitely not `core`

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
yyassi-heartex: if we're going down the timeout approach we should also clear time out in cases where we set the time out again - although given the context is not a big deal - using Form.Indicator is likely the better option as we should double down on existing code

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
yyassi-heartex: +1 to this, it'll also make the form more keyboard friendly which we collectively need to take into consideration going forward

nick-skriabin:
> Can we move this modal to app? That's definitely not `core`

@hlomzik It's out of the scope for this ticket, I'd do it separately.

nick-skriabin:
@yyassi-heartex could you take another look now?

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
```diff
@@ -1,30 +1,30 @@
...
-  });
+  const reloadSettings = () => {
+    queryClient.invalidateQueries({ queryKey: [TOKEN_SETTINGS_KEY] });
+  };
+  if (!settings.isSuccess || settings.isError) {
+    return <div>Error loading settings.</div>;
+  }
+  if ("error" in settings.data) {
+    return <div>Error loading settings.</div>;
+  }
```
yyassi-heartex:
```suggestion
  if (!settings.isSuccess || settings.isError || "error" in settings.data) {
    return <div>Error loading settings.</div>;
  }
```

yyassi-heartex (approved):
minor replacement, but other wise looks good to me

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
hlomzik, line 1: why is it called index? I believe we never used this type of naming and usually it's considered not convenient.

web/libs/core/src/blocks/TokenSettingsModal/index.tsx
```diff
@@ -59,6 +56,10 @@ export const TokenSettingsModal = ({
           />
         </Form.Row>
       )}
+      <Form.Actions>
+        <Form.Indicator />
```
hlomzik: oh, great! we actually don't need to pass children in it, I didn't know, will fix some settings then

hlomzik:
> > Can we move this modal to app? That's definitely not `core`
> 
> @hlomzik It's out of the scope for this ticket, I'd do it separately.

@nick-skriabin I thought there are problems in the modal not using providers, but if not then great, approving!

hlomzik (approved):

PR #7163, opened 2025-03-04
fix: DIA-1982: update agreement when converting preds to annots

pakelley:
-------

This makes sure that annotations created from predictions get their annotation stats generated, resulting in proper agreement scores being calculated.

### DIFF
--- a/label_studio/data_manager/actions/predictions_to_annotations.py
+++ b/label_studio/data_manager/actions/predictions_to_annotations.py
@@ -59,11 +59,11 @@ def predictions_to_annotations(project, queryset, **kwargs):
         # Update counters for tasks and is_labeled. It should be a single operation as counters affect bulk is_labeled update
         project.update_tasks_counters_and_is_labeled(Task.objects.filter(id__in=tasks_ids))
 
-    try:
-        from stats.functions.stats import recalculate_stats_async_or_sync
-        recalculate_stats_async_or_sync(project, all=False)  # TODO I don't think we need all?
-    except (ModuleNotFoundError, ImportError):
-        pass
+    # try:
+    #     from stats.functions.stats import recalculate_stats_async_or_sync
+    #     recalculate_stats_async_or_sync(project, all=False)  # TODO I don't think we need all?
+    # except (ModuleNotFoundError, ImportError):
+    #     pass
 
     return {'response_code': 200, 'detail': f'Created {count} annotations'}
 
--- a/label_studio/data_manager/actions/predictions_to_annotations.py
+++ b/label_studio/data_manager/actions/predictions_to_annotations.py
@@ -63,7 +63,7 @@ def predictions_to_annotations(project, queryset, **kwargs):
             from stats.functions.stats import recalculate_stats_async_or_sync
             recalculate_stats_async_or_sync(project, all=False)
         except (ModuleNotFoundError, ImportError):
-            logger.info("Predictions converted to annotations in LSO, stats recomputation skipped")
+            logger.info('Predictions converted to annotations in LSO, stats recomputation skipped')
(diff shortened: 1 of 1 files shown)


label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -58,6 +58,13 @@ def predictions_to_annotations(project, queryset, **kwargs):
         )
         # Update counters for tasks and is_labeled. It should be a single operation as counters affect bulk is_labeled update
         project.update_tasks_counters_and_is_labeled(Task.objects.filter(id__in=tasks_ids))
+
+    try:
+        from stats.functions.stats import recalculate_stats_async_or_sync
```
makseq, line 63: OSS version doesn't have stats

label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -58,6 +58,13 @@ def predictions_to_annotations(project, queryset, **kwargs):
...
+        recalculate_stats_async_or_sync(project, all=False)  # TODO I don't think we need all?
+    except (ModuleNotFoundError, ImportError):
+        pass
```
makseq: let's write info log that in OSS version we don't have statistics and they were skipped

makseq (changes requested):

label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -58,6 +58,13 @@ def predictions_to_annotations(project, queryset, **kwargs):
         )
         # Update counters for tasks and is_labeled. It should be a single operation as counters affect bulk is_labeled update
         project.update_tasks_counters_and_is_labeled(Task.objects.filter(id__in=tasks_ids))
+
+    try:
+        from stats.functions.stats import recalculate_stats_async_or_sync
+        recalculate_stats_async_or_sync(project, all=False)  # TODO I don't think we need all?
```
makseq:
```suggestion
        recalculate_stats_async_or_sync(project, all=False)
```

all=false is ok here

makseq (changes requested):

label_studio/data_manager/actions/predictions_to_annotations.py
```diff
@@ -58,6 +58,13 @@ def predictions_to_annotations(project, queryset, **kwargs):
         )
         # Update counters for tasks and is_labeled. It should be a single operation as counters affect bulk is_labeled update
         project.update_tasks_counters_and_is_labeled(Task.objects.filter(id__in=tasks_ids))
+
+    try:
```
makseq: it's better to move this block behind `if db_anntations`

makseq: Also, it's better to check how TaskSerializerBulk.post_process_annotations() works. Probably, it makes sense to move recalculate_stats_async_or_sync inside of TaskSerializerBulk.post_process_annotations(). And it can solve other similar problems. Try searching where post_process_annotations() is used to understand it.

makseq (changes requested):

shereezhang:
I understand there can be other occurrences of this issue and other places to make improvements, but I assume a comprehensive review and improvement of TaskSerializerBulk.post_process_annotations() can take a couple days worth of time. (I'm just estimating).

Given this quick fix is meant to unblock a prospect POC, can we ship this as a standalone fix in one place, before investigating further enhancements?

This Prospect has only about 8 days left on their POC of labelstudio enterprise, and if we delay this further, this will impact the success of the POC.

hakan458 (approved):

makseq (approved):

PR #7178, opened 2025-03-07
feat: DIA-1926: return annots/preds in sample data

pakelley:
This PR was automatically generated via Follow Merge. [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/13814502645)
- https://github.com/HumanSignal/label-studio-sdk/pull/427

-------

Adds sample annotations/predictions to sample task, when it can.


matt-bernstein (commented):
Have you checked that this doesn't cause regressions or perf issues for when this endpoint is used to populate the label interface preview? I was imagining to have some kind of use_v2: bool = False param in the endpoint that would isolate it to only the new widget by default. If it just works for everything, then amazing, just gotta call out the extr test surface for QA

matt-bernstein (approved):

PR #7180, opened 2025-03-07
fix: LEAP-1892: [FE] Deeply nested Choices are visible when their parents are hidden

nass600:
Issue: Linked components were not applying correctly the visibility of their relations
conditional-choices-bug

After fix:
Level 3 should be hidden given it depends on Level 2 which is hidden as a consequence from being linked to level 1
conditional-choices-fixed


Gondragos (approved):

nass600:
@Gondragos @hlomzik this PR is ready for review again

web/libs/editor/src/regions/Result.js
```diff
@@ -158,6 +158,24 @@ const Result = types
...
+      function findNodeWithVisibleWhen(control, visibleWhen) {
+        let currentControl = control;
+
+        while (currentControl) {
+          if (currentControl.visiblewhen === visibleWhen) return currentControl;
+
+          try {
+            currentControl = getParent(currentControl);
+            if (!currentControl) break;
+          } catch {
```
hlomzik, line 171: do we expect it to break? I checked our usage of this function and there are no try-catch blocks around it. Is it a recommended way?

web/libs/editor/tests/integration/e2e/control_tags/visibility.cy.ts
hlomzik, line 1: Important missing thing in these tests is to check the serialization of the 3rd level when the first one is not selected and thus the second one is hidden as well. Current tests are checking that the third level is not visible which is great! But would also be helpful that selected choices from third level are also removed from the result.

web/libs/editor/src/regions/Result.js
nass600: Moving up in the tree in recursive manner will throw an exception always at the end node (root)

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13770359969)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/enterprise-badge/enterprise-badge.tsx

web/libs/editor/tests/integration/e2e/control_tags/visibility.cy.ts
nass600: Expanded the test to cover the scenario

hlomzik (approved):

web/libs/editor/src/regions/Result.js
```diff
@@ -201,12 +219,8 @@ const Result = types
...
       };
 
-      if (control.visiblewhen === "choice-selected") {
-        return isChoiceSelected();
-      }
-      if (control.visiblewhen === "choice-unselected") {
-        return !isChoiceSelected();
-      }
+      if (findNodeWithVisibleWhen(control, "choice-selected")) return control.isVisible && isChoiceSelected();
+      if (findNodeWithVisibleWhen(control, "choice-unselected")) return control.isVisible && !isChoiceSelected();
```
Gondragos: As I see, there are `perregion` checks in `isChoiceSelected`. It needs to be checked, but most probably we cannot use `control.isVisible` in that case. As it checks only the current visible state of the control and perregions can contain more that one result which isVisible will not take into account.

I mean this check appears in the context of a region which can be not appeared at the control tag at the time when the check runs, so we cannot rely on the control state.

nass600: Good catch @Gondragos, thanks. I am checking it now. If you have in mind any particular template where we can reproduce this specific case is more than welcome (cc: @hlomzik )

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13806294919)
> Successfully merged:  delete mode 100644 label_studio/templates/401.html

nass600:
As per discussed with @hlomzik, we are leaving out of the scope of this bug the perregion and `choice-unselected` use cases given they are not affected by the fix and they need some extra thorough analysis. cc: @Gondragos

web/libs/editor/src/regions/Result.js
```diff
@@ -201,13 +219,21 @@ const Result = types
...
         return isChoiceSelected();
       }
+
       if (control.visiblewhen === "choice-unselected") {
         return !isChoiceSelected();
       }
 
+      // We need to check if there is any node up in the tree with visibility restrictions so we can determine
+      // if the element is selected considering its own visibility
+      if (findParentWithVisibleWhen(control, "choice-selected")) {
```
hlomzik: I'd make this fix only for non-per-region tags explicitly stating it here. Right now this condition is not fully opposite to the previous one, so we might have per-region tag without visibility params but within the `View` with `visibleWhen="choice-selected"` and this can lead to errors.
```suggestion
      if (!control.perregion && findParentWithVisibleWhen(control, "choice-selected")) {
```

nass600: Added

hlomzik (approved):

MihajloHoma:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13851880908)
> Successfully merged:  delete mode 100644 web/libs/editor/src/components/Timeline/Controls/Info.tsx

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13851972172)

PR #7198, opened 2025-03-11
chore: DIA-2026: copy changes

pakelley:
-------

This improves some of the wording for the personal access token settings.

### DIFF
--- a/web/libs/core/src/pages/AccountSettings/sections/index.tsx
+++ b/web/libs/core/src/pages/AccountSettings/sections/index.tsx
@@ -41,7 +41,7 @@ export const accountSettingsSections = (settings: AuthTokenSettings): SectionTyp
       },
     settings.legacy_api_tokens_enabled && {
       title: isFF(FF_AUTH_TOKENS) ? "Legacy Token" : "Access Token",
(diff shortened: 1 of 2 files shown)


web/libs/core/src/pages/AccountSettings/sections/index.tsx
```diff
@@ -41,7 +41,7 @@ export const accountSettingsSections = (settings: AuthTokenSettings): SectionTyp
       },
     settings.legacy_api_tokens_enabled && {
       title: isFF(FF_AUTH_TOKENS) ? "Legacy Token" : "Access Token",
-      id: "personal-access-token",
+      id: "legacy-token",
```
hakan458, line 44: what is this change for?

hakan458 (approved):

web/libs/core/src/pages/AccountSettings/sections/index.tsx
pakelley: This just fixes the anchor link when you click this item in the side bar, so it goes to the legacy token section (rather than the personal access token section)

PR #7206, opened 2025-03-11
feat: DIA-1924: "Import sample data" ability in the Data Import page

matt-bernstein:
[x] add sample data import component
[x] add configurable manifest file with sample data
[x] create default manifest file with data hosted on our s3

image


nick-skriabin:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13801476531)

web/apps/labelstudio/src/pages/CreateProject/CreateProject.jsx
```diff
@@ -126,13 +128,29 @@ export const CreateProject = ({ onClose, redirect = true }) => {
     }),
     [name, description, config],
   );
+  const uploadSample = useCallback(
```
nass600: I would suggest to adding the interaction with the API to a hook and using it the same way as line 109. Probably the hook should be located in `labelstudio/src/hooks`

web/apps/labelstudio/src/pages/CreateProject/Import/ImportModal.jsx
```diff
@@ -30,6 +31,22 @@ export const Inner = () => {
     return refresh(pathname);
   }, [location, history]);
 
+  const uploadSample = useCallback(
```
nass600: I would use the same hook stated previously

nass600 (changes requested):

web/libs/ui/package.json
```diff
@@ -3,7 +3,9 @@
   "version": "0.0.0",
   "license": "MIT",
   "private": true,
-  "dependencies": {},
+  "dependencies": {
+    "@radix-ui/react-select": "^2.1.6"
+  },
```
bmartel:
```suggestion
  "dependencies": {},
```

This shouldn't have its own dependencies as it is a monorepo

bmartel (changes requested):
small observation on the package.json, and I agree with the other feedback that we should look to make the hooks in the same way to align with the existing.

matt-bernstein:
Sample 1 (images) is not actually importing - looking into it. Sample 2 works.

edit: fixed

web/libs/core/src/blocks/SampleDatasetSelect/SampleDatasetSelect.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const onSelect = useCallback(
+    (value: string) => {
+      onSampleApplied(samples.find((s) => s.url === value));
+    },
+    [samples, onSampleApplied],
+  );
+
+  return (
+    <div className="flex gap-3 items-center">
+      <span className="text-lsNeutralContentSubtler">or use a sample dataset</span>
```
nass600, line 32: Small detail: this text "or use a sample dataset" has a different shade of gray as the "or" right before (between the Add URL and Upload files buttons). I guess they should be the same color unless design tells otherwise right?

web/apps/labelstudio/src/pages/CreateProject/Import/utils.ts
```diff
@@ -0,0 +1,40 @@
+import { API } from "apps/labelstudio/src/providers/ApiProvider";
```
nass600, line 1: If this function is meant to be a hook I believe we should place it under the `labelstudio/src/hooks` directory and name it `useImportFiles`

nass600 (changes requested):

web/apps/labelstudio/src/pages/CreateProject/Import/utils.ts
nick-skriabin: it was not meant to be a hook and it's not used as one. just a utility function. considering there's no reactivity, I don't see a reason to make it a hook

web/libs/core/src/blocks/SampleDatasetSelect/SampleDatasetSelect.tsx
nick-skriabin: the color used for the "or" before is `--sand_600`:

image

image


so the colors match:

`colors.scss`
```
  --sand_600: #6B6860;
```

`tailwind.config.js`
```
lsNeutralContentSubtler: "#6B6860",
```

nick-skriabin:
@yyassi-heartex @bmartel @nass600 please, take another look

web/libs/core/src/blocks/SampleDatasetSelect/SampleDatasetSelect.tsx
nass600: You are totally right, my bad

nass600 (approved):

web/libs/ui/src/tailwind.config.js
bmartel, line 1: We should be looking to use the real tokens from the design system now that they are in, and phase out these intermediary ones.

bmartel (approved):

PR #7209, opened 2025-03-11
fix: OPTIC-1749: Limit only sentry_force error logs to go to Sentry

mcanu:
In https://github.com/HumanSignal/label-studio/pull/7187 we let error logs go to sentry, but this caused an overflow of issues that aren't errors. As an alternative, we'll send error logs only when they have exceptions (current behavior) or when they are forced.


luarmr (approved):

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13798384988)

PR #7210, opened 2025-03-11
fix: OPTIC-1794: Improve handling of missing tasks and projects when fetching data within DataManager

bmartel:
This pull request includes a couple updates to error handling within the `DataManager` component. We are introducing more graceful ways of handling missing tasks and project data, siding with redirects and more subtle toast messages instead of crashing the page and leaving the user in an unusable state of the application. We are now no longer going to treat polling data the same as normal data requests, as one is non critical and nice to have, the other must exist for the application to work. In the case of polling `interaction="timer"` data we will ignore these errors completely and allow it to retry. The non polling data fetching the main project or task will be handled with a redirect back to the page previous, DataManager of the project (missing task), or the projects list itself (missing project) in that order based on what the missing data type is.

[Kapture 2025-03-11 at 20.40.27.webm](https://github.com/user-attachments/assets/d1080771-ec36-41d5-a4ae-d8c9dca276ab)

### Error Handling and Redirection Improvements:

* [`web/apps/labelstudio/src/pages/DataManager/DataManager.jsx`](diffhunk://#diff-ee4ec91c0e13fd4685594a907d22c9ccae9ef8cce012fd603d19584690cb124aL99-R107): Enhanced the crash event handling to redirect users based on the error details, either to the home page or the projects page.
* [`web/libs/datamanager/src/stores/AppStore.js`](diffhunk://#diff-d1e11a839827428aa6aada8e027a22b057a463ccb73c693e464c6e0784a9825fL742-R758): Updated the `crash` method to accept an options parameter, allowing conditional redirection based on the error context.
* [`web/libs/datamanager/src/stores/AppStore.js`](diffhunk://#diff-d1e11a839827428aa6aada8e027a22b057a463ccb73c693e464c6e0784a9825fL507-R516): Modified the project fetching logic to handle cases where the project does not exist, triggering a crash with redirection.

### Project and Task Data Handling:

* [`web/libs/datamanager/src/stores/AppStore.js`](diffhunk://#diff-d1e11a839827428aa6aada8e027a22b057a463ccb73c693e464c6e0784a9825fR487-R498): Added a check for the length of the new project data to ensure valid comparison and updating of the project state.
* [`web/libs/datamanager/src/stores/DataStores/tasks.js`](diffhunk://#diff-524bd37e2870a2414a48568a8f21257ae07d41ce3e1bfe8fecbf02a0ddf4ce5bL148-R160): Enhanced the task data retrieval process to handle task cancellations and non-existent tasks, invoking a crash with redirection if necessary.


nick-skriabin (approved):

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -96,7 +96,27 @@ export const DataManagerPage = ({ ...props }) => {
...
+          message: `The ${
+            isMissingTaskError ? "task" : "project"
+          } you are trying to access does not exist or is no longer available.`,
+          type: ToastType.error,
+          duration: 10000,
+        });
+      }
+
+      if (isMissingTaskError) {
+        history.push(buildLink("/", { id: params.id }));
```
hlomzik: Current code will redirect to pages like `/projects/123/` with trailing slash which we normally don't use
```suggestion
        history.push(buildLink("", { id: params.id }));
```

web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -96,7 +96,27 @@ export const DataManagerPage = ({ ...props }) => {
...
-    dataManager.on("crash", () => setCrashed());
+    dataManager.on("crash", (details) => {
+      const error = details?.error;
+      const isMissingTaskError = error?.startsWith("Task ID:");
+      const isMissingProjectError = error?.startsWith("Project ID:");
+
+      if (isMissingTaskError || isMissingProjectError) {
+        toast.show({
+          message: `The ${
+            isMissingTaskError ? "task" : "project"
```
hlomzik: I'd move this to variable before the `toast.show()` for readability

web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -484,17 +484,18 @@ export const AppStore = types
       try {
         const newProject = yield self.apiCall("project", params);
         const projectLength = Object.entries(self.project ?? {}).length;
+        const newProjectLength = Object.entries(newProject ?? {}).length;
```
hlomzik: The naming is really confusing here. better to rename it to something like `isProjectEmpty` and store boolean directly. We don't actually need a number of properties in project.

hlomzik (approved):
Some non-critical observations but would be really great to fix them, they are simple and precise.

web/libs/datamanager/src/stores/AppStore.js
bmartel:
```suggestion
        const hasExistingProjectData = Object.entries(self.project ?? {}).length > 0;
        const hasNewProjectData = Object.entries(newProject ?? {}).length > 0;
```

web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -484,17 +484,18 @@ export const AppStore = types
...
 
         self.needsDataFetch =
-          options.force !== true && projectLength > 0
+          options.force !== true && projectLength > 0 && newProjectLength > 0
```
bmartel:
```suggestion
          options.force !== true && hasExistingProjectData && hasNewProjectData
```

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13830987940)

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/13834430241)

PR #7216, opened 2025-03-12
fix: OPTIC-1809: Adding option to only update last_activity after a significant difference

wesleylima:
…


label_studio/users/models.py
```diff
@@ -65,6 +65,12 @@ class UserLastActivityMixin(models.Model):
     last_activity = models.DateTimeField(_('last activity'), default=timezone.now, editable=False)
 
     def update_last_activity(self):
+        check_interval = getattr(settings, 'USER_LAST_ACTIVITY_CHECK_INTERVAL', None)
+        if check_interval:
+            if timezone.now() - self.last_activity < check_interval:
+                # Don't update last_activity because it's not necessary since it's close enough to the existing last_activity value
```
jombooth, line 76: should we log a message here so we at least have a record somewhere of the real last_activity?

label_studio/users/models.py
```diff
@@ -65,6 +65,12 @@ class UserLastActivityMixin(models.Model):
     last_activity = models.DateTimeField(_('last activity'), default=timezone.now, editable=False)
 
     def update_last_activity(self):
+        check_interval = getattr(settings, 'USER_LAST_ACTIVITY_CHECK_INTERVAL', None)
```
jombooth: why do we need `getattr` here?

label_studio/users/models.py
wesleylima: added!

label_studio/users/models.py
wesleylima: True, we can access USER_LAST_ACTIVITY_CHECK_INTERVAL directly since it will very reliably exist in settings.

jombooth (approved):
LG, please add a description to the PR before merging!

PR #7269, opened 2025-03-21
chore: OPTIC-1842: Update the css variables for colors to support RGB and drop the primitive prefixes on all variables

bmartel:
## Reason for change

The design tokens required an alignment of the naming used for the css vars, as well as support rbg and transparency values so they will directly work with the upcoming Dark Mode.

### Updated Semantic Tokens

Tailwind tokens to Figma design token mappings that were replaced across this PR.

**Color**

`lsPrimaryContent` -> `primary-content`
`lsPrimaryBackground` -> `primary-background`
`lsPrimaryOnBackground` -> `accent-grape-dark`
`lsPrimaryEmphasisSubtle` -> `primary-emphasis-subtle`
`lsPrimaryIcon` -> `primary-icon`
`lsSubtitle` -> `neutral-content-subtler`
`lsNeutralContent` -> `neutral-content`
`lsNeutralContentSubtle` -> `neutral-content-subtle`
`lsNeutralContentSubtler` -> `neutral-content-subtler`
`lsNeutralContentSubtlest` -> `neutral-content-subtlest`
`lsNeutralSurface` -> `neutral-surface`
`lsNeutralSurfaceActive` -> `neutral-surface-active`
`lsNeutralBorder` -> `neutral-border`
`lsNeutralBorderSubtle` -> `neutral-border-subtle`
`lsNeutralBorderBold` -> `neutral-border-bold`
`lsBorderLightest` -> `primary-emphasis`
`lsBorderSubtle` -> `primary-border-subtle`
`lsButtonBg` -> `primary-content`
`lsPositiveSurfaceHover` -> `positive-surface-hover`
`lsAccentMangoSubtle` -> `accent-mango-subtle`
`lsAccentMangoBold` -> `accent-mango-bold`
`lsAccentGrapeSubtle` -> `accent-grape-subtle`

**Font Size**
`lsLabelMedium` -> `16/24`

### Storybook

Added all design tokens to storybook so they can be searched and visualized, so we can continue to iterate components and other design elements collaboratively and with high visibility.

Screenshot 2025-03-21 at 4 51 52 PM


web/libs/ui/src/tokens/colors.scss
```diff
@@ -16,44 +16,44 @@
...
-  --grape_0: var(--color-primitive-grape-000);
-  --grape_100: var(--color-primitive-grape-100);
+  --grape_0: var(--color-grape-000);
+  --grape_100: var(--color-grape-100);
   --grape_500-raw: 86 111 207;
-  --grape_500: rgb(var(--grape_500-raw));
-  --grape_600: var(--color-primitive-grape-600);
-  --grape_700: var(--color-primitive-grape-700);
-  --grape_800: var(--color-primitive-grape-800);
+  --grape_500: var(--color-grape-500);
```
yyassi-heartex, line 53: lets remove `-grape_500-raw` if we're no longer using that

yyassi-heartex (approved):
minor feedback - other wise looks good

web/libs/ui/src/tokens/colors.scss
bmartel: One thing I have to still figure out, is how we are going to approach the raw values in general. Still sorting that out actually, but also whatever we do there, we have to remember about the other theming that was done against these old raw values.

yyassi-heartex: ahh ok, sounds like a follow up PR

bmartel: I added support, but the alignment and usage will definitely come in a follow up PR, likely as part of completing Dark Mode.

ricardoantoniocm:
image

On the Tokens / Index page:

Are we able to display the name of each color palette without requiring hover?

Also, could we separate the Accent colors from the Semantic ones?

ricardoantoniocm:
Currently there's no differentiation between some primitive variables vs semantic ones:
E.g.
Primitives: 
--spacing-400
--corner-radius-8
--typography-line-height-12
etc

Semantic:  
--spacing--base 
--corner-radius-medium
etc

ricardoantoniocm:
Not sure if this is intended, but the filter for corner-radius variables is in camel case (cornerRadius).

ricardoantoniocm:
The outline-primary example (border: 3px solid rgb(var(--color-primary-focus-outline-raw) / 30%);
) is not a good one, as the focus outline is not intended to be transparent. Let's just keep the shadow example if possible for now.

ricardoantoniocm:
Spacing variables on the Index page are pointing to primitive variables, not the semantic counterparts: none, tighest, tight, etc.

PR #7273, opened 2025-03-24
fix: OPTIC-1850: 'n is not a function' RUNTIME error occur on ls-develop when user tries to create a project

yyassi-heartex:
due to the upgrade to react 18 - we need to make sure that if a useEffect function only gets a destroy function when appropriate. we have use cases that shouldnt be returning to useEffect so we update it here


bmartel (approved):

nass600 (approved):

nick-skriabin (approved):

nick-skriabin:
lgtm!

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14054261279)

PR #7277, opened 2025-03-24
fix: DIA-2096: don't use is_owner in LSO

pakelley:
-------

Don't use is_owner in LSO when trying to determine whether we can get the JWT settings, since this attr isn't available.


hakan458 (approved):

pakelley:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14052646567)

pakelley:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14053061409)

PR #7278, opened 2025-03-24
fix: DIA-2103: rollback JWT token FF enablement

pakelley:
-------

Rollback the enablement of the JWT token FF, as it seems our tests are not set up to handle this being enabled


pakelley:
This was a temporary fix that was implemented using CI instead. Will implement the long-term test fixes in other PRs.

PR #7280, opened 2025-03-25
fix: OPTIC-1857: User can NOT import data from Import modal from DataManager page

yyassi-heartex:
issue was caused by attempting to work with sample when sample was not defined.

This pull request includes a small change to the `ImportModal.jsx` file. The change ensures that the `uploadSample` function is called only if the `sample` variable is defined.

* [`web/apps/labelstudio/src/pages/CreateProject/Import/ImportModal.jsx`](diffhunk://#diff-35ca1010f69b1d6e041757164e573c7f012b2fa00e7f435fefdb66fd7c39ebb4R50-R56): Added a conditional check to call `uploadSample` only when `sample` is defined.


bmartel (approved):

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14078272803)

nass600 (approved):

PR #7285, opened 2025-03-25
fix: DIA-2103: re-enable jwt token FF

pakelley:
-------

Fix tests broken by enabling JWT token FF

### DIFF
--- a/label_studio/tests/conftest.py
+++ b/label_studio/tests/conftest.py
@@ -467,7 +467,7 @@ def project_choices():
     return {'label_config': label, 'title': 'test'}
 
 
-def setup_project(client, project_template, do_auth=True):
+def setup_project(client, project_template, do_auth=True, legacy_api_tokens_enabled=False):
     """Create new test@gmail.com user, login via client, create test project.
     Project configs are thrown over params and automatically grabs from functions names started with 'project_'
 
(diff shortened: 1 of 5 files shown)


label_studio/tests/conftest.py
```diff
@@ -467,7 +467,7 @@ def project_choices():
     return {'label_config': label, 'title': 'test'}
 
 
-def setup_project(client, project_template, do_auth=True):
+def setup_project(client, project_template, do_auth=True, legacy_api_tokens_enabled=False):
```
hakan458, line 470: so we will not be using legacy tokens for any of these tests using setup_project (assume its used in a lot of places) ?

label_studio/tests/webhooks/organizations.tavern.yml
```diff
@@ -57,6 +57,17 @@ stages:
...
   type: ref
 
+# Test this functionality with legacy tokens, need to enable them for the org
+- name: enable_legacy_api_tokens
+  request:
+    url: "{django_live_url}/api/current-user/whoami"
+    method: GET
+  response:
+    status_code: 200
+    verify_response_with:
```
hakan458: why are we using it this way..? enabling something for the org, through a response verify function?? Why cant we enable it with a normal API call (maybe there isnt one)? This seems confusing and will be missed if anyone has to figure out whats going on

pakelley: Ok yeah, fair enough, I'll change that

label_studio/tests/conftest.py
pakelley: Yeah, I tried to strike some kind of balance between migrating tests to using the new system and keeping them on the old system. Tests that seemed to be testing the token functionality more directly I kept on the old tokens, but tests that seemed agnostic I just left using the new system

hakan458 (approved):

PR #7289, opened 2025-03-26
fix: DIA-1907: [FE] TypeError: Failed to fetch

nick-skriabin:
Now handling the Failed to fetch error.

According to MDN it can happen due to network conditions or CORS error. In both cases we don't want to bother user with it, so we fail silently.

---

This pull request includes several important changes to the `useAbortController` hook and its usage across the codebase, as well as improvements to error handling in the `APIProxy` class.

### Changes to `useAbortController` Hook:

* **Refactored `useAbortController` Hook**: The `useAbortController` hook was moved from `web/apps/labelstudio/src/hooks/useAbortController.ts` to `web/libs/core/src/hooks/useAbortController.ts` and refactored to include an `abort` method for better error handling. [[1]](diffhunk://#diff-98547faed7cb056fc233b69511940666472d6f81da5267d02f675c3e89fcc3ecL1-L22) [[2]](diffhunk://#diff-86c7acfe099351a04a8b880df330b20f54baa287702f8b28e6b4a33b543779bbR1-R54)
* **Updated Imports**: Updated the import path for `useAbortController` in `Projects.jsx` to reflect the new location in `@humansignal/core`.
* **Export New Hook**: Added export for the refactored `useAbortController` in `web/libs/core/src/index.ts`.

### Improvements to Error Handling in `APIProxy`:

* **Ignored Errors**: Introduced a new regular expression `IGNORED_ERRORS` to catch and silently handle specific types of errors such as abort errors and network issues.
* **Enhanced Error Handling**: Modified the `APIProxy` class to use the `IGNORED_ERRORS` regex for improved error handling, ensuring certain errors do not reach the user.


yyassi-heartex (approved):

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14102052640)

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14130667682)

web/libs/core/src/lib/utils/safe-fetch.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @returns true if the error should be ignored, false otherwise
+ */
+export function isErrorIgnored(error: unknown) {
+  if (!(error instanceof Error)) return false;
+  // we don't want the user to see some of the errors
+  // so we fail silently
+  if (error.message.match(IGNORED_ERRORS) !== null) {
+    return true;
+  }
+}
```
hlomzik: for consistent results
```suggestion
  return false;
}
```

web/libs/core/src/hooks/useAbortController.ts
```diff
@@ -0,0 +1,54 @@
...
+    try {
+      ctrl.abort();
+    } catch (err) {
+      // handling errors during abort (especially "signal is aborted without a reason")
+      if (err instanceof Error && err.name.match(/abort/gi)) {
+        console.info(`AbortError: ${controller.current.signal.reason ?? "Operation aborted"}`);
+      } else {
+        console.error("Unexpected error during abort:", err);
+      }
+    }
```
bmartel: This isn't needed as it won't ever throw here, the exception is raised in the original fetch call:

> When abort() is called, the fetch() promise rejects with an Error of type DOMException, with name AbortError

https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#examples

nass600 (approved):

PR #7299, opened 2025-03-27
fix: OPTIC-1858: Wait for the store and currentAnnotation before proceeding to complete task initialization in Datamanager Quickview

bmartel:
This change was required to unblock the release candidate build testing. Through the investigation that lead to this solution, it was possible to be able to influence the paths that would be taken with regards to task initialization, as there was an interdependency on the Datamanager component `Label.jsx` with a AppStore `fetchData` call. Towards the bottom of the`fetchData` call there is a call to `startLabeling`, which explicitly requires that the root element be present prior to `startLabeling` being called and it is also only mounted **after** the loading completes. This inevitably was causing the code for deep linking resolution to break consistently, which I don't quite understand what the factor was with regards to that environment specifically, but did note that React 18 was already present on all other environments where it does consistently work and the release candidate is still on React 17, so this patch is written in a way that should work regardless and have no overhead in React 18 which has a completely renewed lifecycle management to React 17.

Changes:
- Always retain the root DM element so it can be bound on first render cycle
- Due to the AppStore being a global element, it can and will call startLabeling at various points where we could be unmounting and destroying the references, so there was a need to wait for LSF.currentAnnotation to become hydrated, as there is a chance it is not currently.


web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -225,7 +211,29 @@ export const DataManagerPage = ({ ...props }) => {
...
+            top: 0,
+            left: 0,
+            right: 0,
+            bottom: 0,
+            width: "100%",
+            height: "100%",
+            display: "flex",
+            alignItems: "center",
+            justifyContent: "center",
+          }}
```
yyassi-heartex: a css class here would be ideal if possible

bmartel: Done

yyassi-heartex (approved):

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14124701968)

nass600 (approved):

PR #7301, opened 2025-03-27
fix: DIA-2116: Export storages can be incorrectly retrieved as import storages

matt-bernstein:
...due to an incorrect use of caching. Replaced it with @cached_property. There was no active code that would have hit this bug, but I was about to write some for DIA-2033 :slightly_smiling_face:

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -30,6 +30,7 @@ from django.conf import settings
 from django.core.validators import MaxLengthValidator, MinLengthValidator
 from django.db import models, transaction
 from django.db.models import Avg, BooleanField, Case, Count, JSONField, Max, Q, Sum, Value, When
+from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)


label_studio/projects/models.py
```diff
@@ -997,23 +998,30 @@ def get_active_ml_backends(self):
...
+    @cached_property
+    def get_all_export_storage_objects(self):
+        from io_storages.models import get_storage_classes
 
         storage_objects = []
-        for storage_class in get_storage_classes(type_):
+        for storage_class in get_storage_classes('export'):
             storage_objects += list(storage_class.objects.filter(project=self))
 
-        self._storage_objects = storage_objects
```
hakan458, line 1010: was this the issue? We would save import storages here, then above 
```
if hasattr(self, '_storage_objects'):
```
We would return import storages even though we requested export?

matt-bernstein: yep, that's my read, we just don't request export anywhere

hakan458 (approved):

PR #7309, opened 2025-03-31
feat: LEAP-1973: Adjustable spans for Text tag

hlomzik:
Allow users to resize selected NER regions.

https://github.com/user-attachments/assets/1e6db970-6719-4edd-ad30-fb0b5fdc020a

The main idea is to use browser selection with styles of the region so the region expansion will look
smooth and natural. At the beginning the selection will be set to wrap the region. So users will just change
the selection during drag-n-drop, no extra code required for the visual part.

Steps to achieve this:
1. Detect that user is about to resize the region (mousedown on a handle)
2. Set selection style to the region's style
3. Set selection range to the region's range for the initial state
4. We do nothing on mousemove, and on mouseup we update the region's offsets
5. Remove selection style

There are some tricky parts here: selection should be created after we started dragging,
if we create it before, then the drag will drag-n-drop the selection instead of continuing it.
Steps and flags on init:
- mousedown: check if we are on a handle, set `draggableRegion` and `dragBackwards` to understand the direction
- mousemove: if we have `draggableRegion` but don't have `initializedDrag`, set it; that means we are dragging
- mousemove: if we have both, create selection around the region, set `currentSelection`;
             we are editing the region now
And we should reset all of them on mouseup.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14311504047)
> Successfully merged:  create mode 100644 web/libs/ui/src/shadcn.ts

web/libs/editor/src/tags/object/RichText/RichText.scss
```diff
@@ -2,9 +2,9 @@
   position: relative;
 
   &__container {
-    overflow: auto;
+    // overflow: auto;
```
bmartel: Should be removed if not needed

web/libs/editor/src/tags/object/RichText/RichText.scss
```diff
@@ -83,8 +83,59 @@
...
+        width: 100%;
+        left: 0;
+      }
+      // popsicle sphere
+      &::after {
+        content: '';
+        position: absolute;
+        width: $handle_r * 2 + 1px;
+        height: $handle_r * 2 + 1px;
+        background: var(--grape_500);
```
bmartel: Should be using a semantic CSS var

web/libs/editor/src/tags/object/RichText/RichText.scss
```diff
@@ -83,8 +83,59 @@
...
+      box-sizing: content-box;
+      margin-left: -$hitbox_w - 1px;
+
+      // popsicle stick
+      &::before {
+        position: absolute;
+        content: '';
+        width: 1px;
+        height: 100%;
+        background: var(--grape_500);
```
bmartel: Should be using a semantic CSS var

web/libs/editor/src/tags/object/RichText/RichText.scss
```diff
@@ -83,8 +83,59 @@
...
+        left: $hitbox_w - $handle_r;
+        z-index: 100;
+        border-radius: 2px;
+      }
+      &:global(.__resize_right)::after {
+        top: unset;
+        bottom: -($handle_r * 2 + 1px);
+      }
+      &:hover::after {
+        outline: 1px solid var(--grape_500);
```
bmartel: Should be using a semantic CSS var

bmartel (approved):

web/libs/editor/src/tags/object/RichText/RichText.scss
bmartel:
```suggestion
        background: var(--color-primary-content-subtle);
```

web/libs/editor/src/tags/object/RichText/RichText.scss
bmartel:
```suggestion
        background: var(--color-primary-content-subtle);
```

web/libs/editor/src/tags/object/RichText/RichText.scss
bmartel:
```suggestion
        outline: 1px solid var(--color-primary-border-subtle);
```

bmartel:
Could you add a screenshot of what this looks like?

web/libs/editor/src/mixins/HighlightMixin.js
```diff
@@ -73,19 +80,32 @@ export const HighlightMixin = types
...
-        const lastSpan = self._spans[self._spans.length - 1];
+        const firstSpan = self._spans[0];
+        const lastSpan = self._spans.at(-1);
+        const offsets = self.globalOffsets;
 
         // @TODO: Should we manage it in domManager?
+        // update label tag (index + attached labels) which sits in the last span
         Utils.Selection.applySpanStyles(lastSpan, { index: self.region_index, label: self.getLabels() });
+        // store offsets in spans for further comparison if region got resized
+        firstSpan.setAttribute("data-start", offsets.start);
```
Gondragos, line 118: Most likely this can be stored in a volatile variable or defined inside `.views(() => ({}))`, to avoid polluting the DOM.

web/libs/editor/src/mixins/HighlightMixin.js
```diff
@@ -101,20 +121,49 @@ export const HighlightMixin = types
...
+     */
+    attachHandles() {
+      const classes = [STATE_CLASS_MODS.leftHandle, STATE_CLASS_MODS.rightHandle];
+      const spanStart = self._spans[0];
+      const spanEnd = self._spans.at(-1);
+
+      classes.forEach((resizeClass, index) => {
+        // html element that can't be encountered in a usual html
+        const handleArea = document.createElement("area");
+
```
Gondragos, line 175: It seems more efficient to store `handleArea` in the model's internal data, so that when using `querySelectorAll`, no more elements than planned are removed.

web/libs/editor/src/tags/object/RichText/RichText.scss
```diff
@@ -83,8 +83,59 @@
       }
     }
   }
+
+  :global(.htx-highlight) {
+    // resize handles around selected region,
+    // they look like popsicles, normal one at the beginning, upside down one at the end
+    area {
```
Gondragos, line 96: It might be better to rely on `__resize_right` and `__resize_left` instead of the tag name. Just in case the tag appears in the content being marked up.

web/libs/editor/src/tags/object/RichText/view.jsx
```diff
@@ -51,12 +52,205 @@ class RichTextPieceView extends Component {
...
+   * And we should reset all of them on mouseup.
+   **/
+
+  /**
+   * Adjust selection style to mimic region's style; this is done by creating a style tag
+   * @param {*} region to mimic
+   * @param {Document} doc document to apply style to
+   */
+  _setSelectionStyle = (region, doc) => {
+    const styleMap = region._spans[0].computedStyleMap();
```
Gondragos: Most likely, we have the necessary data within the model itself.

web/libs/editor/src/tags/object/RichText/view.jsx
```diff
@@ -51,12 +52,205 @@ class RichTextPieceView extends Component {
...
+  /**
+   * Check if the target is a handle and prepare dragging if it is. Set the `draggableRegion` to mark this.
+   * @param {Event} ev Mouse down event
+   */
+  _checkHandlesAndStartDragging = (ev) => {
+    const { item } = this.props;
+    const target = ev.target;
+    const region = this._determineRegion(target);
+    const classes = [STATE_CLASS_MODS.leftHandle, STATE_CLASS_MODS.rightHandle];
+    const isHandle = target.classList.contains(classes[0]) || target.classList.contains(classes[1]);
```
Gondragos, line 127: If you could store references to handle, you can compare the refs instead of the classes.

Gondragos (approved):

web/libs/editor/src/mixins/HighlightMixin.js
```diff
@@ -29,30 +33,44 @@ export const HighlightMixin = types
...
         .${className}.${STATE_CLASS_MODS.active}:not(.${STATE_CLASS_MODS.hidden}) {
-          color: ${Utils.Colors.contrastColor(initialActiveColor)} !important;
-          background-color: ${initialActiveColor} !important;
+          color: ${colors.activeText} !important;
+          background-color: ${resize ? colors.resizeBackground : colors.activeBackground} !important;
         }
       `;
     },
+    get styles() {
+      return this.generateStyles(self.className, self.getColors());
```
Gondragos, line 69: It looks like styles and resizeStyles are the same thing and could simply be recalculated based on an internal volatile flag instead of creating generators and complicating dependencies.

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14361897145)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/control_tags/from-prediction.cy.ts

web/libs/editor/src/mixins/HighlightMixin.js
hlomzik: Unfortunately the whole styles approach here is too complex anyway. For example, styles are changed by the method and colors they rely on are not in `views` for some reason. And colors are multiple values. But I'll add this as an idea to follow up ticket. And I added a todo already about colors and views.

MihajloHoma:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14374942516)
> Successfully merged:  create mode 100644 web/libs/storybook/addons/theme-toggle/theme-toggle.tsx

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14374987905)

nass600 (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14381797285)

web/libs/editor/src/tags/object/RichText/RichText.scss
hlomzik: rewrote it with FF just in case, because it might change the overall experience with Text tag. thank you!

hlomzik:
@Gondragos thank you, I added link to your comments to follow-up ticket (LEAP-2000)

hlomzik:
/help
> Command | Description
> --- | ---
> /frontend [\<args\> ...] | Actions with frontend. Type `/frontend help` for an additional help.

hlomzik:
> Could you add a screenshot of what this looks like?

Added a video to PR description, thank you for call out!

PR #7323, opened 2025-04-02
fix: DIA-2122: allow all LSO users to modify JWT settings

pakelley:
-------

Changes JWT settings permissions, so that in the case of LSO all users will be able to change JWT permissions.

Added test cases for this, and the LSE cases already have tests in LSE to verify that the functionality works properly there (i.e. that only owners/admins can change JWT settings).

### DIFF
--- a/label_studio/core/utils/common.py
+++ b/label_studio/core/utils/common.py
@@ -750,3 +750,17 @@ def empty(*args, **kwargs):
 def get_ttl_hash(seconds: int = 60) -> int:
     """Return the same value within `seconds` time period"""
     return round(time.time() / seconds)
+
+def is_enterprise():
+    """Determine if the current Label Studio instance is Enterprise (LSE) or open-source (LSO).
+
+    Returns
+    -------
+    bool
+        True if running Label Studio Enterprise, False if running open-source Label Studio
+    """
+    try:
+        import label_studio_enterprise
+        return True
+    except ImportError:
+        return False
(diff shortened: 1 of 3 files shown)


label_studio/core/utils/common.py
```diff
@@ -750,3 +750,17 @@ def empty(*args, **kwargs):
...
+    Returns
+    -------
+    bool
+        True if running Label Studio Enterprise, False if running open-source Label Studio
+    """
+    try:
+        import label_studio_enterprise
+        return True
+    except ImportError:
+        return False
```
jombooth: Proposing that we invert this, because elsewhere in the code we check for enterprise meaning "enterprise, as opposed to starter cloud" rather than "enterprise, as opposed to LSO"

```suggestion
def is_community():
    """Determine if the current Label Studio instance is the community edition (aka LSO).

    Returns
    -------
    bool
        True if running open-source Label Studio, False otherwise.
    """
    try:
        import label_studio_enterprise
        return False
    except ImportError:
        return True
```

pakelley: GH wouldn't let me apply this directly, but I made these changes. Good suggestion, I totally agree with the thinking here 👍

label_studio/jwt_auth/models.py
```diff
@@ -2,6 +2,7 @@
 from typing import Any
 
 from annoying.fields import AutoOneToOneField
+from core.utils.common import is_enterprise
```
jombooth:
```suggestion
from core.utils.common import is_community
```

jombooth (approved):
LG!

PR #7334, opened 2025-04-04
feat: DIA-2092:  Support reading multiple tasks from each file in a cloud storage

matt-bernstein:
[x] allow reading multiple tasks from the same file during import storage sync in scan_and_create_links
[x] update get_data for all import storage types to return list[dict] instead of dict
[x] handle blob urls? ~currently, still assuming only 1 task per file using these. Changing this would have a larger blast radius, would have to read the url at import time to determine how many tasks are in it, and then make up a task format for pointing to an object within a url, and read the url n times for n tasks during normal usage unless it's cached somehow...~ use blob urls == "Treat every bucket object as a source file", so it doesn't make sense to do this
[x] do the same for LSE S3 and GCS WIF import storage
[x] add test coverage - currently most test coverage for io_storages is in a single tavern file. Each import storage class definitely needs its own coverage for multiple tasks per key, so this'll be pretty lengthy


pakelley (approved):

label_studio/io_storages/azure_blob/models.py
```diff
@@ -122,20 +122,28 @@ def iterkeys(self):
...
         blob = container.download_blob(key)
         blob_str = blob.content_as_text()
         value = json.loads(blob_str)
-        if not isinstance(value, dict):
+        if isinstance(value, dict):
+            return [value]
+        elif isinstance(value, list):
+            for idx, item in enumerate(value):
+                if not isinstance(item, dict):
+                    raise ValueError(
```
pakelley, line 226: Unfortunately, I don't think this message will actually reach the user. I think the better behavior may be to log the exception (so we see it in sentry) and import anything we can

pakelley: Worth checking w/ Sheree if she has an opinion here though tbh

label_studio/io_storages/base_models.py
```diff
@@ -398,26 +398,27 @@ def _scan_and_create_links(self, link_class):
...
-            # `emit_webhooks_for_instance`, and `tasks_for_webhook` is cleared for new tasks.
-            # If tasks remain in `tasks_for_webhook` at process end (less than `WEBHOOK_BATCH_SIZE`), they're sent in a final webhook
-            # call to ensure all tasks are processed and no task is left unreported in the webhook.
-            if len(tasks_for_webhook) >= settings.WEBHOOK_BATCH_SIZE:
-                emit_webhooks_for_instance(
-                    self.project.organization, self.project, WebhookAction.TASKS_CREATED, tasks_for_webhook
-                )
-                tasks_for_webhook = []
+            for task_data in tasks_data:
+                task = self.add_task(task_data, self.project, maximum_annotations, max_inner_id, self, key, link_class)
```
pakelley, line 435: Mind adding a TODO comment to potentially make an `add_tasks` function so we can batch these?

label_studio/io_storages/azure_blob/models.py
matt-bernstein: Check out the existing implementation for s3 - some storages raise an error here and some don't, I chose to add the same error to all of them to keep it consistent. I could be missing some storage-specific logic outside this function that means they're handled differently though.

pakelley (approved):
I don't see anything blocking, lgtm

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14720984501)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14722513446)

label_studio/io_storages/base_models.py
```diff
@@ -429,26 +429,29 @@ def _scan_and_create_links(self, link_class):
...
+                # add task to webhook list
+                tasks_for_webhook.append(task)
+
+                # settings.WEBHOOK_BATCH_SIZE
+                # `WEBHOOK_BATCH_SIZE` sets the maximum number of tasks sent in a single webhook call, ensuring manageable payload sizes.
+                # When `tasks_for_webhook` accumulates tasks equal to/exceeding `WEBHOOK_BATCH_SIZE`, they're sent in a webhook via
+                # `emit_webhooks_for_instance`, and `tasks_for_webhook` is cleared for new tasks.
+                # If tasks remain in `tasks_for_webhook` at process end (less than `WEBHOOK_BATCH_SIZE`), they're sent in a final webhook
+                # call to ensure all tasks are processed and no task is left unreported in the webhook.
+                if len(tasks_for_webhook) >= settings.WEBHOOK_BATCH_SIZE:
```
hakan458, line 450: This could potentially be huge and I dont know what our limitations are for this webhook but I figure we should batch these to WEBHOOK_BATCH_SIZE as to not send thousands of tasks at once

hakan458: Oh wait I guess we append one task at a time to the list right, we dont add the entire list of tasks at once, so all good

matt-bernstein: yep, this block is just moved inside a new loop, logic stays the same

hakan458 (approved):

PR #7342, opened 2025-04-08
feat: DIA-2148: write exports using thread pool

pakelley:
-------

Optimize export job by threading writes to storage


label_studio/io_storages/base_models.py
```diff
@@ -551,15 +554,16 @@ def save_all_annotations(self):
...
-            annotation.cached_user = self.cached_user
-            self.save_annotation(annotation)
+        max_workers = min(32, (os.cpu_count() or 4) * 4)
 
-            # update progress counters
-            annotation_exported += 1
-            self.info_update_progress(last_sync_count=annotation_exported, total_annotations=total_annotations)
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = []
+            for annotation in Annotation.objects.filter(project=self.project).iterator(chunk_size=settings.STORAGE_EXPORT_CHUNK_SIZE):
```
pakelley: notably, we have to submit all annots here before we update any progress, so I'm going to try either updating the counts in the thread or batching the updates (just using the same batch size that we use to fetch from the db, defaults to 100)

label_studio/io_storages/base_models.py
```diff
@@ -551,15 +554,16 @@ def save_all_annotations(self):
         self.info_set_in_progress()
         self.cached_user = self.project.organization.created_by
 
-        for annotation in Annotation.objects.filter(project=self.project).iterator(
-            chunk_size=settings.STORAGE_EXPORT_CHUNK_SIZE
-        ):
-            annotation.cached_user = self.cached_user
-            self.save_annotation(annotation)
+        max_workers = min(32, (os.cpu_count() or 4) * 4)
```
matt-bernstein: Do you see a need to let storage instances override this? (for different cloud providers, or in the future, provider-specific handling of max concurrent writes based on storage details) If so, it can be `self.max_workers`

matt-bernstein (approved):

pakelley:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14504694094)

PR #7354, opened 2025-04-08
feat: OPTIC-1938: Proxy for storages when presigned urls are off

makseq:
# Storage proxy added

**Our current implementation with presigned urls:**

image

**This PR adds proxying from storages instead of redirecting to presigned urls:**

image

# Minor fixes

Fixed race condition on Project => Settings => Cloud Storages page when project=undefined on the first page loading. 

image


farioas:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14364600032)

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14396934665)

bmartel (approved):
Looks great! The one thing I think that remains is how we align the browsers ability to cache the requests. One part of this whole solution was a special ServiceWorker that caches the resolution of the URI's so that the browser can effectively cache the requests when able, like with images for example. This saves on bandwidth significantly in presigned url use cases with normal usage within DataManager -> QuickView. The ServiceWorker only acted on url's that had the word `presign` in it, but now I see the endpoint is `resolve` instead. This will mean that the previous behaviour will not work completely, so we will have to update the ServiceWorker accordingly as well. I can make that change on this PR.

makseq:
@bmartel great point! Thank you for the fix! `/resolve` now makes more sense. I kept `/presign` for backward compatibility.

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14439543164)

makseq:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14457159024)
> **Error**: Workflow failed

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14462835950)

nikitabelonogov:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14464981467)

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,178 @@
...
+            if not storage:
+                logger.error(f'Could not find storage for URI {fileuri}')
+                return Response(status=status.HTTP_404_NOT_FOUND)
+            # Not all storages support presigned URLs
+            if not hasattr(storage, 'presign'):
+                logger.error(f'Storage {storage} does not support presign URLs')
+                return Response(status=status.HTTP_404_NOT_FOUND)
+            presign = storage.presign
+        else:
+            presign = True
```
hakan458, line 55: why do we default to true here?

hakan458: I guess so that if the FF is not set, we default to the old behavior, correct?

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,178 @@
...
+    def proxy_data_from_storage(self, request, uri, project, storage):
+        """Proxy the data directly from storage without redirecting"""
+        try:
+            # Use the storage-specific method to get data stream and content type
+            data, content_type = storage.get_bytes_stream(uri)
+
+            # If we have the data and content type, return the response
+            if data is not None:
+                content_type = content_type or 'application/octet-stream'
+                response = RangedFileResponse(request, data, content_type=content_type)
```
hakan458: What is RangedFileResponse and why do we need it?

makseq:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14489597608)
> **Error**: Workflow failed

makseq:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14489670052)

label_studio/io_storages/proxy_api.py
makseq: Yes, you are right, it's default for the old behavior.

label_studio/io_storages/proxy_api.py
makseq: It's a standard response that supports streaming. We use it here for uploaded files and more - https://github.com/search?q=repo%3AHumanSignal%2Flabel-studio%20RangedFileResponse&type=code

makseq:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14492054959)

hakan458 (approved):

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,327 @@
...
+        deadline = start_time + timeout
+        chunks_yielded = 0
+        total_bytes = 0
+        
+        # Flag to indicate if timeout occurred
+        timed_out = False
+        
+        try:
+            for chunk in stream_body.iter_chunks(chunk_size=chunk_size):
+                print(f"==> chunk: {chunks_yielded}, {total_bytes}\n")
```
jombooth: If this is intended for the final version, better to use `logger.info` than `print`

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,327 @@
...
+        try:
+            # Process and limit the range header for downloaded files
+            range_header = self.override_range_header(request)
+            
+            # Use the storage-specific method to get data stream and content type
+            stream, content_type, metadata = storage.get_direct_stream(uri, range_header=range_header)
+            
+            if stream is None:
+                logger.error(f'Failed to get direct stream from storage {storage}')
+                return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
jombooth: I know we do this in a few places, but it seems a bit iffy to me whether we should be intentionally returning 500s from the API; usually these indicate errors that we did not anticipate. How would https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/424 sound to you?

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,327 @@
...
+            
+            # Create time-limited stream
+            time_limited_stream = self.time_limited_chunker(stream)
+            
+            # Set up streaming response with S3's status code
+            status_code = metadata.get('StatusCode', 200)
```
jombooth: Should it be an error if `StatusCode` wasn't in metadata? Also, is this (as the comment suggests) just an S3 thing, or does it apply to other storage backends?

label_studio/io_storages/proxy_api.py
```diff
@@ -0,0 +1,327 @@
...
+            logger.debug(f">> stream > end: {int(end or 0)/1024/1024} MB")
+            logger.debug(f">> stream > range_header: {range_header}")
+            
+        return range_header
+
+    def proxy_data_from_storage(self, request, uri, project, storage):
+        """
+        Proxy the data using iter_chunks directly from storage streaming object.
+        
+        This implementation forwards Range headers to S3 and streams the response
```
jombooth: same question about S3 here, is this S3 specific?

jombooth:
overall LG (and very impressive work), biggest concerns are whether we've introduced S3 specific details to code that's used for other storages as well, as in https://github.com/HumanSignal/label-studio/pull/7354#discussion_r2049528505

label_studio/io_storages/proxy_api.py
makseq: fixed

label_studio/io_storages/proxy_api.py
makseq: fixed. it's for any storages

label_studio/io_storages/proxy_api.py
makseq: Good questions. It should be applicable to all storages. I will fix it

makseq:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14550330038)

PR #7372, opened 2025-04-15
feat: DIA-2062: Batch writes for cloud storage import

hakan458:
Changes how cloud storage import works so that we are bulk_create-ing Tasks, Annotations, and Predictions 
Before we were calling add_task for every single new key (path) from cloud storage. Now we will create tasks in batches (setting controlled to 50 currently)


label_studio/io_storages/all_api.py
```diff
@@ -127,6 +127,7 @@ def _get_response(self, api, request, *args, **kwargs):
             response = view(request._request, *args, **kwargs)
             payload = response.data
             if not isinstance(payload, list):
+                print(f'{payload=}')
```
matt-bernstein: remove print

label_studio/io_storages/base_models.py
```diff
@@ -398,30 +423,37 @@ def _scan_and_create_links(self, link_class):
...
+                self.info_update_progress(last_sync_count=tasks_created, tasks_existed=tasks_existed)
 
-            # settings.WEBHOOK_BATCH_SIZE
-            # `WEBHOOK_BATCH_SIZE` sets the maximum number of tasks sent in a single webhook call, ensuring manageable payload sizes.
-            # When `tasks_for_webhook` accumulates tasks equal to/exceeding `WEBHOOK_BATCH_SIZE`, they're sent in a webhook via
-            # `emit_webhooks_for_instance`, and `tasks_for_webhook` is cleared for new tasks.
-            # If tasks remain in `tasks_for_webhook` at process end (less than `WEBHOOK_BATCH_SIZE`), they're sent in a final webhook
-            # call to ensure all tasks are processed and no task is left unreported in the webhook.
-            if len(tasks_for_webhook) >= settings.WEBHOOK_BATCH_SIZE:
+                tasks_for_webhook.extend(created_tasks)
```
matt-bernstein, line 439: Have you checked what WEBHOOK_BATCH_SIZE usually is, and if the batch size mismatch between this and CLOUD_IMPORT_TASK_BATCH_SIZE would cause any problems? Rn the comment says webhooks can be sent when num tasks >= this size, but the code will always send when num_tasks == this size, so I wonder if we depend on that anywhere. Also, the pathological case is something like, import batch size is 10000 and webhook batch size is 10, which'd significantly change behavior, so maybe check for import batch size > webhook batch size and logger.warning it?

hakan458: Its 100. To prevent it from getting too large. Now we will just use CLOUD_IMPORT_TASK_BATCH_SIZE which is 50, so not a problem. Dont foresee CLOUD_IMPORT_TASK_BATCH_SIZE ever getting too large here as we want to keep the size of the batch relatively small (the json tasks we are importing could be large in size)

matt-bernstein (approved):

hakan458:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14539306273)

makseq (changes requested):
I believe we should add FF and keep the old implementation in place until we are completely sure that the new one is safe

robot-ci-heartex:

robot-ci-heartex:

PR #7374, opened 2025-04-15
fix: LEAP-1945: [FE][lso] Fix Babel has inefficient RexExp complexity

nass600:
Fixes [CVE-2025-27789](https://nvd.nist.gov/vuln/detail/CVE-2025-27789) security vulnerability

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -107,7 +107,7 @@
     "zod": "^3.22.4"
   },
   "devDependencies": {
-    "@babel/core": "^7.23.2",
+    "@babel/core": "^7.26.10",
     "@babel/plugin-proposal-class-properties": "^7.12.13",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.12.13",
     "@babel/plugin-proposal-optional-chaining": "^7.12.17",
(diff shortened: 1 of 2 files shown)


web/package.json
```diff
@@ -249,7 +249,9 @@
     "xml2js": "0.5.0",
     "jsonwebtoken": "9.0.0",
     "moment": "2.29.4",
-    "qs": "6.5.3"
+    "qs": "6.5.3",
+    "@babel/helpers": "^7.26.10",
+    "@babel/runtime": "^7.26.10"
```
hlomzik, line 254: why do we need them here?

hlomzik: also these two libs should be placed in a correct order, otherwise any other `yarn install` will reorder them producing unnecessary diff in unrelated PR

hlomzik (approved):

web/package.json
nass600: They are needed because they are transitive dependencies so, if not forced, the vulnerability is not fixed given any package using a lower version will still be using it

yyassi-heartex (approved):

bmartel (approved):

web/package.json
hlomzik: thanks for explanation! what about order?

nass600:
/selenium test
> See '/help'

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14474207733)

PR #7376, opened 2025-04-15
feat: OPTIC-2125: Add sync between TimeSeries, Video, Audio

### LINKED ISSUE #5359: Video/audio sync doesn't work
I am trying to do Video Timeline Segmentation in LSF, I have an issue with the audio not syncing to the video, I can play each of them separately but each play button doesn't start both video and audio together.
This is the config I am using:
```
 <View>
            <Header>Video timeline segmentation</Header>
            <Video name="video" value="$video_url" sync="audio"/>
            <Labels name="videoLabels" toName="audio" choice="multiple">
                <Label value="Moving" />
                <Label value="Not moving" />
            </Labels>
            <Audio name="audio" value="$video_url" sync="video"/>
 </View>
```
I tried with several different videos, all have the same issue.
I also tried different config formats, with different decoder/player combinations for the audio.
Here is an example video I tried to use: https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/VolkswagenGTIReview.mp4

My LS version is 1.10.1, I am running it on Windows 11, using Chrome.
This is a warning that I see in the console logs:
```
Waveform.js:320 The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page. https://goo.gl/7K7WLu
value @ wavesurfer.js:5820
i @ wavesurfer.js:5718
(anonymous) @ wavesurfer.js:1846
i @ wavesurfer.js:1832
value @ wavesurfer.js:4390
value @ wavesurfer.js:4174
value @ wavesurfer.js:5582
componentDidMount @ Waveform.js:320
vl @ react-dom.production.min.js:220
Ts @ react-dom.production.min.js:259
t.unstable_runWithPriority @ scheduler.production.min.js:18
Fo @ react-dom.production.min.js:122
Ms @ react-dom.production.min.js:252
bs @ react-dom.production.min.js:243
ps @ react-dom.production.min.js:237
$s @ react-dom.production.min.js:284
(anonymous) @ react-dom.production.min.js:289
xs @ react-dom.production.min.js:244
iu @ react-dom.production.min.js:289
t.render @ react-dom.production.min.js:296
createApp @ LabelStudio.js:62
await in createApp (async)
qoe @ LabelStudio.js:31
(anonymous) @ LabelStudioComponent.js:13
commitHookEffectListMount @ react-dom.development.js:23150
commitPassiveMountOnFiber @ react-dom.development.js:24926
commitPassiveMountEffects_complete @ react-dom.development.js:24891
commitPassiveMountEffects_begin @ react-dom.development.js:24878
commitPassiveMountEffects @ react-dom.development.js:24866
flushPassiveEffectsImpl @ react-dom.development.js:27039
flushPassiveEffects @ react-dom.development.js:26984
(anonymous) @ react-dom.development.js:26769
workLoop @ scheduler.development.js:266
flushWork @ scheduler.development.js:239
performWorkUntilDeadline @ scheduler.development.js:533
```
I don't have any browser extensions active.

cloudmark:
# TimeSeries Synchronization

https://github.com/user-attachments/assets/79c18b32-93ae-4e5e-b702-57e857f8f2fa

## Overview
TimeSeries can synchronize with Audio and Video components, allowing coordinated playback and seeking across different media types.

## Time Units
- All sync messages use relative seconds from the start of each component
- TimeSeries converts these relative seconds based on its configuration:
  - Time-based: Converts to absolute timestamps
  - Index-based: Uses relative seconds as indices

## Configuration
1. Time-based TimeSeries:
   ```xml
   <TimeSeries name="ts" timeColumn="timestamp" timeFormat="%Y-%m-%d %H:%M:%S">
   ```

2. Index-based TimeSeries:
   ```xml
   <TimeSeries name="ts">
   ```

## Synchronization Behavior

1. **Time-based TimeSeries:**
   - Converts relative seconds to absolute timestamps
   - Maintains precise temporal alignment with media, any custom sampling rate
   - Example: 30s in video = 30s from start in TimeSeries
   - If the TimeSeries starts at the 5th second, it will serve as the starting point for other media. Therefore, the 0 second of the video will begin at the 5th second of the TimeSeries, and this offset will always be applied.

2. **Index-based TimeSeries:**
   - Uses direct indices as relative seconds
   - Each second in media = one index in TimeSeries, sampling rate is 1 Hz always
   - Example: 30s in video = index 30 in TimeSeries
   - Suitable mostly for debugging and tests

## Length Mismatches
- Sync works up to the length of the shorter component
- Components stop at their respective ends
- No data loss occurs, but sync stops at the shorter component's end

## Best Practices
1. Use time-based TimeSeries when:
   - Data has actual timestamps
   - Precise temporal alignment is needed
   - Working with multiple media types

2. Use index-based TimeSeries when:
   - Data is sequential
   - No actual timestamps are available
   - Simple 1 sample <=> 1 second mapping with media time is sufficient

# Configurations

# Index based TimeSeries + Video + TimeSeriesLabels

* Index-based TimeSeries (no timestamps at X axis)
* _Note_: One value equals one second because the time axis is not specified in the TimeSeries tag. The video is synced with this idea — one sample equals one second if timestamps are not provided.

```
<View>
  <Video name="video" value="$video" sync="group_a"/>
  
  <TimeSeries name="timeseries"
              value="$ts" valueType="json"
              sync="group_a" sep=","
              overviewWidth="10%"
              >
    <Channel column="value" strokeColor="#FF0000"/>
    <Channel column="value" strokeColor="#00FF00"/>
  </TimeSeries>
  
  <TimeSeriesLabels name="labels" toName="timeseries">
    <Label value="action"/>
    <Label value="pause"/>
  </TimeSeriesLabels>

</View>

<!-- {
  "video": "/static/samples/opossum_snow.mp4",
  "ts": {
      "value": [
        10.7036820361892644,
        -0.18120536109567212,
        -0.39251488391214157,
        1.3384817293995075,
        0.8779675446349394,
        -0.1511946071051955,
        -0.7955547028255082,
        1.0736798948078534,
        1.1266164855584428,
        -0.440291574562604,
        -0.8436786901744359,
        -0.24956239687939094,
        1.268049926141147,
        0.6300808834120004,
        1.7946935071842107,
        -0.37700464705843,
        0.706518542026297,
        -0.45787451607104046,
        -2.3643354623876607,
        0.13984274721398307,
        0.3174445171792305,
        -1.8162371732091722,
        -0.30289394872251374,
        -0.730112449190387,
        -1.6852497246079239,
        -1.0473893262227658,
        0.10416951356137397,
        -2.0266185534759633,
        -0.05196549263706541,
        0.4436085233243668,
        -0.0956064205420074,
        -1.1790065141112944,
        -0.015063840978932763,
        0.28691755509866407,
        1.4122332721986657,
        0.40127732957527523,
        1.546243544663401,
        0.11119508061291504,
        -0.499517691828469,
        -0.02922576888373752,
        -0.8454178734108769,
        0.19122400060485445,
        0.6914340334390281,
        -0.18047241277757645,
        -0.6394589243120249,
        1.0019886671810008
      ]
  }
} -->
```

## Time-based multiple TimeSeries + Audio + Video + TimeSeriesLabels

* Time-based TimeSeries
* TimeSeries, Audio and Video are synced together
* Choices and timeline labels are used as control tags for labeling

```
<View>
  <Video name="video" value="$video" sync="group_a"/>
  <Audio name="audio" value="$video" sync="group_a"/>
  
  <TimeSeriesLabels name="timelinelabels" toName="accel_timeseries">
    <Label value="A"/>
    <Label value="B"/>
  </TimeSeriesLabels>	

  <TimeSeries 
              name="accel_timeseries"
              value="$accel_data"
              sync="group_a"
              timeColumn="time"
              timeFormat="%H:%M:%S.%f"
              timeDisplayFormat="%H:%M:%S.%f"
              overviewWidth="10%"
  >
    <Channel column="accel_x" strokeColor="#FF0000" height="100"/>
    <Channel column="accel_y" strokeColor="#00FF00" height="100"/>
  </TimeSeries>
  
  <TimeSeries 
              name="gyro_timeseries" 
              value="$gyro_data" 
              sync="group_a"
              timeColumn="time"
              timeFormat="%H:%M:%S.%f"
              timeDisplayFormat="%H:%M:%S.%f"
              overviewWidth="10%"
  >
    <Channel column="gyro_x" strokeColor="#0000FF" height="100"/>
    <Channel column="gyro_y" strokeColor="#FF00FF" height="100"/>
  </TimeSeries> 

</View>

<!-- {
  "video": "/static/samples/opossum_snow.mp4",
  "accel_data": "/samples/time-series.csv?time=time&values=accel_x%2Caccel_y&sep=%2C&tf=%H:%m:%d.%f",
  "gyro_data": "/samples/time-series.csv?time=time&values=gyro_x%2Cgyro_y&sep=%2C&tf=%H:%m:%d.%f"
}
-->
```

**Example for time-series-accel.csv for accel_x, accel_y**
```
time,accel_x,accel_y
00:01:01.000000,-0.056646571671882806,2.1066649495524605
00:01:02.000000,-0.6888765232989033,0.35646668995794306
00:01:03.000000,-0.23512086306647553,0.5799351613084716
00:01:04.000000,-0.9314772647682944,-0.5195693066279311
00:01:05.000000,1.321119143958512,-0.622026749003922
00:01:06.000000,0.10592100887528152,0.15477501359739493
00:01:07.000000,-0.6261150686384155,0.5624264458111049
00:01:08.000000,1.0829322997587332,-1.9590268928992862
00:01:09.000000,-1.2267135177322928,-0.4538764395229617
00:01:10.000000,1.6705781810127622,0.38407182850093363
```

**Example for time-series-gyro.csv for gyro_x, gyro_y**
```
time,gyro_x,gyro_y
00:01:01.000000,-0.776563940219835,-1.1115451852904443
00:01:02.000000,0.17111212343134966,-1.377696478819913
00:01:03.000000,-1.168085910547026,-0.8500307427257534
00:01:04.000000,-0.13947878605597916,0.9062482653127198
00:01:05.000000,0.3079887618179474,-1.6722497873634719
00:01:06.000000,-0.3825838786476411,-1.242585234780504
00:01:07.000000,-0.7015245817392025,-1.712515499827561
00:01:08.000000,-0.3437952109000775,-0.9337512501019165
00:01:09.000000,-0.19464021971045084,-0.9653381620475747
00:01:10.000000,-0.29753925483100785,-0.7699832734123578
```

# Advanced Information

Fixes #4892: How can we label video and time series signals while watching them together?
Fixes #5359: Video/audio sync doesn't work

- Add sync attribute to TimeSeries tag model to enable group-based synchronization
- Compose TimeSeries with SyncableMixin for event broadcasting and handling
- Implement sync event handlers for seek, play, and pause events
- Add smooth view updates during audio/video playback
- Prevent event flooding by suppressing sync events during playback
- Maintain view window size during playback
- Map media timeline to TimeSeries range (start/end points)

This enables synchronized playback and seeking between Video, Audio, and TimeSeries components when they share the same sync group. Multiple TimeSeries components can also sync with each other, allowing for coordinated view updates across different time series data. For example:

**Feature flag**
```
fflag_feat_optic_2125_timeseries_sync=1 label-studio
```


cloudmark:
@bmartel @makseq updated to include the feature flag.  Read for review.

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15208746807)

web/libs/editor/src/tags/object/Video/Video.js
```diff
@@ -184,28 +187,40 @@ const Model = types
     },
 
     handleSync(data) {
+      if (data.initiator === self.name) return;
```
hlomzik: Hmmm, this should be handled by SyncableMixin itself. Is it broken?

https://github.com/HumanSignal/label-studio/blob/2660ec9857597f35130ce0e1e7e2ccde225cda4f/web/libs/editor/src/mixins/Syncable.ts#L71-L72

web/libs/editor/src/tags/object/Video/Video.js
```diff
@@ -184,28 +187,40 @@ const Model = types
...
       if (!self.ref.current) return;
+      self.isProcessingIncomingSync = true;
```
hlomzik: This should also be theoretically handled by mixin but this was never tested on TimeSeries of course. Which issues are there?

web/libs/editor/src/tags/object/Video/Video.js
```diff
@@ -184,28 +187,40 @@ const Model = types
...
+      try {
+        const video = self.ref.current;
 
-      const video = self.ref.current;
+        if (isDefined(data.playing)) {
```
hlomzik: Do we need this additional check? both `null` and `undefined` will fail on `if()`

hlomzik:
This is awesome job, @cloudmark! I tried it and was impressed, that's exactly why SyncableMixin was introduced, to allow improvements like this! 

- I see a lot of `isAlive()` checks, what's the reason for such a precaution? Can we fix the original issue when we are trying render or interact with TimeSeries when they are detached?
- What are you fixing in `brushstarted()`? https://github.com/HumanSignal/label-studio/pull/7376/files?diff=unified&w=1#diff-31822f7cd5c02fc2986d0c6fb4642f4a350a338699574efba86e68e5ad984b39R909
- Also I added some questions about how original `SyncableMixin` worked and your fixes around it

Besides this it's awesome! Looking forward for answers, thank you!

hlomzik:
Also we have to document how does it sync with Audio and Video. I can write a doc, but I need a guidance.

As I understand time from media works as relative delta for TimeSeries if it's time based, right? Questions:
- What if it's indices-based?
- What if lengths are different?

web/libs/editor/src/tags/object/Video/Video.js
cloudmark: Yes, this is required. Without this, components will begin receiving their own messages, creating noise and glitches. You can test this by adding the following

```
for (const target of this.syncTargets.values()) {
    if (origin !== target.name) {
        target.syncReceive(data, event);
    } else {
        console.warn(`[WARNING] [Syncable] Sync target sending message to origin ${target.name}, -> ${origin}`)
    }
}
```

- Audio play, pause, or seek will show this warning.  
- Video play, pause, or seek will show this warning.  
- Timeline seek will show warnings.  
From what you are saying, it is clear that the Syncable is not working as it should. It’s hard for me to fix this without fully understanding the semantics you wish to achieve. There do not seem to be tests around this either. I think it's best if we leave this in place and contact the author to investigate the cause. The fix should then remove this part once we have tested it.  

See video: https://www.loom.com/share/d5c10858c8df4c79bfa9c91baecdb04f

web/libs/editor/src/tags/object/Video/Video.js
cloudmark: The mixin does not handle this; I added some defensive coding because I was receiving numerous messages from within the component itself. Clearly, the Syncable is not functioning as it should.

web/libs/editor/src/tags/object/Video/Video.js
cloudmark: No will remove

cloudmark:
> This is awesome job, @cloudmark! I tried it and was impressed, that's exactly why SyncableMixin was introduced, to allow improvements like this!
> 
> * I see a lot of `isAlive()` checks, what's the reason for such a precaution? Can we fix the original issue when we are trying render or interact with TimeSeries when they are detached?
> * What are you fixing in `brushstarted()`? https://github.com/HumanSignal/label-studio/pull/7376/files?diff=unified&w=1#diff-31822f7cd5c02fc2986d0c6fb4642f4a350a338699574efba86e68e5ad984b39R909
> * Also I added some questions about how original `SyncableMixin` worked and your fixes around it
> 
> Besides this it's awesome! Looking forward for answers, thank you!

I did the `isAlive()` checks because the component was detaching; I have reverted this now. It could have been when the component was receiving a substantial amount of messages due to the sync bug. I took another look at the `brushstarted()` and will revert this as well. I'm sure there was an edge case I was trying to fix with this; however, I cannot reproduce it. I created a video to show the SyncableMixin bug; this should be tackled separately, I believe, since it’s not an issue with the sync per se.

cloudmark:
> Also we have to document how does it sync with Audio and Video. I can write a doc, but I need a guidance.
> 
> As I understand time from media works as relative delta for TimeSeries if it's time based, right? Questions:
> 
> * What if it's indices-based?
> * What if lengths are different?

_The text was moved to the first PR message_

web/libs/editor/src/tags/object/Video/Video.js
hlomzik: Your snippet shows logging inside SyncableMixin and it will log same names, yes. That's what we are trying to avoid inside mixin and feels like we are doing it well. I added logging to the line we are discussing and it always has different names, so my comment still stands — SyncableMixin does this prevention internally, so we don't need to extend event with `initiator`.

web/libs/editor/src/tags/object/Video/Video.js
hlomzik: The same here. I added logging to line 220, the only place you check `if (self.isProcessingIncomingSync)` and it's always false there because SyncableMixin has the concept of sync window where all extra events are blocked while their bouncing back and forth in media elements.
The only issue here could be is that the size of this window is hardcoded into constant to be ~6 ticks and it might be not enough. But in config with Video+Audio+TS+TS and throttled network I couldn't catch any glitches in sync and any wrong values.
Really curious what could go wrong here...

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15356923309)
> Successfully merged:  create mode 100644 web/libs/editor/src/components/TaskSummary/types.ts

web/libs/editor/src/tags/object/Video/Video.js
hlomzik: After discussion we decided to remove `initiator` and `isProcessingIncomingSync` checks. Right now from all the cases we tested SyncableMixin seemed to do the good job of keeping event flow one-directional. So we removed these checks for now in atomic commits, so they can be reverted if needed.

hlomzik (approved):

Issue #7189, opened 2025-03-09
Usage of `Function` means that Content Security Policy for websites using Label Studio must include 'unsafe-eval'


benglewis:
Currently, since Label Studio is using `Function` in the `utils.js` file for Label Studio's Data Manager's utilities, anyone that wants to use Label Studio must allow 'unsafe-eval' in their Content Security Policy (CSP) on the website which uses Label Studio:

https://github.com/HumanSignal/label-studio/blob/827e3a6edc66b4c3425be1fde5291674808ca732/web/libs/datamanager/src/components/Common/Table/utils.js#L16

I suggest fixing this by using a different implementation.

benglewis:
Hi all,
I've created a Pull Request to fix the issue with the help of Copilot Workspace

heidi-humansignal:
Hi there,

Thanks for posting this, and for opening up the PR to help dive into this further! I'll pass this along to our team for review. We’ll evaluate this option and similar alternatives to see if they meet our performance and compatibility requirements. Your feedback is very valuable, and we’ll keep you updated as we work on this improvement.

Thanks,

Tyler Conlee
Head of Support
HumanSignal

> Comment by Tyler Conlee
> [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/13767995685)

benglewis:
@TylerConlee Is there a particular timeline in place for when this issue/PR may get attention from the development team? This is a pretty bad security flaw to be honest :/

benglewis:
@TylerConlee @bmartel Any ideas regarding this 🙏 I apologize for _nagging_, but we would really appreciate if this can be merged.

BTW, it was great to see the appreciation that you had for @ddishi 's contribution and how you took the change and built upon it! :)

PR #7377, opened 2025-04-15
fix: OPTIC-1966: Address CSP issue by removing unsafe-eval usage

bmartel:
Internal PR for testing workflows, all work and comments shall be addressed in the original #7189


web/libs/datamanager/src/components/Common/Table/utils.js
```diff
@@ -7,15 +7,17 @@ export const prepareColumns = (columns, hidden) => {
...
-    // eslint-disable-next-line no-new-func
-    const fn = new Function("object", `return object${normalizedPath}`);
+    const properties = path.split(".");
+    let result = object;
+
+    for (const property of properties) {
+      result = result?.[property];
+      if (result === undefined) {
+        return undefined;
+      }
```
yyassi-heartex, line 17: bit of a nit-pick but technically this'll do too
```suggestion
      if (!result) {
        return result;
      }
```

yyassi-heartex (approved):
i have a bit of a nitpick since it it'll return 1 step quick in case of boolean false but beyond that it looks good - would be really good to have a unit test for this since it seems very re-usable

PR #7380, opened 2025-04-15
fix: LEAP-1928: support deletions of annotations and tasks in pausing code

jombooth:
Prepwork for LSE change - just start passing Project to the `after_bulk_delete_actions` function which does nothing on LSO.

### DIFF
--- a/label_studio/tests/webhooks/test_webhooks.py
+++ b/label_studio/tests/webhooks/test_webhooks.py
@@ -352,7 +352,7 @@ def test_webhooks_for_action_delete_tasks_annotations(configured_project, busine
 
 # ACTION: DELETE TASKS
 @pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


label_studio/tests/webhooks/test_webhooks.py
```diff
@@ -352,7 +352,7 @@ def test_webhooks_for_action_delete_tasks_annotations(configured_project, busine
 
 # ACTION: DELETE TASKS
 @pytest.mark.django_db
-def test_webhooks_for_action_delete_tasks_annotations(configured_project, business_client, organization_webhook):
+def test_webhooks_for_action_delete_tasks(configured_project, business_client, organization_webhook):
```
jombooth, line 355: drive-by: I noticed there's another test with the same name as the old name (probably copypasta) so this wasn't running.

triklozoid (approved):

Issue #7382, opened 2025-04-16
[Bug] Cache Labels returns blank for Taxonomy Annotations


sbatchelder:
**Describe the bug**
The experimental feature "Cache Labels", which extracts annotations such that they can be interacted with as though they were data fields, does not work with Taxonomy type annotations. Instead of returning one or more values, cache_ Taxonomy column fields are all blank.

**To Reproduce**
1. Set environment flag  `EXPERIMENTAL_FEATURES=1` when running label studio
2. create an annotation template that features a Taxonomy control tag
3. Create and annotate one or more tasks.
4. Select tasks in the Data Manager View
5. Click Actions (a dropdown button with "n Tasks" text, first on the menu bar for the DM tab) and click Cache Labels
6. Select the Control Tag corresponding to the Taxonomy annotation you previously defined
7. Refresh the page. The new cache_ annotation appears, but all entries are blank

**Expected behavior**
The new column should show the value(s) selected from the Taxonomy annotation.

**Environment (please complete the following information):**
 - OS: PopOS, Brave browser
 - Label Studio Version: 1.15.0

heidi-humansignal:
Hello,

Thank you for reporting this issue. The behavior you’re seeing is caused by a limitation in the current implementation of the Cache Labels action when used with Taxonomy annotations. The extraction logic in the cache labels job is designed to work with flat arrays (e.g., lists of strings). However, Taxonomy annotations return the selected values as a nested list (for example, a list of lists), so the current checks do not extract any values—resulting in a blank cached column.

I'll file a bug for engineering team.

Thank you,
Abu

> Comment by Abubakar Saad
> [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/14595993168)

sbatchelder:
Hi @AbubakarSaad, I've made a pull request (#7383) and have been going back n forth with Max a bit on a solution for this issue, though it seems like the conversation has stalled out. Feel free to point the engineering team that way :)

heidi-humansignal:
Hello,

Great, Max will reach out and let you know about the PR! thank you very much with your contribution!

Thank you,
Abu

> Comment by Abubakar Saad
> [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/14652107519)

PR #7383, opened 2025-04-16
feat: Cache Labels for Taxonomy 

sbatchelder:
See issue  #7382 

This fix allows for the value(s) of Taxonomy annotations to be displayed by the experimental Cache Labels feature.
Prior to this PR, Cache Labels would create an all-blank column when applied to a Control Tag for a Taxonomy annotation.
With this PR, values selected in a Taxonomy annotation are correctly displayed.

The issue arose from Taxonomy annotation values being returned as a list, as opposed to a str. For Taxonomy type annotations, a list is required for each selected element so as to capture the hierarchical path taken for any given value in a taxonomy tree. In this PR, consideration was taken so as to only show the selected value, and not capture parts of the path leading up to the selected value. One or more selected values, as well as leaf and intermediate/branch values are supported but this fix. 

image

image

This is a very simple, very small, low risk PR. It is also my first PR with Label Studio, yay!
My thanks to @makseq who pointed me towards the `cache_labels.py` file lines most likely responsible for the blank-taxonomy-cache-labels bug.


sbatchelder:
In discussions with Max, it was determined that displaying just the leaf node of a selected taxonomy was not always desirable, as important information about taxonomic hierarchy paths can be lost. 

I've made an update such that cache_labels will match the formatting schema of the given Taxonomy tag's labeling interface `showFullPath` and `pathSeparator` attributes. This was achieved by instantiating a `label_studio_sdk.label_interface.LabelInterface` object with config of the given project. `label_studio/data_manager/actions/cache_labels.extract_labels(...)` now optionally accepts this label_interface object.

Now `<Taxonomy name="Type" toName="image" showFullPath="true" pathSeparator=".">`  leads to:
image

Defaults are respected, so `<Taxonomy name="Type" toName="image" showFullPath="true">` becomes: 
image

And all defaults, ie a basic `<Taxonomy name="Type" toName="image">` becomes: 
image

Defaults are hardcoded and come from Label Interface [Taxonomy Params](https://labelstud.io/tags/taxonomy#Parameters) . 
If default taxonomy parameters for `showFullPath` and `pathSeparator` are every updated they will need to be manually updated here as well, as I am unaware of where default values for control tags are stored.

If the taxonomy control tag is a Custom control tag, naturally it will not have any Labeling Interface configuration to specify its formatting. In this case, `showFullPath='true'` and `pathSeparator='/'` are used. If a Labeling Interface object is not provided to `extract_labels(...)` (or is None) these defaults are also used. Although `showFullPath='true'` differs from the default for Label Interface Taxonomy Params, I think here it made more sense to err on the side of displaying/transcribing more information (the full path) to the cache_label than not.

label_studio/data_manager/actions/cache_labels.py
```diff
@@ -57,20 +59,35 @@ def cache_labels_job(project, queryset, **kwargs):
...
+                if region['value'][key] and isinstance(region['value'][key], list):
+                    
+                    if key == 'taxonomy':
+                        showFullPath = 'true'
+                        pathSeparator = '/'
+                        if label_interface is not None and label_interface.find_tags('control', match_fn=lambda tag: tag.name==region['from_name']):
+                            # if from_name is not a custom_control tag, then we can try to fetch taxonomy formatting params
+                            showFullPath = label_interface.get_control(region['from_name']).attr.get('showFullPath', 'false')
+                            pathSeparator = label_interface.get_control(region['from_name']).attr.get('pathSeparator', '/')
+
```
makseq: this label_interface.* operations are pretty heavy, because they work with XML parsing each time. I think, we need to cache them

makseq: something like this **draft**: 

```
# Add a cache dictionary outside the loop to store results for each 'from_name'
control_cache = {}

if label_interface is not None:
    # Check if we already cached this control
    if region['from_name'] not in control_cache:
        # Cache the result of find_tags and get_control for this 'from_name'
        if label_interface.find_tags('control', match_fn=lambda tag: tag.name == region['from_name']):
            control = label_interface.get_control(region['from_name'])
            control_cache[region['from_name']] = {
                'showFullPath': control.attr.get('showFullPath', 'false'),
                'pathSeparator': control.attr.get('pathSeparator', '/')
            }
        else:
            # Cache a default value if no control is found
            control_cache[region['from_name']] = {
                'showFullPath': 'false',
                'pathSeparator': '/'
            }
    ...
    # Use the cached values
    showFullPath = control_cache[region['from_name']]['showFullPath']
    pathSeparator = control_cache[region['from_name']]['pathSeparator']

    if showFullPath == 'false':
        for elems in region['value'][key]:
            labels.append(elems[-1])  # Just the leaf node of a taxonomy selection
    else:
        for elems in region['value'][key]:
            labels.append(pathSeparator.join(elems))  # The full delimited taxonomy path
```

makseq (changes requested):

sbatchelder:
new commit: I've pre-loaded the label interface tags prior to the tasks and annotations loops. The project's label interface control tags are easily accessible via dict now, should be orders of magnitude more efficient. 
Also corrected the "delim-" typo :)

label_studio/data_manager/actions/cache_labels.py
```diff
@@ -57,20 +60,36 @@ def cache_labels_job(project, queryset, **kwargs):
...
+                        if showFullPath == 'false':
+                            for elems in region['value'][key]:
+                                labels.append( elems[-1] )  # just the leaf node of a taxonomy selection
+                        else:
+                            for elems in region['value'][key]:
+                                labels.append( pathSeparator.join(elems) )  # the full delimited taxonomy path
+
+                    # other control tag types like Choices & TextAreas
+                    elif isinstance(region['value'][key][0], str):
+                        labels.extend( region['value'][key] )
```
makseq:
```suggestion
                        labels.extend(region['value'][key]) 
```

label_studio/data_manager/actions/cache_labels.py
```diff
@@ -18,7 +19,9 @@ def cache_labels_job(project, queryset, **kwargs):
     source_class = Annotation if source == 'annotations' else Prediction
     control_tag = request_data.get('custom_control_tag') or request_data.get('control_tag')
     with_counters = request_data.get('with_counters', 'Yes').lower() == 'yes'
-
+    label_interface = LabelInterface(project.label_config)
+    label_interface_tags = {tag.name:tag for tag in label_interface.find_tags('control')}
```
makseq: I believe it's better to create on the fly after the line 80:
https://github.com/HumanSignal/label-studio/pull/7383/files#diff-edcc7be940299b140a6cfd30643998168b462e92102a5418d5f570d8babf0a53R80

```
else: 
    from_name = region['from_name']
    label_interface_tags[from_name] = label_interface.get_control(from_name)
```

because, as I wrote earilier:
> Better to call it for every annotation and for every task on the fly and cache it on the fly as well. Your label config can be changed and a part of annotations will stay out of label config. So, the only way to know which from_names you need to process is just to go through all the annotations

makseq (changes requested):

label_studio/data_manager/actions/cache_labels.py
sbatchelder: I don't really understand. The project's label interface configuration could change DURING the label_caching process, granted, but wouldn't the label caching have to be restarted anyways to capture the changes? If it caches on the fly, then different parts of a cached column could have different formatting. That seems worse than a whole-ly out of date formatting. Can you explain? The way I have it right now, all control tags from the label interface configs, incl. everything that a from_name would point to, is cached, once, at the start of the cache_label operation.

If you're concerned about efficiency, then moving it out of the `extract_labels( )` makes the most sense. The efficiency gained from moving it to where you propose is only relevant for when the "ALL" cache label tag is used. For any specific tag, it's only called once, because the potentially large  `for region in annotation.result:` loop is mitigated by the `region['from_name'] == control_tag` clause. Moving it to where I've done means the expensive XML parsing only happens once and not 1 or N (ALL) times per Task per Annotation.

The way I have it you get taxonomy cache_labels columns that reflect the the Label Interface UI with minimal additional calculation overhead. 

It's possible I'm missing something. Why is it important to re-parse the label interface config for every annotation? I'm happy to change my commits if a compelling use case is presented.

makseq: No.. I meant another situation. Imagine this: 

1. You set label config for your project

```
<View> 
  <Text value="$text" name="text"/>
  <Choices name="a" toName="text">
    <Choice value="1"/>
  </Choices>
</View>
```

2. You submitted a few annotations, they include `from_name="a"` regions now. 
3. Then you changed your label config:
```
<View> 
  <Text value="$text" name="text"/>
  <Choices name="b" toName="text">
    <Choice value="x"/>
  </Choices>
</View>
```
4. You submitted a few more annotations. 
5. You called Cache labels for all annotations.

Now, you will see that annotations with Choice "1" are not in the cache columns.

makseq (changes requested):

label_studio/data_manager/actions/cache_labels.py
sbatchelder: Hey @makseq ,

I tried following your steps above to setup the situation you described, but I find myself unable to remove `Choices name="a"` using the UI, after having submitted a couple tasks containing the `name=a` Choices. I get a `Created annotations are incompatible with provided labeling schema, we found: 2 with from_name=a, to_name=image, type=choices` error when attempting to remove the `name=a` Choices field; the new config does not save. Note that I am using an Image data object instead of Text, because that's what I have on-hand for my project but this I'd image shouldn't affect this step.

image

So unfortunately I am unable to test the situation you describe.

That said, I think my implementation still works as intended and I am not convinced the label interface config needs to be rechecked on-the-fly. If an annotation is not present in the `label_interface_tags = {tag.name:tag for tag in label_interface.find_tags('control')}` dict but _is_ present in the annotation results, then currently my code treats the value like a Custom Control Tag and it would get included in a `cache_all` column with default formatting (`showFullPath='true'` and `pathSeparator='/'`). 

Have you tested my code against the situation you describe and found it to fail? If yes, I'd love a snapshot of the issue because I can't reproduce your scenario. 
I am open to stepping through the code with you in a live call to test or discuss if you like.

Also, I'm not sure how the situation you describe above relates to rechecking the label interface config on-the-fly.
Is that still a concern you hold?

sbatchelder:
Hey @makseq ,

I tried following your steps above to setup the situation you [described above](https://github.com/HumanSignal/label-studio/pull/7383#discussion_r2059256618), but I find myself unable to remove `Choices name="a"` using the UI, after having submitted a couple tasks containing the `name=a` Choices. I get a `Created annotations are incompatible with provided labeling schema, we found: 2 with from_name=a, to_name=image, type=choices` error when attempting to remove the `name=a` Choices field; the new config does not save. Note that I am using an Image data object instead of Text, because that's what I have on-hand for my project but this I'd image shouldn't affect this step.

image

So unfortunately I am unable to test the situation you describe.

That said, I think my implementation still works as intended and I am not convinced the label interface config needs to be rechecked on-the-fly. If an annotation is not present in the `label_interface_tags = {tag.name:tag for tag in label_interface.find_tags('control')}` dict but _is_ present in the annotation results, then currently my code treats the value like a Custom Control Tag and it would get included in a `cache_all` column with default formatting (`showFullPath='true'` and `pathSeparator='/'`). 

Have you tested my code against the situation you describe and found it to fail? If yes, I'd love a snapshot of the issue because I can't reproduce your scenario. 
I am open to stepping through the code with you in a live call to test or discuss if you like.

Also, I'm not sure how the situation you describe above relates to rechecking the label interface config on-the-fly.
Is that still a concern you hold?

makseq:
@sbatchelder yes, this situation is not very often and it's a bit tricky to reproduce it. Maybe we can ignore it. 

I am going to run pytests for it tomorrow. If they pass, we can merge it.

makseq:
Could you please merge the latest develop to your branch? It should fix pytests

makseq:
Also, you need to fix Linter. To do this, run:
```
pip install pre-commit
make fmt-all
```

makseq (approved):

sbatchelder:
Aww I'm sad this didn't make it into the recent version release. 
@makseq , thank you for the approval. What are the failing checks about?

PR #7387, opened 2025-04-16
fix: LEAP-1947: Fix interactive view all FF in image tag case

Gondragos:
This pull request introduces changes to enable and integrate the `FF_DEV_3391` feature flag, which enhances the handling of interactive views and image regions in the annotation tool.

- fix interactive view all working with MIG approach
- fix the way Tools Manager works in case of annotation has its own independent tags models tree

### Feature Flag Integration:

* Enabled the `FF_DEV_3391` feature flag in `flags.json` to activate the new interactive view functionality.
* Updated `renderItem` in `Tree.tsx` to include a fallback mechanism (`|| el`) for rendering image regions when the feature flag is active.

### Tool and Manager Enhancements:

* Added feature flag checks in `ToolMixin` to dynamically adjust the `obj` and `control` getters based on the selected annotation. [[1]](diffhunk://#diff-abc71b87397ed7e37a55c3f41193d3421b8a86a5613f1cbed3f381d2dd4ed125R15-R20) [[2]](diffhunk://#diff-abc71b87397ed7e37a55c3f41193d3421b8a86a5613f1cbed3f381d2dd4ed125R29-R37)
* Introduced a `root` getter in `ToolsManager` to provide access to the annotation store's root.

### Annotation and Image Model Updates:

* Enhanced `Annotation.js` to include `imageEntities` in the `updateIds` function for better handling of image-related annotations.
* Modified `Image.js` to:
  - Append annotation IDs to image entity IDs for uniqueness.
  - Skip tool initialization if no annotation is selected (that means that we are in the model tree not related to the annotations at all but to the annotation store)


Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14574417559)
> Successfully merged:  delete mode 100644 web/dist/apps/labelstudio/version.json

web/libs/editor/src/core/feature-flags/flags.json
```diff
@@ -1,7 +1,7 @@
 {
   "ff_front_DEV_1713_audio_ui_150222_short": false,
   "ff_front_dev_2715_audio_3_280722_short": true,
-  "fflag_fix_front_dev_3391_interactive_view_all": false,
+  "fflag_fix_front_dev_3391_interactive_view_all": true,
```
Gondragos:
```suggestion
```

web/libs/editor/src/tags/object/Image/Image.js
```diff
@@ -567,7 +569,11 @@ const Model = types
       };
     },
   }))
-
+  .volatile((self) => {
+    return {
+      manager: null,
+    };
+  })
```
hlomzik: we usually use simpler version
```suggestion
  .volatile((self) => ({
    manager: null,
  }))
```

hlomzik (approved):
Add more comments please around views — what and how are you trying to retrieve.
PR itself seems good, just would be better to fully understand what's going on.

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14613428423)
> Successfully merged:  delete mode 100644 web/libs/ui/src/shad/components/ui/select.tsx

PR #7420, opened 2025-04-24
fix: OPTIC-2016: Removing AntD reset to aid in UI consistency between page navigations

bmartel:
Because AntD does not allow for explicit imports of the components/utilities classes it uses, separate from its reset css, this PR is introducing a postinstall step to strip out the reset css and gitignore that resulting file. This stops the issue of having certain pages impeded by shifting styles like typography and other global considerations which causes unforeseen consistency issues and overall a lot of bugs.


yyassi-heartex (approved):

Dockerfile
farioas, line 1: Please preserve indentation in Dockerfile

farioas (dismissed):

Dockerfile
bmartel: Will do

hlomzik:
@bmartel Could you describe what are you actually doing in this PR, please? I see some script and even a docker changes. What's the reasoning behind this?

bmartel:
@hlomzik the description I wrote is accurately why I'm doing this. AntD has no way to allow you to ignore the reset, so all of these changes are necessary to rip out the reset only code through a postinstall. The resulting file we end up using instead of the original AntD import (which has everything in it) we import this generated file which is AntD without the reset css. I gitignored the file as it is huge and went with a compiled step here.

web/webpack.config.js
```diff
@@ -169,7 +169,7 @@ module.exports = composePlugins(
 
           // we also don't need css modules as these are used directly
           // in the code and don't need prefixing
-          if (testString.match(/module|raw/)) return false;
+          if (testString.match(/module|raw|antd-no-reset/)) return false;
```
bmartel:
```suggestion
          if (testString.match(/module|raw|antd/)) return false;
```

PR #7438, opened 2025-04-29
feat: DIA-2157: Disable "Add Storage" button when storage limit is reached

matt-bernstein:
Forward billing check from LSE in response header for storage list, so the page has access to it


matt-bernstein:
Made necessary changes in LSE only

PR #7444, opened 2025-04-30
fix: DIA-1435: [FE, cust-Comet] Some errors on Prompt task list are not displayed gracefully on UI

nick-skriabin:
This pull request makes a small enhancement to the tooltip styling by limiting its maximum width.

* [`web/libs/ui/src/lib/Tooltip/Tooltip.module.scss`](diffhunk://#diff-1689043134504edaad1d5eb358066805af994052d48d6273eeefa63580a5bca6R21): Added a `max-width` property with a value of 250px to ensure tooltips do not exceed this width, improving readability.


hlomzik (approved):

nass600:
@nick-skriabin @ricardoantoniocm Do we foresee any special case where we would like to override this `max-width`?

If we do I would recommend adding the tailwind class and allow merging from the outside with `twMerge` otherwise looks good

nass600 (approved):

ricardoantoniocm:
> @nick-skriabin @ricardoantoniocm Do we foresee any special case where we would like to override this `max-width`?
> 
> If we do I would recommend adding the tailwind class and allow merging from the outside with `twMerge` otherwise looks good

Thanks for raising this, Nacho. 

Yes, I foresee that. Let's not set a max width in stone by ensuring it can be overwritten if necessary. Also, this particular change could affect how specific tooltips display info across the app. Let's make sure this change doesn't negatively affect the contents of tooltips across the app.

nick-skriabin:
@ricardoantoniocm @nass600 the width is customizable via `style` or `className` properties, so i don't see the problem here

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14830445565)

web/libs/ui/src/lib/Tooltip/Tooltip.module.scss
```diff
@@ -18,6 +18,9 @@
   pointer-events: none;
   background-color: var(--color-neutral-inverted-background);
   border-radius: 3px;
+  white-space: normal;
+  overflow-wrap: break-word;
+  max-width: 250px;
```
yyassi-heartex, line 23: the ticket seems to be prompt task list specific but the change will apply to all tooltips - is this intended? I see class name is being passed in to tooltip now which is great and needed but why a sweeping change to address a specific issue?

MihajloHoma:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14853321127)

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14878760207)

web/libs/ui/src/lib/Tooltip/Tooltip.module.scss
nick-skriabin: while yes, i think the rest of the tooltips will benefit from this change. it's not that big of a change anyway and most tooltips fall under the 250px min width

PR #7452, opened 2025-05-01
feat: OPTIC-752: Filtering deleted users from DM filter

wesleylima:
Delete users all have the same name "Deleted User" so are of very limited usefulness for filtering in data manager. This changes it so that users named "Deleted User" do not appear as a user that can be filtered by in Data Manager

### DIFF
--- a/web/libs/datamanager/src/components/Filters/types/List.jsx
+++ b/web/libs/datamanager/src/components/Filters/types/List.jsx
@@ -2,7 +2,7 @@ import { observer } from "mobx-react";
 import { FilterDropdown } from "../FilterDropdown";
 // import { Common } from "./Common";
 
-function filterItems(items) {
+function defaultFilterItems(items) {
   return items?.toJSON ? items.toJSON() : items;
 }
 
@@ -17,7 +17,7 @@ export const VariantSelect = observer(({ filter, schema, onChange, multiple, val
     return Array.isArray(value) ? value : (value ?? []);
   })();
 
-  const filterItems = filter.cellView?.filterItems || filterItems;
+  const filterItems = filter.cellView?.filterItems || defaultFilterItems;
   const FilterItem = filter.cellView?.FilterItem;
   return (
     <FilterDropdown
--- a/web/libs/datamanager/src/components/Filters/types/List.jsx
+++ b/web/libs/datamanager/src/components/Filters/types/List.jsx
@@ -1,13 +1,14 @@
 import { observer } from "mobx-react";
 import { FilterDropdown } from "../FilterDropdown";
-import { useSDK } from "../../../providers/SDKProvider";
-import { useMemo } from "react";
 // import { Common } from "./Common";
+
+function filterItems(items) {
+  return items?.toJSON ? items.toJSON() : items;
+}
 
 export const VariantSelect = observer(({ filter, schema, onChange, multiple, value, placeholder }) => {
   if (!schema) return <></>;
   const { items } = schema;
-  const sdk = useSDK();
 
   const selectedValue = (() => {
     if (!multiple) {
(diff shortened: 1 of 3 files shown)


web/libs/datamanager/src/components/Filters/types/List.jsx
```diff
@@ -14,9 +16,21 @@ export const VariantSelect = observer(({ filter, schema, onChange, multiple, val
...
+  // Filter out users with firstName "Deleted" for annotator filters
+  const filteredItems =
+    filter.field?.alias === "annotators" || filter.field?.alias === "reviewers"
+      ? (items?.toJSON ? items.toJSON() : items).filter((item) => {
+          const user = sdk.store.users.find((u) => u.id === item);
+          return !(user?.firstName === "Deleted" && user?.lastName === "User");
+        })
+      : items?.toJSON
+        ? items.toJSON()
+        : items;
```
yyassi-heartex: Would be great to memoize this using `useMemo` will need to add `filter.field`, `items` and `sdk.store.users` as dependencies.
This is beneficial since the filter through items will get worse with the length of that list and if we can reduce the number of times we do that, it would be great

yyassi-heartex: also Im guessing we do want to display deleted items else where - this is why we're not changing the request to only get non-deleted users from the backend?

bmartel: We have to still be able to pull the user record for purposes of displaying tasks annotators, reviewers, but searching for deleted users in the system is not feasible, since they are soft deleted and only kept for the purpose of not removing data produced by them in a given org's project. To be quite honest, I don't really see the need for soft deletion fwiw, we could have possibly just used a a simple method of marking user records in these selects with their role/status text like `First Last (Deactivated)` or `example@email.org (Inactive)` etc.

Filters with the ability to search the list plus better identification in the rendered list items would have been my pick for dealing with the cases of people that are not currently, or are no longer active in an org.

wesleylima: Added memoization!

Yes deleted users should still appear in other parts of datamanager like the annotator icons in the DM list. This is why we're only filtering here: to make sure they don't show up in the DM filters

yyassi-heartex (approved):

web/libs/datamanager/src/components/Filters/types/List.jsx
```diff
@@ -14,9 +17,21 @@ export const VariantSelect = observer(({ filter, schema, onChange, multiple, val
   })();
 
   const FilterItem = filter.cellView?.FilterItem;
+
+  const filteredItems = useMemo(() => {
+    if (filter.field?.alias === "annotators" || filter.field?.alias === "reviewers") {
+      const itemsArray = items?.toJSON ? items.toJSON() : items;
+      return itemsArray?.filter((item) => {
+        const user = sdk.store.users.find((u) => u.id === item);
```
bmartel: Is `item` guaranteed to always be an id, also does `sdk.store.users` already have the full list possible for the org prior to this dropdown being accessed?

Next thought, in the cases of a very large number of users, this is running possibly `n^2` or `mXn` depending on whether the itemsArray is the full list of user ids or not.

bmartel: Also, this all feels like something that should be lifted out of here and closer to the implementation point where it is needed, this code looks very agnostic to record types.

bmartel (approved):

web/libs/datamanager/src/components/Filters/types/List.jsx
```diff
@@ -2,6 +2,10 @@ import { observer } from "mobx-react";
 import { FilterDropdown } from "../FilterDropdown";
 // import { Common } from "./Common";
 
+function filterItems(items) {
```
bmartel:
```suggestion
function defaultFilterItems(items) {
```

web/libs/datamanager/src/components/Filters/types/List.jsx
```diff
@@ -13,10 +17,11 @@ export const VariantSelect = observer(({ filter, schema, onChange, multiple, val
     return Array.isArray(value) ? value : (value ?? []);
   })();
 
+  const filterItems = filter.cellView?.filterItems || filterItems;
```
bmartel:
```suggestion
  const filterItems = filter.cellView?.filterItems || defaultFilterItems;
```

PR #7453, opened 2025-05-01
fix: LEAP-2063: LEAP-2062: No more annotations message is displayed in Review Stream when Interactive View All FF is turned ON

nass600:
Having the `Interactive View All` flag enabled is generating extra renders in the tree with some undefined data which is throwing uncontrolled errors hence the app is behaving erratically given we are accessing a detached state tree's settings.

It should fix:
- https://humansignal.atlassian.net/browse/LEAP-2063
- https://humansignal.atlassian.net/browse/LEAP-2062


web/libs/editor/src/mixins/AnnotationMixin.js
```diff
@@ -23,18 +26,9 @@ export const AnnotationMixin = types.model("AnnotationMixin", {}).views((self) =
...
         return null;
       }
-
-      // if annotation history item selected
-      if (root.annotationStore?.selectedHistory) {
-        return root.annotationStore.selectedHistory;
-      }
-
-      // return connected annotation, not the globally selected one
-      return Types.getParentOfTypeString(self, "Annotation");
```
hlomzik: That is the crucial part of feature flagged code brand, we should not remove it just like this.

hlomzik (changes requested):

hlomzik (approved):

AndrejOros:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14831533545)

PR #7455, opened 2025-05-01
fix: Audio warnings when scroll and zoom

makseq:


cloudmark:
@makseq we should use https://github.com/HumanSignal/label-studio/pull/7457, this does not solve it completely.  Do not merge.

PR #7458, opened 2025-05-01
feat: DIA-2202: Support imported tasks that point to different buckets

hakan458:
This use case was not supported before:

```
Import JSON tasks from bucket A
Inside these tasks, there are references to files in bucket B
```

Now instead of relying on Task.storage (the storage that the Task was imported from), we check all storages in the project, to see if the scheme (e.g. `s3`) PLUS the bucket name (e.g. `my-bucket-b`) match the task data's URI. If a matching storage is found, use that to resolve the URI.


matt-bernstein (approved):

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14841050431)
> Successfully merged:  create mode 100644 web/tools/extract-antd-no-reset/package.json

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14841120955)

label_studio/io_storages/base_models.py
```diff
@@ -255,8 +255,17 @@ def can_resolve_scheme(self, url: Union[str, None]) -> bool:
...
-        if prefix == self.url_scheme:
-            return True
+        bucket_uri = parse_bucket_uri(url, self)
+
+        # If there is a prefix and the bucket matches the storage's bucket/container/path
+        if prefix == self.url_scheme and bucket_uri:
+            # Check for different storage types
+            if hasattr(self, 'bucket') and bucket_uri.bucket == self.bucket:
+                return True
+            if hasattr(self, 'container') and bucket_uri.bucket == self.container:
```
makseq, line 266: can you add comment - what storage does have 'container' ? the same about 'path'?

label_studio/tasks/models.py
```diff
@@ -468,10 +466,8 @@ def resolve_uri(self, task_data, project):
 
                 # project storage
                 # TODO: to resolve nested lists and dicts we should improve get_storage_by_url(),
-                # TODO: problem with current approach: it can be used only the first storage that get_storage_by_url
-                # TODO: returns. However, maybe the second storage will resolve uris properly.
-                # TODO: resolve_uri() already supports them
-                storage = self.storage or get_storage_by_url(task_data[field], storage_objects)
+                # Now always using get_storage_by_url to ensure the storage with the correct bucket is used
+                storage = get_storage_by_url(task_data[field], storage_objects)
```
makseq: What if we use this line? Is it safer, or does it not make sense? As I understand, self.storage exists when a task is linked with a storage using StorageLink. So it makes sense to have the self.storage check as the last resort.. does not it? or I am wrong.. 
```
get_storage_by_url(task_data[field], storage_objects) or self.storage
```

makseq (approved):

PR #7459, opened 2025-05-01
fix: DIA-2092: skip failing test

matt-bernstein:
Skip failing test in LSO due to FF mocking being broken in new tests outside the `tests/` directory. These tests are inherited and run again in LSE, so test coverage is unchanged when both repos are considered.


jombooth (approved):
approving to unblock development, but I think it's important we figure this out and remove the skip soon - thanks for diligently working on this 👍

PR #7468, opened 2025-05-05
fix: LEAP-2060: Dynamic choices are not displayed when Interactive View All FF is turned ON

nass600:
At the moment, when building the `DynamicChoices` from the task data we are generating an error due to not using ids in the choice nodes which is needed when the `Interactive View All` flag is enabled.

This exception causes the Dynamic Choices to not render in the View either in the Review Page and Review All.

This fix generates ids for the nodes and creates an initial check to avoid adding the same node several times.

Screenshot 2025-05-05 at 12 39 33


web/libs/editor/src/mixins/DynamicChildrenMixin.js
```diff
@@ -14,9 +15,15 @@ const DynamicChildrenMixin = types
     const prepareDynamicChildrenData = (data, store, parent) => {
       if (data && data.length) {
         for (const obj of data) {
+          // Multiple traversal of the tree will cause the same object to be added multiple times
+          if (parent.children?.find((child) => child.value === obj.value)) continue;
```
bmartel: This part makes me wonder, if we detect that the first child value exists, can we skip the remainder entirely? Or is it possible to have an incomplete tree?

bmartel (approved):

web/libs/editor/src/mixins/DynamicChildrenMixin.js
nass600: The Tree is traversed multiple times (probably wrong but not sure entirely) and that's why the check was added. On the first run it will parse the whole Tree (nested as well) and it should work

Gondragos: Most probably we could skip all preparations when the feature flag is switched on and annotation is not defined.
Oh… or even the opposite, if this data is the same for all annotations, then preparations may be skipped for the case when annotation.id is defined.

hlomzik: @nass600 what Brandon meant is that if you already have one dynamic children in a tree, then you'll have all the rest children there as well, so every iteration of this loop will be skipped. So basically it can be changed to this:
```suggestion
          if (parent.children?.find((child) => child.value === obj.value)) break;
```

hlomzik: But after talking with Sergey we came up with even simpler check on the top of the method: `if (store.initialized) return;`. It will only be false within `initRoot()` call where we are propagating the values to tree nodes. All the other calls inside annotation's `afterCreate()` will use children, added before.
Also this check should only work for FF ON.

nass600: Got it, will double check it tomorrow

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14855561747)
> Successfully merged:  create mode 100644 .github/ISSUE_TEMPLATE/config.yml

web/libs/editor/src/mixins/DynamicChildrenMixin.js
nass600: Had to make a few more changes but it should be ready to review again @hlomzik @Gondragos @bmartel.
Note: I added the FF check for extra safety but it is not needed

web/libs/editor/src/mixins/DynamicChildrenMixin.js
```diff
@@ -51,9 +62,13 @@ const DynamicChildrenMixin = types
         // we may need to rewrite this, initRoot and the other related methods
         // (actually a lot of them) to work asynchronously as well
 
-        setTimeout(() => {
+        if (ff.isActive(FF_DEV_3391)) {
```
hlomzik, line 65: why do you need this change? I'm not sure about side effects of it being both sync and async.

nass600: The setTimeout is forcing this code to be executed on the next tick so, by the time it reaches the point to populate the array, the store is always initialized.

If we are planning to get rid of the flag in the short term we should be able to remove the setTimeout and the the ff check on store initialization, it works

nass600: Why did we add it in the first place by the way?

hlomzik (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14861133520)
> Successfully merged:  1 file changed, 35 insertions(+), 1 deletion(-)

Gondragos (approved):

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14870520523)
> Successfully merged:  create mode 100644 web/libs/core/src/lib/utils/query-client.ts

web/libs/editor/src/mixins/DynamicChildrenMixin.js
nass600: Selenium passed, resolving

web/libs/editor/src/mixins/DynamicChildrenMixin.js
nass600: Selenium passed, resolving

PR #6187, opened 2024-08-12
fix: OPTIC-966: converting stylus files to scss

carly-bartel:
converting stylus files to scss in LS web/libs/editor/src/tags/

PR #7481, opened 2025-05-06
fix: OPTIC-1916: Filter Tag is Not Working on Label Studio

### LINKED ISSUE #7347: Filter does not work for "Choices" tags
**Describe the bug**
Filter is not working for "Choices" tag

**To Reproduce**
1. Create a new project
2. Go to Settings  -> Labelling Interface
3. Copy paste the following XML
```XML
<View>
    <View style="display: flex; align-items: start; gap: 8px; flex-direction: column; white-space: pre-wrap;">
        <Text name="text" value="$text" granularity="word"/>
    </View>
    <Filter name="filter" toName="category" hotkey="shift+f" minlength="0" placeholder="Search"/>
    <Choices name="category" toName="text" choice="multiple" showInLine="false" required="true">
		<Choice value="oak"/>
		<Choice value="pine"/>
		<Choice value="maple"/>
		<Choice value="birch"/>
		<Choice value="elm"/>
		<Choice value="cedar"/>
    </Choices>
</View>
```
4. Input text in filter box
5. As you type, no filtering happens. List of choices remains as it is.

**Expected behavior**
Filtered set of choices, after any text is input in the filter box

**Screenshots**
should just show option "maple"
Image

 - OS: Ubuntu 24
 - Label Studio Version 1.17.0

luarmr:
The style .lsf-choice_hidden contain no styles this was introduce in a recent migration [here](https://github.com/HumanSignal/label-studio/pull/6187/files#diff-346bb02c58e7899bdebe8ea92303f9fd29f75ba3121f61fb48aa1543c24f70acL17)

The current problem is that the filter does not hide the options. But the right className is being attach

Also fixes: https://github.com/HumanSignal/label-studio/issues/7347

### DIFF
--- a/web/libs/editor/src/tags/control/Choice/Choice.scss
+++ b/web/libs/editor/src/tags/control/Choice/Choice.scss
@@ -18,8 +18,8 @@
     line-height: 40px;
   }
 
-  &_leaf&_hidden,
-  &_notLeaf&_hidden {
+  @at-root #{selector-append(&,_leaf,&,_hidden)},
+  #{selector-append(&,_notLeaf,&,_hidden)} {
     display: none;
   }
 
--- a/web/libs/editor/src/tags/control/Choice/Choice.scss
+++ b/web/libs/editor/src/tags/control/Choice/Choice.scss
@@ -18,8 +18,8 @@
     line-height: 40px;
   }
 
-  &_leaf &_hidden,
-  &_notLeaf &_hidden {
+  &_leaf&_hidden,
+  &_notLeaf&_hidden {
     display: none;
   }
 


web/libs/editor/src/tags/control/Choice/Choice.scss
```diff
@@ -18,8 +18,8 @@
     line-height: 40px;
   }
 
-  &_leaf &_hidden,
-  &_notLeaf &_hidden {
+  &_leaf&_hidden,
+  &_notLeaf&_hidden {
```
bmartel: What are the full output classes here?

luarmr: `&=.lsf-choice`

luarmr: well more than that. Notice I try to pair with the original stylus code.

bmartel (approved):

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14868734865)

web/libs/editor/src/tags/control/Choice/Choice.scss
```diff
@@ -18,8 +18,8 @@
     line-height: 40px;
   }
 
-  &_leaf &_hidden,
-  &_notLeaf &_hidden {
+  @at-root #{selector-append(&,_leaf,&,_hidden)},
+  #{selector-append(&,_notLeaf,&,_hidden)} {
```
luarmr, line 22: remove the space showed:
`"&" may only used at the beginning of a compound selector.`
The usual solution `#{&}` did not work, so to avoid using important or attaching extra specificity in front and generating the same selector as in the past and not introduce random bugs, I had to be creative.

bmartel: Tested it locally, works well! For cases like this I have often just used the full selector, but because this component is still using BEM, it also has weird nesting requirements that typically are not a problem to just use the proper class reference in full.

yyassi-heartex (approved):

luarmr:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14873008813)

PR #7484, opened 2025-05-07
fix: LEAP-2028: Fix brush cursor updating and visibility

Gondragos:
Fix generating cursor image for brush.
Unified cursor updating logic across tools and regions by introducing a shared `updateCursor` method. Refactored brush and erase tools to use the `BrushCursorMixin` for consistent custom cursor generation. Simplified and removed redundant cursor management code from multiple regions and tools, improving maintainability.

---

### Reason for change

The brush cursor was not updating correctly or consistently across tools and regions. The solution introduces a shared `updateCursor` method to unify and simplify cursor update logic. The `BrushCursorMixin` is now leveraged for consistent custom cursor management, improving maintainability and removing redundant code.

Image bigger that 128px can't be displayed as a cursor by most browsers.

#### Before

image
image

#### After

image
image

### Testing

This change has been tested locally and manually:
- Verified that the brush and erase tools display the appropriate updated cursors.
- Ensured no regression issues in cursor management across all regions and tools.

### Risks

No significant risks are identified. Changes simplify the cursor management logic without altering unrelated functionality.


nass600 (approved):
Nice refactor, pretty neat 👏 . My only comment is, if the cursor will be used at least in `Brush` and `Erase` maybe `CursorMixin` and having its own file could be more semantic but not sure if I might be missing some other context

Gondragos:
> fix: LEAP-2028: Fix brush cursor updating and visibility HumanSignal/label-studio-enterprise#7458

I thought about its own file for that, but this mixin is too specific (only Brush tools pack), so I prefer to keep it coupled for now.

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14906328129)
> Successfully merged:  create mode 100644 docs/source/guide/install_prompts.md

web/libs/editor/src/mixins/KonvaRegion.js
```diff
@@ -51,6 +53,28 @@ export const KonvaRegionMixin = types
...
+        const stage = self.parent?.stageRef;
+        if (!stage) return;
+        const style = stage.container().style;
+
+        if (isHovered) {
+          if (self.annotation.isLinkingMode) {
+            style.cursor = Constants.LINKING_MODE_CURSOR;
+          } else {
+            if (self.type !== "brushregion") style.cursor = Constants.POINTER_CURSOR;
+          }
```
hlomzik:
```suggestion
          } else if (self.type !== "brushregion") {
            style.cursor = Constants.POINTER_CURSOR;
          }
```

web/libs/editor/src/utils/canvas.js
```diff
@@ -345,13 +345,27 @@ function Region2RLE(region) {
   return rle;
 }
 
-function brushSizeCircle(size) {
+/**
+ * Creates a custom circular cursor visual representation based on the provided brush size.
+ * Max size: 128px according to browser limitations.
```
hlomzik, line 350: Does it make sense to add checks for this and limit the `size` in the code?

hlomzik (approved):
I don't really like the name of the method with this parameter, because then `updateCursor(true)` looks unclear. One of the ways to go could be to use named param: `updateCursor({ hovered: true })`.
But that's a very focused feature anyway, so I'm okay with current version.

web/libs/editor/src/utils/canvas.js
Gondragos: For now we do not have such a big brush size. So I ignored it temporarily.

PR #7488, opened 2025-05-07
fix: LEAP-2064: Choices on imported annotations are not displayed in View All on initial load

nass600:
There are two flags at play related with this bug:

- `fflag_fix_front_dev_3391_interactive_view_all`
- `fflag_fix_front_leap_443_select_annotation_once`

The issue comes from selecting "Compare all annotations" view and having the flag `fflag_fix_front_leap_443_select_annotation_once` enabled given that introduces performance improvements so we can keep the interactivity of the annotations while we do not `select` all of them just the current one.

Now, during the select process, control tags (`Choices`, `Taxonomy`, `Rating`, `DateTime`, `Number`) values are computed so they can be rendered with the right checked/unchecked state and having this flag enabled skips this process.

The fix introduces the computation of these selections based on the results when the user enters in the "Compare all annotation" mode without "selecting" them

**Additionally**, the PR will fix some cosmetic issues with `antd` checkboxes


yyassi-heartex (approved):

web/libs/editor/src/stores/Annotation/store.js
```diff
@@ -68,6 +68,24 @@ const AnnotationStoreModel = types
...
+        // annotation but not the rest of them.
+        // This fix aims to mimic the behaviour of selectAnnotation when it comes to updating the objects only without
+        // actually executing the full process of selecting the annotation.
+        if (isFF(FF_SIMPLE_INIT)) {
+          [...self.predictions, ...self.annotations].forEach((c) => {
+            // Skip the current annotation as it's already handled
+            if (c === self.selected) return;
+
+            // Set results for each annotation without selecting it
+            c.updateObjects();
```
hlomzik: I believe `c` is leftover from old `completion`, so let's at least fix it to more correct `a`.

```suggestion
          [...self.predictions, ...self.annotations].forEach((a) => {
            // Skip the current annotation as it's already handled
            if (a === self.selected) return;

            // Set results for each annotation without selecting it
            a.updateObjects();
```

web/libs/editor/src/tags/control/Choice/Choice.scss
```diff
@@ -74,7 +74,7 @@ body :global(.ant-checkbox-wrapper) {
 
 body :global(.ant-checkbox .ant-checkbox-inner) {
   background-color: var(--color-neutral-surface);
-  border-color: var(--color-neutral-border);
+  border-color: var(--color-neutral-border) !important;
```
hlomzik, line 77: How neccessary is this? I know that we will rewrite Choice for our own `@ui/checkbox` so not against this dirty fix, but just curious.

hlomzik (approved):

web/libs/editor/src/tags/control/Choice/Choice.scss
nass600: the `antd reset` css file is applying `!important` already and cannot bypass the specificity without altering it so it is neccessary I am afraid.

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14907747039)
> Successfully merged:  create mode 100644 docs/source/guide/install_prompts.md

Gondragos (approved):

PR #7502, opened 2025-05-09
fix: LEAP-2073: Handle root property in history item

Gondragos:
Ensure `root` is correctly assigned when creating or modifying annotation history. Updated `createItem` to utilize `options.root` or fallback to `self.root` and added conditional logic to define `root` in `addHistory` when feature flag FF_DEV_3391 is enabled.

---

### Reason for change

MIG didn't work with interactive `View All`, when we go through the history of annotation.

### Rollout strategy

- The changes are gated under the feature flag `FF_DEV_3391`.\

### Risks

- Minimal risk as changes are behind a feature flag. If the logic fails, history-related operations might be affected.

### Reviewer notes

In interactive `View All` each annotation has its own root, to be able to render it at the same time (each tag should have the same state). But that was not taken into account for setting history. History won't be displayed at the same time with the current annotation state as well as each history item won't be displayed at the same time with others. That means that a history item can relay on annotation's root version.

### General notes

- This change is limited to history management logic and doesn't alter external interfaces.

### DIFF
--- a/web/libs/editor/src/stores/Annotation/store.js
+++ b/web/libs/editor/src/stores/Annotation/store.js
@@ -329,7 +329,7 @@ const AnnotationStoreModel = types
         id: guidGenerator(5),
         // pk and id may be missing, so undefined | string
         pk: pk && String(pk),
-        root: self.root,
+        root: options.root ?? self.root,
       };
 
       if (user && !("createdBy" in node)) node.createdBy = user.displayName;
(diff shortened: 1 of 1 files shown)


hlomzik (approved):

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/14940392002)
> Successfully merged:  1 file changed, 28 insertions(+), 9 deletions(-)

PR #7506, opened 2025-05-09
fix: OPTIC-2148: Icons in the circles under pre-annotation regions are missing

ricardoantoniocm:
### Reason for change
The checkmark and cross icons on the preannotation circles were not displaying correctly. Additionally, the colors of the Accept and Reject buttons were not aligned with the new branding guidelines. This PR addresses both of these issues to improve the visual consistency and user experience.

This PR also introduces a smoother transition animation to the hover state of the buttons as well as displaying the pointer cursor when hovering them.

### Screenshots
- Before: 
image
image
image

- After:
image
image
image

### Rollout strategy
No specific rollout strategy is required for this visual fix. The changes will be deployed with the next release.

### Testing
1. Log in to the LSO application in the specified environment.
2. Navigate to a project with a connected model and interactive preannotations enabled (as per the preconditions).
3. Open a task and ensure a preannotation is displayed for a selected label.
4. Hover over the circles under the preannotation region.
5. Verify that the left circle now displays a white cross on a red background (Reject).
6. Verify that the right circle now displays a white checkmark on a green background (Accept).
7. Confirm that the Accept and Reject buttons in the preannotation interface reflect the new branding colors.

### Risks
Low. This PR primarily involves visual corrections and should not introduce any functional regressions.

### Reviewer notes
- Please pay close attention to the correct rendering of the icons and the updated button colors.
- Ensure that the changes are consistent across different browsers if possible.

### General notes
- The color adjustments for the Accept and Reject buttons were made to align with the latest branding specifications.
- The fix ensures that users can clearly distinguish between the accept and reject actions for preannotations through both color and iconography.

### DIFF
--- a/web/libs/editor/src/components/ImageView/SuggestionControls.jsx
+++ b/web/libs/editor/src/components/ImageView/SuggestionControls.jsx
@@ -116,12 +116,12 @@ const ControlButton = ({ x = 0, fill, iconColor, onClick, icon }) => {
     iconImage.onload = () => {
       setImg(iconImage);
     };
-    iconImage.width = 12;
-    iconImage.height = 12;
+    iconImage.width = 20;
+    iconImage.height = 20;
 
-    const iconElement = React.createElement(icon, { color: iconColor, width: 12, height: 12 });
+    const iconElement = React.createElement(icon, { color: iconColor, width: 20, height: 20 });
     const svgString = ReactDOMServer.renderToStaticMarkup(iconElement);
-    const base64 = btoa(unescape(encodeURIComponent(svgString)));
+    const base64 = btoa(decodeURIComponent(encodeURIComponent(svgString)));
     iconImage.src = `data:image/svg+xml;base64,${base64}`;
   }, [icon, iconColor]);
 
--- a/web/libs/editor/src/components/ImageView/SuggestionControls.jsx
+++ b/web/libs/editor/src/components/ImageView/SuggestionControls.jsx
@@ -1,10 +1,12 @@
 import { useCallback, useEffect, useState } from "react";
 import { Circle, Group, Image, Layer, Rect } from "react-konva";
-import { IconCheck, IconCross } from "@humansignal/icons";
 import Konva from "konva";
 import chroma from "chroma-js";
 import { observer } from "mobx-react";
 import { isDefined } from "../../utils/utilities";
+import ReactDOMServer from "react-dom/server";
+import React from "react";
+import { IconCheck, IconCross } from "../../../../ui/src/assets/icons";
 
 const getItemPosition = (item) => {
   const { shapeRef: shape, bboxCoordsCanvas: bbox } = item;
@@ -100,10 +102,13 @@ export const SuggestionControls = observer(({ item, useLayer }) => {
 
 const ControlButton = ({ x = 0, fill, iconColor, onClick, icon }) => {
   const [img, setImg] = useState(new window.Image());
-  const imageSize = 16;
+  const imageSize = 20;
   const imageOffset = 32 / 2 - imageSize / 2;
   const color = chroma(iconColor ?? "#FFFFFF");
   const [hovered, setHovered] = useState(false);
+  const [animatedOpacity, setAnimatedOpacity] = useState(0.2);
+  const [animatedFill, setAnimatedFill] = useState("#fff");
+  const animationRef = React.useRef();
 
   useEffect(() => {
     const iconImage = new window.Image();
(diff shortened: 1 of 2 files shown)


web/libs/editor/src/components/ImageView/SuggestionControls.jsx
```diff
@@ -73,15 +73,15 @@ export const SuggestionControls = observer(({ item, useLayer }) => {
...
+          iconColor="#FFFFFF"
           icon={IconCross}
         />
         <ControlButton
           x={32}
           onClick={() => item.annotation.acceptSuggestion(item.id)}
-          fill="#98C84E"
-          iconColor="#fff"
+          fill="#287A72"
+          iconColor="#FFFFFF"
```
yyassi-heartex, line 86: these colors will be the same for light and darkmode, is that intended? if we need separate colors we can look at `getCurrentTheme()` from `web/libs/ui/src/lib/ThemeToggle/ThemeToggle.tsx`

hlomzik: and can we use css vars here?

yyassi-heartex: dug into it a little, css var probably wont work because fill gets passed into a Konva `Circle` component

ricardoantoniocm: I tried CSS vars, it didn't work.

ricardoantoniocm: @yyassi-heartex the colors look ok on both Light and Dark Mode even though we aren't using the correct vars.

That said, this will need more work because the icons themselves aren't visible and it seems they had not been visible for some time (issue seems unrelated to recent work). The above fix by Brandon unfortunately didn't address the issue.

@hlomzik is this something you could help us look into?

hlomzik: could you show some screenshots? also if you add them to description that would be helpful

ricardoantoniocm: image
image
image

ricardoantoniocm: I've updated the description and added screenshots of the current state as well as the improvements so far (adjusting the background colors of the buttons to better match the brand). @hlomzik

ricardoantoniocm: @hlomzik  The fix required using the relative path to the SVGs, instead of importing them from `@humansignal/icons`.

I made some adjustments to the size of the icon and made the hover transition subtler. This was done with AI so I would appreciate a second set of eyes to make sure it didn't do anything funky that we wouldn't approve.

CC: @bmartel @yyassi-heartex

ricardoantoniocm: image
image

ricardoantoniocm: https://www.loom.com/share/e0f1f36a18e04fbaaf30d22826e0f0b1?sid=a898db3e-a9ce-4b2c-98f6-810480a41bd8

hlomzik: oh holy moly... what is this...
that's not funky, that's scary a little. we should really improve our cursor rules to stick to tools we already have.

hlomzik (changes requested):

web/libs/editor/src/components/ImageView/SuggestionControls.jsx
ricardoantoniocm: I'd appreciate constructive feedback. What is there to fix?

hlomzik: honestly — everything. we just need to render two icons. what is the current problem with them? it's on a wrong path or what?

bmartel: @hlomzik The problem here is simple, though the solution may not be.

The icons never were able to be rendered previously, because it was always requiring an image to be used for canvas, but our icons have always been SVG components through SVGR. So that means this never worked, because the resulting icons for as long as I have been here have been handled through SVGR, and would produce an SVG React component, not a base64 url or url to an svg.

Now, the solution here I would say we can't entertain as is, because it involves using React Server Rendering utilities which I am still surprised it even works on the client like this to be honest. As to what else could we do here, maybe have a special import of the svg which produces a url to the asset instead of a component with the svg inside, that might be a simpler approach.

ricardoantoniocm: Thanks for the guidance @bmartel . I'll try that approach.

hlomzik: these icons were definitely working at some point...
I see 2 options:
- to modify imports as you mentioned to get the path to svg and add it via Konva.Image. doesn't SVGR allow to get the path or stringified svg?
- to recreate this panel in a real svg, this might be more complicated

ricardoantoniocm: I'll leave this as unresolved for the time being. It's a low priority fix in any case.

In short, I was not successful in applying the recommended approach (removing React Server Rendering and attempting to pull icons as URLs) after many prompt iterations. Happy to jump back to it with some help.

hlomzik: yeah, that's not a styling issue anymore, it requires some subtle but fundamential changes

robot-ci-heartex:

robot-ci-heartex:

PR #7527, opened 2025-05-14
fix: BROS-5: Last fixes for Interactive View All

Gondragos:
---

### Reason for change

- Fixed tools initialization to prevent multiple initializations which led to the loss of default tool functionality.
- Added a unique identifier to the Collapse model to support an interactive "View All" feature without breaking the application.
- Remove last (?) occurencies of reaching to `annotationStore.selected` from tags
- Quick fix TimeSeries async init to not crash the app when original instance is missing
- Fixed lost cursor managing
- Fixed Image's labels tools (didn't work like a tool)


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15111983801)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/sparkles/utils.ts

hlomzik (approved):

PR #7534, opened 2025-05-14
feat: ROOT-1: Add offset to storage link model to support storage link to task within file of tasks on cloud storage

matt-bernstein:
Model changes to prepare to flip `fflag_feat_dia_2092_multitasks_per_storage_link`, supporting JSON[L] and Parquet files

This is the minimal change necessary atm; future possible changes are:
- add a filetype enum when that can no longer be inferred from row_idx+row_group
- previously this model was both file-level and task-level. Now that one file can contain many tasks, file-level fields can be extracted into a separate table to remove data duplication.
- add filesize, checksum fields to support appending to files in place


label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,27 @@ class Meta:
 class ImportStorageLink(models.Model):
 
     task = models.OneToOneField('tasks.Task', on_delete=models.CASCADE, related_name='%(app_label)s_%(class)s')
+
+    # ---
+    # NOTE: ImportStorageLink is task-level, not key-level. These fields could be split out into a new, key-level table to optimize DB access patterns.
+
     key = models.TextField(_('key'), null=False, help_text='External link key')
+
+    # NOTE: unused
```
jombooth: This isn't entirely true, I found a reference to the field here: https://github.com/HumanSignal/label-studio/blob/ade0e61ef9c41cc0941cc04707139c57bf2e1294/label_studio/io_storages/base_models.py#L746

In general I'm not a fan of comments unless they provide essential information; I'm in favor of removing
```suggestion
```

label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,27 @@ class Meta:
 class ImportStorageLink(models.Model):
 
     task = models.OneToOneField('tasks.Task', on_delete=models.CASCADE, related_name='%(app_label)s_%(class)s')
+
+    # ---
+    # NOTE: ImportStorageLink is task-level, not key-level. These fields could be split out into a new, key-level table to optimize DB access patterns.
```
jombooth: We should remove this, it's too speculative. Happy to have this in a doc, but in code, I don't feel it's helpful.

```suggestion
```

label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,27 @@ class Meta:
...
         _('object exists'), help_text='Whether object under external link still exists', default=True
     )
+    # ---
+
     created_at = models.DateTimeField(_('created at'), auto_now_add=True, help_text='Creation time')
 
+    # NOTE: we could support other file formats in future with an equivalent of row_group - ORC, Feather, Avro.
+    # In that case, we would need a file_type enum or similar to distinguish them in filters.
+    row_group = models.PositiveIntegerField(null=True, blank=True, help_text='Parquet row group')
+    row_idx = models.PositiveIntegerField(
```
jombooth: I'd prefer `row_index` - no need to use abbreviations in field names.

label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,27 @@ class Meta:
...
+    # NOTE: unused
     object_exists = models.BooleanField(
         _('object exists'), help_text='Whether object under external link still exists', default=True
     )
+    # ---
+
     created_at = models.DateTimeField(_('created at'), auto_now_add=True, help_text='Creation time')
 
+    # NOTE: we could support other file formats in future with an equivalent of row_group - ORC, Feather, Avro.
+    # In that case, we would need a file_type enum or similar to distinguish them in filters.
```
jombooth:
```suggestion
```

Let's remove this, again, too speculative

label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,27 @@ class Meta:
...
+    # ---
+    # NOTE: ImportStorageLink is task-level, not key-level. These fields could be split out into a new, key-level table to optimize DB access patterns.
+
     key = models.TextField(_('key'), null=False, help_text='External link key')
+
+    # NOTE: unused
     object_exists = models.BooleanField(
         _('object exists'), help_text='Whether object under external link still exists', default=True
     )
+    # ---
```
jombooth:
```suggestion
```

jombooth:
We also need not to use PositiveIntegerField. I ran `python manage.py sqlmigrate` to see what sql this migration is generating:

```
> python manage.py sqlmigrate io_storages 0019

...blah blah...

BEGIN;
--
-- Add field row_group to azureblobimportstoragelink
--
ALTER TABLE "io_storages_azureblobimportstoragelink" ADD COLUMN "row_group" integer NULL CHECK ("row_group" >= 0);
--
-- Add field row_idx to azureblobimportstoragelink
--
ALTER TABLE "io_storages_azureblobimportstoragelink" ADD COLUMN "row_idx" integer NULL CHECK ("row_idx" >= 0);
--
-- Add field row_group to gcsimportstoragelink
--
ALTER TABLE "io_storages_gcsimportstoragelink" ADD COLUMN "row_group" integer NULL CHECK ("row_group" >= 0);
--
-- Add field row_idx to gcsimportstoragelink
--
ALTER TABLE "io_storages_gcsimportstoragelink" ADD COLUMN "row_idx" integer NULL CHECK ("row_idx" >= 0);
--
-- Add field row_group to localfilesimportstoragelink
--
ALTER TABLE "io_storages_localfilesimportstoragelink" ADD COLUMN "row_group" integer NULL CHECK ("row_group" >= 0);
--
-- Add field row_idx to localfilesimportstoragelink
--
ALTER TABLE "io_storages_localfilesimportstoragelink" ADD COLUMN "row_idx" integer NULL CHECK ("row_idx" >= 0);
--
-- Add field row_group to redisimportstoragelink
--
ALTER TABLE "io_storages_redisimportstoragelink" ADD COLUMN "row_group" integer NULL CHECK ("row_group" >= 0);
--
-- Add field row_idx to redisimportstoragelink
--
ALTER TABLE "io_storages_redisimportstoragelink" ADD COLUMN "row_idx" integer NULL CHECK ("row_idx" >= 0);
--
-- Add field row_group to s3importstoragelink
--
ALTER TABLE "io_storages_s3importstoragelink" ADD COLUMN "row_group" integer NULL CHECK ("row_group" >= 0);
--
-- Add field row_idx to s3importstoragelink
--
ALTER TABLE "io_storages_s3importstoragelink" ADD COLUMN "row_idx" integer NULL CHECK ("row_idx" >= 0);
COMMIT;
```

every "CHECK" is going to hold a lock on the table until the constraint can be built; this operation is going to be slow on large tables and could crash SaaS. In practice, I think there's very low risk of us writing a negative integer into either row_index or row_group, but if we want, we could implement a check on that in app-level code.

label_studio/io_storages/base_models.py
matt-bernstein: Yes, it always gets/creates using the default value of `True`, which is equivalent to the field not existing since it's never accessed. I wrote the comment because I had the question "what's the difference between `exists` and `object_exists`?" - as I think anyone reading this code would, and found that it had a clear answer.

Will remove it if you think it's still out of scope, but this should be documented somewhere, imho, if not here.

jombooth: Something like this then - "unused" could indicate to the developer (incorrectly) that the field doesn't appear in sql queries and is safe to delete, which isn't true.

```suggestion
    # This field is set to True on creation and never updated; it should not be relied upon.
```

label_studio/io_storages/base_models.py
```diff
@@ -688,12 +688,19 @@ class Meta:
 class ImportStorageLink(models.Model):
 
     task = models.OneToOneField('tasks.Task', on_delete=models.CASCADE, related_name='%(app_label)s_%(class)s')
+
```
jombooth:
```suggestion
```

jombooth (approved):

label_studio/io_storages/base_models.py
matt-bernstein: That's a better wording, thanks. I was relying too heavily on the implicit understanding that we never delete fields anyways :)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15027678115)

PR #7535, opened 2025-05-14
ci: OPTIC-2173: Run yarndedup

luarmr:
Testing yarn-deduplicate

node_modules  pass from 1.7G to 1.3G 
Probably we should bump our yarn version

### DIFF
--- a/web/yarn.lock
+++ b/web/yarn.lock
@@ -2,12 +2,7 @@
 # yarn lockfile v1
 
 
-"@adobe/css-tools@^4.0.1":
-  version "4.3.2"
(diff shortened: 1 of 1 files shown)


bmartel:
Did you run just `yarn dedupe` or with a specific strategy?

luarmr:
Currently, I'm just yarn dedupe. I was aiming for --higher, but it's failing in Babel right now, so I'm not sure that will fix the issue. It looks like fine-tuning will be required, and this would need to happen every time we add new libraries. Seems annoying, doesn't it?

luarmr:
All of this is almost free in newer versions of yarn

PR #7555, opened 2025-05-16
feat: ROOT-11: Support reading JSONL from source cloud storages

matt-bernstein:
[x] Factor out JSON parsing and validation logic from import storages into a function that can be hot-swapped in LSO/LSE
[x] add it to settings as an app-specific variable
[x] replace stdlib parsing with pyarrow parsing to handle JSONL as well as JSON
[x] test coverage for JSONL
[x] feature flag
[x] unskip multitask import tests in LSO, to prepare for testing different behavior in LSE and LSO

Implementation notes:
- had to move computation of row_index and row_group into get_data so that load_tasks_json, which will be swapped in LSO/LSE, will have the correct type in all cases - makes for a pretty large diff
- simplified error messages significantly since the relevant logic is shared/deduped
- had to stick with stdlib json parsing instead of removing it entirely:
  - in the case of single dict, pyarrow can't distinguish between this and a one-line JSONL file, so need stdlib json to set row_index=None
  - in the case of top-level list, pyarrow can't parse this at all since it assumes each line is an entry in a table. There is an [open issue](https://github.com/apache/arrow/issues/46486) for this.
  - consider peeking at file extensions to avoid double-parsing (for clarity, not speed) - see TODO comment


matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15074933586)

label_studio/io_storages/azure_blob/models.py
```diff
@@ -209,28 +209,16 @@ def iterkeys(self):
                 continue
             yield file.name
 
-    def get_data(self, key) -> Union[dict, list[dict]]:
+    def get_data(self, key) -> tuple[list[dict], list[int | None], list[int | None]]:
```
jombooth: Once we're at a 3-tuple return type, I think it'd be better to create a simple object with a constructor and return that, or use a solution like NamedTuple or TypedDict. Python errors from unpacking tuples into too few variables in assignment statements are quite common

jombooth: also, for the `list[int | None]` types, I think it'd be better to say `list[int] | Literal[[None]]`, since there is no other list containing `None`s that we could want returned afaik.

matt-bernstein: Not a bad idea. I'm thinking `(task_data: list[dict] (or pa.Table), storagelink_params: list[StorageLinkParam])`

matt-bernstein: though tbf, this function will only ever have one caller

jombooth: I also think that perhaps we should find an alternative to returning lists containing a single `[None]` - I get why we have them now, but it seems to me we could find a better way of representing this concept in Python

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,61 @@ def parse_range(range_header):
...
+        list[dict]: parsed tasks.
+        list[int|None]: row_index for each task.
+        list[int|None]: row_group for each task.
+    """
+
+    def _error_wrapper(exc: Optional[Exception] = None):
+        raise ValueError(
+            (
+                f"Can't import JSON-formatted tasks from {key}. If you're trying to import binary objects, "
+                f'perhaps you\'ve forgot to enable "Treat every bucket object as a source file" option?'
```
jombooth:
```suggestion
                f'perhaps you forgot to enable "Treat every bucket object as a source file" option?'
```

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,61 @@ def parse_range(range_header):
...
+    Returns:
+        list[dict]: parsed tasks.
+        list[int|None]: row_index for each task.
+        list[int|None]: row_group for each task.
+    """
+
+    def _error_wrapper(exc: Optional[Exception] = None):
+        raise ValueError(
+            (
+                f"Can't import JSON-formatted tasks from {key}. If you're trying to import binary objects, "
```
jombooth:
```suggestion
                f"Can’t import JSON-formatted tasks from {key}. If you’re trying to import binary objects, "
```

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,61 @@ def parse_range(range_header):
...
+                f'perhaps you\'ve forgot to enable "Treat every bucket object as a source file" option?'
+            )
+        ) from exc
+
+    # TODO: rely on file extensions here instead of pure exception-chaining? Would be more readable, potentially less reliable, would work for all storage types except redis.
```
jombooth: I think exception chaining is actually a better answer than file extensions, because it's not uncommon to use .txt or similar to represent JSON. Would favor removing this comment

label_studio/io_storages/gcs/models.py
```diff
@@ -180,28 +180,17 @@ def iterkeys(self):
...
-            return {settings.DATA_UNDEFINED_NAME: GCS.get_uri(self.bucket, key)}
-        data = GCS.read_file(
+            task = {settings.DATA_UNDEFINED_NAME: GCS.get_uri(self.bucket, key)}
+            return [task], [None], [None]
+        blob_str = GCS.read_file(
             client=self.get_client(),
             bucket_name=self.bucket,
             key=key,
-            convert_to=GCS.ConvertBlobTo.JSON,
+            convert_to=GCS.ConvertBlobTo.NOTHING,
```
jombooth: do we have to pass `convert_to` at all in this case?

matt-bernstein: Nope! Just tried to be very explicit. But it's not needed, will remove it

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,61 @@ def parse_range(range_header):
...
+    try:
+        value = json.loads(blob_str)
+    except json.decoder.JSONDecodeError as e:
+        if flag_set('fflag_feat_root_11_support_jsonl_cloud_storage'):
+            try:
+                table = pyarrow.json.read_json(
+                    pa.py_buffer(blob_str), parse_options=pa.json.ParseOptions(newlines_in_values=True)
+                )
```
jombooth, line 169: hmm, seems to me we may want to handle this differently - as written, with the FF enabled we have to fail to parse JSON to even attempt parsing JSONL. Could we condition whether we try to `json.loads()` vs `pyarrow.json.read_json()` on whether the FF is on or not? It might be okay as is, if we're confident that `json.loads()` will fail quickly on a large JSONL file.

matt-bernstein: `json.loads()` will fail on the first character of the second line of a JSONL file:

```
JSONDecodeError                           Traceback (most recent call last)
Cell In[1], line 1
----> 1 json.loads(b'{"a":1}\n{"a":2}')

File ~/anaconda3/lib/python3.12/json/__init__.py:346, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    341     s = s.decode(detect_encoding(s), 'surrogatepass')
    343 if (cls is None and object_hook is None and
    344         parse_int is None and parse_float is None and
    345         parse_constant is None and object_pairs_hook is None and not kw):
--> 346     return _default_decoder.decode(s)
    347 if cls is None:
    348     cls = JSONDecoder

File ~/anaconda3/lib/python3.12/json/decoder.py:341, in JSONDecoder.decode(self, s, _w)
    339 end = _w(s, end).end()
    340 if end != len(s):
--> 341     raise JSONDecodeError("Extra data", s, end)
    342 return obj

JSONDecodeError: Extra data: line 2 column 1 (char 8)
```

so it's linear in the size of the first task, not the number of tasks. Will verify this, though, I see the concern

matt-bernstein: reasonably confident in ^ after reading through stdlib json source

label_studio/io_storages/azure_blob/models.py
matt-bernstein: done

jombooth:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15147464154)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15148947939)

label_studio/core/settings/base.py
```diff
@@ -598,6 +598,7 @@
 MEMBER_PERM = 'core.api_permissions.MemberHasOwnerPermission'
 RECALCULATE_ALL_STATS = None
 GET_STORAGE_LIST = 'io_storages.functions.get_storage_list'
+STORAGE_LOAD_TASKS_JSON = 'io_storages.utils._load_tasks_json'
```
jombooth: Usually the meaning of a function or method that starts with a `_` is "this should not be used outside of this class/module/whatever" - seems to me that we are treating this function as more of a public one, since it's being referred to in this other file. Would it be a pain to rename?

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,73 @@ def parse_range(range_header):
         end = ''
 
     return start, end
+
+
+@dataclass
+class StorageObjectParams:
```
jombooth: I think `StorageObject` would be a better name for this, since it actually contains the task now

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,73 @@ def parse_range(range_header):
...
+    task_data: dict
+    key: str
+    row_index: int | None = None
+    row_group: int | None = None
+
+    @classmethod
+    def bulk_create(
+        cls, task_datas: list[dict], key, row_idxs: list[int] | None = None, row_groups: list[int] | None = None
```
jombooth: I would prefer `row_indexes` - generally in favor of avoiding abbreviations in variable names unless they're truly ubiquitous.

matt-bernstein:
Note for posterity: `pa.Table.from_pylist` has a footgun in that it only constructs the schema from the first row; use `from_json` with BytesIO next time since it constructs the schema from all rows

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,73 @@ def parse_range(range_header):
...
+def _load_tasks_json(blob_str: bytes, key: str) -> list[StorageObjectParams]:
+    """
+    Parse blob_str containing task JSON(s) and return the validated result or raise an error.
+
+    Args:
+        blob_str (bytes): The blob string to parse.
+        key (str): The key of the blob. Used for error messages.
+
+    Returns:
+        list[dict]: parsed tasks.
```
jombooth:
```suggestion
```

pyproject.toml
```diff
@@ -49,6 +49,7 @@ dependencies = [
     "ordered-set (==4.0.2)",
     "pandas (>=2.2.3)",
     "psycopg2-binary (==2.9.10)",
+    "pyarrow (>=18.0.0,<19.0.0)",
```
jombooth:
```suggestion
    "pyarrow (>=18.1.0,<19.0.0)",
```

Would recommend we don't allow people to go older than the version in poetry.lock, note this will require a relock

label_studio/io_storages/utils.py
```diff
@@ -109,3 +116,73 @@ def parse_range(range_header):
...
+
+    if isinstance(value, dict):
+        return [StorageObjectParams(key=key, task_data=value)]
+    if isinstance(value, list):
+        return StorageObjectParams.bulk_create(value, key, range(len(value)))
+
+    _error_wrapper()
+
+
+def load_tasks_json(blob_str: str, key: str) -> tuple[list[dict], list[StorageObjectParams]]:
```
jombooth:
```suggestion
def load_tasks_json(blob_str: str, key: str) -> list[StorageObjectParams]:
```

label_studio/io_storages/base_models.py
```diff
@@ -515,6 +523,7 @@ def sync(self):
                 self.info_set_queued()
                 import_sync_background(self.__class__, self.id)
             except Exception:
+                logger.debug(f'Storage {self} failed', exc_info=True)
```
jombooth, line 526:
```suggestion
                # needed to facilitate debugging storage-related testcases, since otherwise no exception is logged
                logger.debug(f'Storage {self} failed', exc_info=True)
```

jombooth: This is a rare case where I'm in favor of a comment - the reason it's here is fairly subtle and not at all obvious from the context

label_studio/io_storages/base_models.py
```diff
@@ -431,7 +438,8 @@ def _scan_and_create_links(self, link_class):
 
             logger.debug(f'{self}: found new key {key}')
             try:
-                tasks_data = self.get_data(key)
+                # list of (task data + ImportStorageLink details)
+                links_params = self.get_data(key)
```
jombooth: with the change to `StorageObject`, this can be `storage_objects`, which I think will be a lot more readable than `links_params` vs `link_params` (although, valiant effort at making the plurals situation readable enough!)

label_studio/io_storages/base_models.py
```diff
@@ -431,7 +438,8 @@ def _scan_and_create_links(self, link_class):
 
             logger.debug(f'{self}: found new key {key}')
             try:
-                tasks_data = self.get_data(key)
+                # list of (task data + ImportStorageLink details)
```
jombooth:
```suggestion
```

label_studio/io_storages/base_models.py
```diff
@@ -341,17 +342,20 @@ def _scan_and_create_links_v2(self):
         raise NotImplementedError
 
     @classmethod
-    def add_task(cls, data, project, maximum_annotations, max_inner_id, storage, key, row_index, link_class):
+    def add_task(cls, project, maximum_annotations, max_inner_id, storage, link_params, link_class):
+        link_kwargs = asdict(link_params)
+        data = link_kwargs.pop('task_data')
```
jombooth: is there a risk of a crash if the key is missing from the dict? genuinely asking, not extremely familiar with dataclasses

matt-bernstein: nope, at this point it's a regular dict and `.pop` has an implicit default `None`

jombooth: It's counterintuitive next to `dict#get`, but:

```
>>> {}.pop('foo')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'foo'
```

you have to pass `pop()` the `None` default if that's the expected behavior.

matt-bernstein: oops, you're right!

matt-bernstein: I don't think the dataclass could have had it missing to begin with, but good if someone accidentally passes another kind of object in there

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15188365408)

jombooth (approved):

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15192083566)
> Successfully merged:  1 file changed, 47 insertions(+), 2 deletions(-)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15192136431)

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15193781619)
> Successfully merged:  5 files changed, 44 insertions(+), 38 deletions(-)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15210874380)

PR #7565, opened 2025-05-19
chore: UTC-3: [FE] Refactor Popover components in LSE

nass600:
Refactored Agreement cell to support new Spark Popovers. Major changes involve triggering the Popover from LSO and getting the content from LSE given the actual component tree and how the triggering must work according to Shadcn specs.

Additionally:
- Removed BEM and adapted score colors to the new Palette
- After discussing with Design, the shadows have been hardcoded until we get the final tokens

Screenshot 2025-05-19 at 11 41 08

Screenshot 2025-05-19 at 11 41 25


web/libs/ui/src/lib/popover/popover.tsx
hlomzik, line 1: Let's make default `align="start"`, we don't use `"center"` anywhere as far as I know

web/libs/datamanager/src/components/CellViews/Agreement/Agreement.jsx
```diff
@@ -24,20 +28,31 @@ const formatNumber = (num) => {
...
   );
 
   return (
-    <div className={agreementCN.toString()} onClick={handleClick}>
-      <span className={clsx(scoreElem.toString(), scoreElem.mod({ [agreement(value)]: true }).toString())}>
-        {isDefined(value) ? `${formatNumber(value)}%` : ""}
-      </span>
+    <div className="flex items-center" onClick={handleClick}>
+      {isAgreementPopoverEnabled ? (
+        <Popover trigger={score} align="start">
```
hlomzik: so what's the idea here? is it triggered by clicking on `score` but the `content` is empty, so it will be in a confusing state? can we move content loading closer to Popover (does it have any events?) and add a loading state?

hlomzik: I see that all this logic is inside `agreementCellClick` handler in LSE... Given all this interconnection that looks like a good option to trigger it, I see...

hlomzik:
Can we make Popover more distinguishable in dark mode? dark shadow of the dark popover over dark background is invisible

web/libs/ui/src/lib/popover/popover.tsx
nass600: Good point, changing

nass600:
> Can we make Popover more distinguishable in dark mode? dark shadow of the dark popover over dark background is invisible

This has been discussed with @ricardoantoniocm already and, once we have ready the shadow tokens defined by the elevation, it could be reassessed if needed. For now, this shadow is coming as a requirement

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15130255175)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/view_all/smoke.cy.ts

hlomzik (approved):

PR #7575, opened 2025-05-20
fix: BROS-31: Inconsistent load for Pdf pages

niklub:
Use `<embed>` instead of `<iframe>` as 
- it is more suitable for embedding PDFs with browser native support
- it is consistent with [the previous method of loading PDF documents](https://github.com/HumanSignal/label-studio/blob/8e79aa535547675624e0fc237e88b649e992d851/label_studio/annotation_templates/structured-data-parsing/pdf-classification/config.yml#L15)

### DIFF
--- a/web/libs/editor/src/tags/object/Pdf.jsx
+++ b/web/libs/editor/src/tags/object/Pdf.jsx
@@ -40,6 +40,7 @@ const HtxPdf = inject("store")(
       <embed
         src={item._url}
         style={{ width: "100%", height: "600px", border: "none" }}
+        type="application/pdf"
       />
     );
   }),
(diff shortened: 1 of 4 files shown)


web/libs/editor/src/tags/object/Pdf.jsx
```diff
@@ -37,7 +37,7 @@ const HtxPdf = inject("store")(
   observer(({ item }) => {
     if (!item._url) return null;
     return (
-      <iframe
+      <embed
         src={item._url}
         title="PDF Document"
```
hlomzik: that feels like something we don't need. if it's showing up as a popup, that's annoying

hlomzik: also `allowFullScreen` would not work for `embed`, so remove it as well

hlomzik (approved):

yyassi-heartex (approved):

PR #7583, opened 2025-05-21
fix: Add activeLabelKey to generate unique key for selected labels

### LINKED ISSUE #7590: Bug Report: Newly Added Labels May Not Reflect on Waveform When Multiple Labels is Enabled
#### 🧾 Summary
When multiple labels are allowed , after selecting one label and annotating a region on the waveform, newly added labels are **not always reflected** on the waveform regions — even though they are selected.

---

#### 🕒 Steps to Reproduce

1. Configure an audio task using:
```
  <View>
    <Labels name="label" toName="audio"  choice="multiple">
      <Label value="Event A" background="red"/>
      <Label value="Event B" background="green"/>
    </Labels>
    <Audio name="audio" value="$audio"/>
  </View>
```

2. Select "Label A" and annotate a region on the waveform.
3. Add another label (e.g., "Label B") to the same region.
4. Observe that the waveform region may still show only "Label A".

---

#### 📷 Expected Behavior
All selected labels should be reflected in the waveform region's display.

---

#### 🖼️ Actual Behavior
Only the first selected label appears visually on the waveform region, even if more labels are selected. The UI doesn't update unless the primary label (first in list) changes.

Image

---

#### 📌 Additional Context
- A fix has been proposed in PR #7583  to address this issue by introducing a new computed property to track full label selection changes.

---

xxxinsxxx:
### 🧩 Feature Description

In the audio annotation feature, when a user selects a label and then adds a new one, the waveform region will not be updated if `selectedValues()[0]` remains unchanged. This causes newly added labels to not be reflected in the selected audio regions.

To address this issue, this PR introduces a new computed property called `activeLabelKey`, which generates a unique key based on the full list of selected labels. This ensures that any change in the label selection will trigger the update logic.

---

### ✅ What Changed

1. **New Computed Property: `activeLabelKey`**
   ```js
   get activeLabelKey() {
     const labels = self.activeState?.selectedValues();
     return labels ? labels.join(',') : '';
   }
   ```

2. **Updated Observer Logic**  
   Modified the observer from watching `activeLabel` to watching `activeLabelKey`, ensuring that changes to the full label selection are always detected.

3. **Kept `activeLabel` Interface Unchanged**  
   - The original behavior of `activeLabel` is preserved.
   - Ensures backward compatibility with existing components/modules that rely on it.

---

### 🔍 Problem Solved

- Fix #7590 
- Users can now add new labels while keeping the first selected item unchanged, and the waveform regions will still reflect the updated label set.
- Improves UI consistency and responsiveness to user actions.

screen_recording_2025-05-21_18-02-43

---

### 📌 Impact Scope

- Only affects the label synchronization logic in the audio annotation module.
- No breaking changes to external interfaces or other modules.

---

### 🧪 How to Test

1. Load an audio annotation task in the UI.
2. Add multiple labels and select one of them.
3. Add a new label while keeping the previously selected label still selected.
4. Verify that the selected regions in the waveform are updated with the latest label.


yyassi-heartex (approved):

yyassi-heartex:
/jira create FIT
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15211021634)
> Jira issue [TRIAG-1386](https://humansignal.atlassian.net/browse/TRIAG-1386) is created

[TRIAG-1386]: https://humansignal.atlassian.net/browse/TRIAG-1386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hlomzik:
Hey, @xxxinsxxx, what exactly is the bug here?
When you have audio config with `Labels choice="multiple"` the color of the region equals the color of the first label assigned to this region. And it reflects changes of labels set for me. Is not it the case for you? Can you record a video with the buggy behaviour?

xxxinsxxx:
> Hey, @xxxinsxxx, what exactly is the bug here? When you have audio config with `Labels choice="multiple"` the color of the region equals the color of the first label assigned to this region. And it reflects changes of labels set for me. Is not it the case for you? Can you record a video with the buggy behaviour?

Hi @hlomzik , thanks for checking! The core issue isn't about colors but missing text for multi-select labels. Details are in issue #7590 .

hlomzik:
Ah great, makes sense! Could you please also add text "Fixes #7590" to PR description?

hlomzik (approved):

xxxinsxxx:
> Ah great, makes sense! Could you please also add text "Fixes #7590" to PR description?

Added! 😊 Thanks!

PR #7607, opened 2025-05-23
fix: BROS-8: 504 Gateway Timeout when Fetching Tasks via API using Filters Involving "Updated at"

makseq:
This problem happens when there are huge amount of annotations per a task and a user tries to filter by Update at column in the data manager.


triklozoid (approved):
It will be nice to test it on db snapshot. Annotations is a large table and we already had a lot of problem with it.

PR #7609, opened 2025-05-23
fix: FIT-155: Followup fixes to address any issues noticed in docs

bmartel:
Brush docs not correct for current application.
Launch in playground not using the value from url param as precedence.

### DIFF
--- a/docs/source/playground/index.ejs
+++ b/docs/source/playground/index.ejs
@@ -442,11 +442,19 @@ meta_description: Label Studio interactive demo and playground for data labeling
     return "https://label-studio-playground.netlify.app/";
   }
 
+  function getStartingConfigValue() {
+    const urlParams = new URLSearchParams(window.location.search);
+    const config = urlParams.get('config');
+    if (config) {
+      return config;
+    }
+    const startTemplate = document.getElementById('start-template');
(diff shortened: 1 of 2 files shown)


docs/source/playground/index.ejs
```diff
@@ -442,11 +442,19 @@ meta_description: Label Studio interactive demo and playground for data labeling
...
   }
 
+  function getStartingConfigValue() {
+    const urlParams = new URLSearchParams(window.location.search);
+    const config = urlParams.get('config');
+    if (config) {
+      return config;
+    }
+    const startTemplate = document.getElementById('start-template');
+    return encodeURIComponent(startTemplate.innerHTML.replace(/(\r\n|\n|\r)/gm, "<br>"));
```
hlomzik, line 452: I'm curious how does this work with multiline... Like should not it break the config? And how do we work with `config` from url — is it okay to not transform newlines there? If it's ok like this maybe we can also get rid of this transform as well?

hlomzik (approved):
Curious about newline processing, but we had this before, so it's more like a possible improvement.

docs/source/playground/index.ejs
bmartel: Likely, I will try and make another PR if so.

PR #7610, opened 2025-05-23
fix: FIT-155: Better handling for config encoding/decoding in docs playground

bmartel:
Normalize newlines, and encode the config with standard newlines.
Also fixes an issue with the collision of a redefinition of labelStudioPlaygroundUrl which is possible due to the templates and included scripts.

### DIFF
--- a/docs/source/playground/index.ejs
+++ b/docs/source/playground/index.ejs
@@ -442,6 +442,14 @@ meta_description: Label Studio interactive demo and playground for data labeling
     return "https://label-studio-playground.netlify.app/";
   }
 
+  function normalizeNewlines(text) {
+    return text.replace(/(\r\n|\n|\r)/gm, "\n");
+  }
+
+  function encodeConfig(text) {
+    return encodeURIComponent(normalizeNewlines(text));
+  }
+
   function getStartingConfigValue() {
     const urlParams = new URLSearchParams(window.location.search);
     const config = urlParams.get('config');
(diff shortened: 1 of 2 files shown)


docs/source/playground/index.ejs
```diff
@@ -442,14 +442,22 @@ meta_description: Label Studio interactive demo and playground for data labeling
     return "https://label-studio-playground.netlify.app/";
   }
 
+  function normalizeNewlines(text) {
+    return text.replace(/(\r\n|\n|\r)/gm, "\n");
```
hlomzik: `\n` is redundant here I guess? it's not appearing in next options in this union, so should be safe to remove
```suggestion
    return text.replace(/(\r\n|\r)/gm, "\n");
```

docs/themes/v2/source/js/code-blocks-enhancements.js
```diff
@@ -10,7 +10,13 @@ function getLabelStudioPlaygroundUrl() {
   return "https://label-studio-playground.netlify.app/";
 }
 
-const labelStudioPlaygroundUrl = getLabelStudioPlaygroundUrl();
+function normalizeNewlines(text) {
+  return text.replace(/(\r\n|\n|\r)/gm, "\n");
```
hlomzik: the same here
```suggestion
  return text.replace(/(\r\n|\r)/gm, "\n");
```

hlomzik (approved):
only small suggestion

docs/source/playground/index.ejs
bmartel: that is true, yep let me remove it from the union options.

PR #7613, opened 2025-05-23
feat: ROOT-16: Support reading Parquet from source cloud storages in LSE

matt-bernstein:
fix test mock


label_studio/tests/utils.py
```diff
@@ -124,7 +124,7 @@ def generate_signed_url(self, **kwargs):
         def download_as_bytes(self):
             data = f'test_blob_{self.key}'
             if self.is_json:
-                return json.dumps(self.sample_json_contents)
+                return json.dumps(self.sample_json_contents).encode('utf-8')
             return data
```
jombooth: should `data` also be returned encoded?

jombooth (approved):
just one question

label_studio/tests/utils.py
matt-bernstein: that codepath is never hit currently, but for consistency, yes

label_studio/tests/utils.py
```diff
@@ -124,8 +124,8 @@ def generate_signed_url(self, **kwargs):
         def download_as_bytes(self):
```
jombooth, line 124: slightly DRY-er approach

```
def download_as_bytes(self):
    data = json.dumps(self.sample_json_contents) if self.is_json else f'test_blob_{self.key}'
    return data.encode()
```

matt-bernstein: would agree in isolation, but want to make the parallel with `download_as_string` very obvious

jombooth: ah, even simpler then would be just:

```
def download_as_bytes(self):
    return self.download_as_string().encode()
```

matt-bernstein: fair enough :)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15216702267)

jombooth:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15233912348)
> Successfully merged:  create mode 100644 docs/source/guide/labeling_bulk.md

PR #7623, opened 2025-05-26
fix: FIT-136: Duration field at Info panel is not optimized for Dark mode

yyassi-heartex:
This pull request updates the styling of the `TimelineRegionEditor` component to improve consistency and accessibility. The changes include replacing hardcoded color variables with standardized design tokens and enhancing the behavior of read-only input fields.

after update light mode
image

after update dark mode
image

Styling updates:

* Replaced `--sand_300` and `--sand_100` with standardized design tokens `--color-neutral-border` for border and background colors to ensure consistency across the application.

Behavior enhancements:

* Updated read-only input fields to include a `cursor: default` property and removed the focus outline for better user experience.

### DIFF
--- a/web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
+++ b/web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
@@ -12,7 +12,7 @@
 
 .input {
   display: block;
-  border: 1px solid var(--sand_300);
+  border: 1px solid var(--color-neutral-border);
   border-radius: 4px;
   padding: 0 0 0 8px;
   width: 80px;
@@ -20,7 +20,12 @@
   text-align: right;
 
   &[readonly] {
-    background-color: var(--sand_100);
+    background-color: var(--color-neutral-border);
     color: var(--color-neutral-content-subtler);
+    cursor: default;
+
+    &:focus {
+      outline: none;
+    }
   }
 }
--- a/web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
+++ b/web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
@@ -20,8 +20,7 @@
   text-align: right;
 
   &[readonly] {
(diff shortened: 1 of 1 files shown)


web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
```diff
@@ -12,15 +12,20 @@
...
+  border: 1px solid var(--color-neutral-border);
   border-radius: 4px;
   padding: 0 0 0 8px;
   width: 80px;
   font-size: 14px;
   text-align: right;
 
   &[readonly] {
-    background-color: var(--sand_100);
+    background-color: var(--color-neutral-border);
```
bmartel: This should be a different token for background, borders should only be semantic to border usage.

bmartel (changes requested):

web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
yyassi-heartex: yeah but the standard `--color-neutral-background` didnt make the field look like its read only, and there was no subtle or subtler version. border color gave the effect that I was looking for - we can sync to figure out a better token

hlomzik: palette token is better than wrong semantic token. that breaks the idea of semantics.

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15253797075)
> Successfully merged:  10 files changed, 22 insertions(+), 9 deletions(-)

web/libs/editor/src/components/SidePanels/DetailsPanel/TimelineRegionEditor.module.scss
yyassi-heartex: ended up going another route with maintaing the same background color as the other inputs but made the input text color more subtle to make it clear its a read only field. will update the description of the PR to show the update momentarily

bmartel (approved):

PR #7644, opened 2025-05-28
fix: UTC-22: Validation error when updating task data via API

mcanu:
There was an issue where an existing project with tasks, changed labeling config. When updating the data of the tasks through the API it failed because the current task data was checked instead of the new data. This PR makes the TaskValidator validates the new data when updated instead of the current one.


hakan458 (approved):

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15350785737)

PR #7662, opened 2025-05-30
fix: BROS-38: Fix sentry issue (i.map is not a function) +1 more

hlomzik:
Two sentry issues are fixed here:

1. Very often one with comments sdk. Fetch can return null and this case was not handled in comments sdk.
`TypeError: i.map is not a function`

3. Rare, but with app crash and easy to fix. OutlinerTree can't process regions list and crashes the app.
This happens when region list has groupping by label and there is a region with no label.
Empty labels should be fixed properly later, but the case is rare and we are fixing app crash here.
`Cannot read properties of undefined (reading 'locked')`


bmartel (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15351508396)
> **Error**: Workflow failed

nikitabelonogov:
/git merge


> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15351709519)
> Successfully merged:  create mode 100644 label_studio/tasks/tests/test_api.py

PR #7664, opened 2025-05-30
chore: FIT-174: Refactor DataManager Table out into reusable component

yyassi-heartex:
this is a cherry pick of a change that went into develop on May15 - refactoring the datamanger table and adding margin at the bottom to prevent last row from clipping

### DIFF
--- a/web/libs/datamanager/src/components/Common/Table/Table.jsx
+++ b/web/libs/datamanager/src/components/Common/Table/Table.jsx
@@ -1,15 +1,12 @@
 import { observer } from "mobx-react";
 import { createContext, forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
-import AutoSizer from "react-virtualized-auto-sizer";
-import { VariableSizeList } from "react-window";
-import InfiniteLoader from "react-window-infinite-loader";
 import { useSDK } from "../../../providers/SDKProvider";
 import { isDefined } from "../../../utils/utils";
 import { Button } from "../Button/Button";
 import { Icon } from "../Icon/Icon";
 import { modal } from "../Modal/Modal";
 import { IconCode, IconGear, IconGearNewUI } from "@humansignal/icons";
-import { Tooltip } from "@humansignal/ui";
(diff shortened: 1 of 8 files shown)


web/libs/datamanager/src/components/Common/Table/Table.jsx
```diff
@@ -386,34 +383,19 @@ const StickyList = observer(
...
-        </AutoSizer>
+        <AutoSizerTable
+          ref={listRef}
+          totalCount={totalCount}
+          loadMore={loadMore}
+          isItemLoaded={isItemLoaded}
+          itemData={itemData}
+          itemSize={itemSize}
+          initialScrollOffset={initialScrollOffset}
+          className={tableCN.elem("auto-size").toString()}
```
nass600, line 394: If we foresee a chance that we might need to pass className changes to inner components I will consider prefixing the `className`. If we don't forget about this or we can tackle it in the future

web/libs/datamanager/src/components/Common/Table/TableHead/TableHead.scss
```diff
@@ -10,6 +10,7 @@
   font-size: 14px;
   border-bottom: 1px solid var(--color-neutral-border);
   color: var(--color-neutral-content);
+  position: sticky;
```
nass600, line 13: All our tables will had a sticky header? Sounds good to me but not sure if there will be any case to consider this to be customizable? cc: @ricardoantoniocm

web/libs/datamanager/src/components/MainView/DataView/Table.scss
```diff
@@ -4,6 +4,7 @@
   overflow: hidden;
   flex-direction: column;
   background: var(--color-neutral-background);
+  margin-bottom: 2.5rem;
```
nass600, line 7: If we are considering Table a design system atom we should never add anything beyond its bounding box, in this case margins. The spacing should be allocated by the parent

nass600 (commented):
Very happy to see this moving on 👏 . Just added minor comments

nass600 (approved):

PR #7676, opened 2025-06-02
feat: UTC-50: Add session timeout policy configuration

hakan458:
Adds new model `SessionTimeoutPolicy` which allows an Organization to set custom `max_session_age` and `max_time_between_activity` in hours.


label_studio/session_policy/models.py
```diff
@@ -0,0 +1,27 @@
+from annoying.fields import AutoOneToOneField
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from organizations.models import Organization
+
+
+class SessionTimeoutPolicy(models.Model):
```
mcanu, line 7: Why create a new model and app and not add two new fields to the Organization model? 
An alternative option would be to consider creating an `OrganizationSettings` model (which I think its what part of `OrganizationBilling` is being used for, but there isn't one in LSO).
Also, adding a new model adds an extra query to all requests passing through the middleware, instead of the already fetched organization.

hakan458: It seems that none of the org level settings are stored in the Organization model itself... for example jwt settings, billing settings, etc. They are all separate models so I am just following the same pattern. Allows this model to be more flexible without changing the organization model. My first thought was to add it directly to the Org as well but it seems we don't do that for anything we have

mcanu (approved):

label_studio/session_policy/models.py
mcanu: yeah, I don't really have a strong opinion about this and it works.

PR #7681, opened 2025-06-03
fix: BROS-71: Add playback cursor for timeseries sync

makseq:
This PR adds a playback cursor. 

https://github.com/user-attachments/assets/944f6375-147c-48dc-94df-511c01ad3240


yyassi-heartex:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15420639371)

web/libs/editor/src/tags/object/TimeSeries.jsx
```diff
@@ -1068,6 +1109,17 @@ const Overview = observer(({ item, data, series }) => {
...
     gb.current.select(".handle--w").style("transform", "translate(-1px, 0)");
     gb.current.select(".handle--e").style("transform", "translate(1px, 0)");
+
+    // Playhead cursor line
+    cursorLine.current = focus.current
+      .append("line")
+      .attr("class", "overview-playhead")
+      .attr("y1", 0)
+      .attr("y2", focusHeight)
+      .attr("stroke", item.cursorcolor || "#1e90ff")
```
bmartel:
```suggestion
      .attr("stroke", item.cursorcolor || "var(--color-neutral-inverted-surface)")
```
@ricardoantoniocm what is the semantic default color we would use here for a playhead?

web/libs/editor/src/tags/object/TimeSeries.jsx
```diff
@@ -89,6 +90,9 @@ const TagAttrs = types.model({
   // visibilitycontrols: types.optional(types.boolean, false), // show channel visibility controls
   hotkey: types.maybeNull(types.string),
   sync: types.maybeNull(types.string),
+
+  // Color of the playback cursors (hex or any SVG-compatible color string)
+  cursorcolor: types.optional(types.string, "#1e90ff"),
```
bmartel:
```suggestion
  cursorcolor: types.optional(types.string, "var(--color-neutral-inverted-surface)"),
```

@ricardoantoniocm what is the semantic default color we would use here for a playhead?

web/libs/editor/src/tags/object/TimeSeries/Channel.jsx
```diff
@@ -837,6 +847,47 @@ class ChannelD3 extends React.Component {
...
+   */
+  renderPlayhead = () => {
+    if (this.playhead) return; // already rendered
+
+    this.playhead = this.main
+      .append("line")
+      .attr("class", "playhead")
+      .attr("y1", 0)
+      .attr("y2", this.height)
+      .attr("stroke", this.props.item.parent?.cursorcolor || "#1e90ff")
```
bmartel:
```suggestion
      .attr("stroke", this.props.item.parent?.cursorcolor || "var(--color-neutral-inverted-surface)")
```
@ricardoantoniocm here as well

cloudmark:
Spoke to @makseq and we agreed that we would tackle the dragging of the play heads later. This is so that this feature is in line with the other components like video and audio.  Great stuff @makseq , this was needed.

web/libs/editor/src/tags/object/TimeSeries.jsx
ricardoantoniocm: We should use `--color-neutral-inverted-surface` in this instance.

This ensures consistency with the playhead of the audio player.
image

Red usually carries a negative/destructive connotation so we should avoid it unless there's a very good reason for it.

web/libs/editor/src/tags/object/TimeSeries.jsx
ricardoantoniocm: We should use `--color-neutral-inverted-surface` in this instance.

web/libs/editor/src/tags/object/TimeSeries/Channel.jsx
ricardoantoniocm: We should use `--color-neutral-inverted-surface` in this instance.

Gondragos (approved):

PR #7684, opened 2025-06-04
fix: BROS-65: Persistent storage use `attachment` instead of `inline` and can't load PDF

makseq:
**Original problem:** 
Persistent storage uses `attachment` instead of `inline` as the content disposition. Because of this, users experience a problem: instead of displaying the PDF within the labeling interface, the browser starts downloading the file as an attachment.

**Follow-up problem**
In order to debug the issue, I needed to deploy Label Studio with persistent storage; however, debugging with nginx proved to be inconvenient. Additionally, I was aware that some users encountered issues when attempting to deploy Label Studio with persistent storage: essentially, persistent storage was not fully functional, preventing them from downloading any media. I now understand the reason: Label Studio does not support persistent storage without nginx. This pull request addresses the problem, enabling Label Studio to function without nginx. To achieve this, users must configure `USE_NGINX_FOR_UPLOADS=false` and `USE_NGINX_FOR_EXPORT_DOWNLOADS=false` (which was previously supported).


label_studio/data_import/api.py
```diff
@@ -706,43 +708,79 @@ def get(self, *args, **kwargs):
...
+      3. NGINX intercepts and serves the file directly from storage
+    - **Benefits**: Reduces Django server load, better performance for large files
+
+    ### 2. Direct Mode (USE_NGINX_FOR_UPLOADS=False)
+    - **Direct Serving**: Django serves files using RangedFileResponse
+    - **How it works**:
+      1. Validates user permissions and file access
+      2. Opens file from storage and streams it with range request support
+      3. Supports partial content requests (HTTP 206)
+    - **Benefits**: Works without NGINX, supports range requests for media files
```
triklozoid, line 734: > supports range requests for media files

I don't believe nginx doesn't now support range requests,maybe we need to configure something. 
This option, for example: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_force_ranges

label_studio/tests/data_import/test_persistent_storage_data.py
```diff
@@ -0,0 +1,336 @@
...
+
+class TestDownloadStorageData:
+    @pytest.fixture
+    def api_factory(self):
+        return APIRequestFactory()
+
+    @pytest.fixture
+    def user(self):
+        """Create a test user with organization"""
+        org = Organization.objects.create(title='Test Org')
```
triklozoid, line 75: I think we need to use either API or factories for data preparation in the tests, not the direct orm requests.

triklozoid (approved):

Issue #6594, opened 2024-11-02
Number tag not working with any value of `step`.


ptkenny:
Config:
```html
<View>
  <Image name="image" value="$image"/>
  <RectangleLabels name="face_label" toName="image">
    <Label value="Face"/>
  </RectangleLabels>
  <Number name="face_rating" toName="image" min="0.0" max="10.0" step="0.1" defaultValue="5.0" perRegion="true"/>
  <Choices name="valid_image" toName="image" required="true">
    <Choice value="Invalid Image (for whatever reason)"/>
  </Choices>
</View>
```

Currently, no numbers are accepted at any step for any labeled region. Screenshot:
![image](https://github.com/user-attachments/assets/e7bbdb65-1df7-4919-9c9d-9f813a5c65bd)

Context and validation by Caitlin on Slack: https://label-studio.slack.com/archives/CQ8LYQ0J2/p1730563809608669?thread_ts=1730562724.748609&cid=CQ8LYQ0J2

caitlinwheeless:
Note that I was able to recreate this and also got some strange behavior. 

For example, 1.6 is accepted. 

1.7 results in the following error: 

`Number 1.7 is not valid. The two nearest valid  values are 1.6 and 1.70000000000000000002. `

heidi-humansignal:
Hi there,

Thanks for opening up an issue! I followed the Slack thread with Caitlin, and it looks like this is similar to an issue that we had worked on somewhat recently.  I'll check with our engineering team to see if the fix they implemented is available in the latest open source version of Label Studio.

Just to confirm and ensure that I give as much context to our team as possible, what version of Label Studio are you currently running?

Thanks,

Tyler Conlee
Head of Support
HumanSignal

> Comment by Tyler Conlee
> [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/11709234694)

nass600:
The issue was coming from a strict floating-number comparison which is now fixed

PR #7690, opened 2025-06-04
fix: UTC-74: [FE] Error Message on Numeric Value Field

nass600:
Fixes the `<Number/>` tag validation by introducing a small EPSYLON (1e-8) for the comparison when we are validating the input.
This is needed to bypass the imprecision generated by javascript when handling floating-number decimals form the binary source (IEEE 754). Example:
```
0.1 + 0.2 // returns 0.30000000000000004
```

Screenshot 2025-06-04 at 13 58 14


nick-skriabin (approved):

nass600:
Fixes #6594

PR #7691, opened 2025-06-04
fix: BROS-72: Fix grid view not updating when switching "Fit images to width"

nick-skriabin:
This pull request makes a minor update to the `GridWidthButton` component in `GridWidthButton.jsx`. The change improves the `handleResponsiveImagesToggle` function by adding `view` to its dependency array, ensuring proper reactivity.

### DIFF
--- a/web/libs/datamanager/src/components/DataManager/Toolbar/GridWidthButton.jsx
+++ b/web/libs/datamanager/src/components/DataManager/Toolbar/GridWidthButton.jsx
@@ -15,12 +15,12 @@ const injector = inject(({ store }) => {
     view,
     isGrid: view.type === "grid",
     gridWidth: view?.gridWidth,
-    responsiveImage: view?.gridResponsiveImage,
+    fitImageToWidth: view?.gridFitImageToWidth,
     hasImage,
   };
 });
 
-export const GridWidthButton = injector(({ view, isGrid, gridWidth, responsiveImage, hasImage, size }) => {
+export const GridWidthButton = injector(({ view, isGrid, gridWidth, fitImageToWidth, hasImage, size }) => {
   const [width, setWidth] = useState(gridWidth);
 
   const setGridWidth = useCallback(
@@ -33,9 +33,12 @@ export const GridWidthButton = injector(({ view, isGrid, gridWidth, responsiveIm
     [view],
   );
 
-  const handleResponsiveImagesToggle = useCallback((e) => {
-    view.setGridResponsiveImage(!e.target.checked);
-  }, []);
+  const handleResponsiveImagesToggle = useCallback(
+    (e) => {
+      view.setgridFitImageToWidth(e.target.checked);
+    },
+    [view],
+  );
 
   return isGrid ? (
     <Dropdown.Trigger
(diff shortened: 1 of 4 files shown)


web/libs/datamanager/src/stores/Tabs/tab.js
```diff
@@ -290,8 +290,8 @@ export const Tab = types
       self.save();
     },
 
-    setGridResponsiveImage(responsive) {
-      self.gridResponsiveImage = responsive;
+    setgridFitImageToWidth(responsive) {
```
bmartel:
```suggestion
    setGridFitImageToWidth(responsive) {
```

web/libs/datamanager/src/components/DataManager/Toolbar/GridWidthButton.jsx
```diff
@@ -33,9 +33,12 @@ export const GridWidthButton = injector(({ view, isGrid, gridWidth, responsiveIm
     [view],
   );
 
-  const handleResponsiveImagesToggle = useCallback((e) => {
-    view.setGridResponsiveImage(!e.target.checked);
-  }, []);
+  const handleResponsiveImagesToggle = useCallback(
+    (e) => {
+      view.setgridFitImageToWidth(e.target.checked);
```
bmartel:
```suggestion
      view.setGridFitImageToWidth(e.target.checked);
```

bmartel (commented):
I don't quite understand this change, we renamed the property but I don't see how this fixes anything, seems more or less equivalent to before?

web/libs/datamanager/src/components/DataManager/Toolbar/GridWidthButton.jsx
```diff
@@ -33,9 +33,12 @@ export const GridWidthButton = injector(({ view, isGrid, gridWidth, responsiveIm
...
   );
 
-  const handleResponsiveImagesToggle = useCallback((e) => {
-    view.setGridResponsiveImage(!e.target.checked);
-  }, []);
+  const handleResponsiveImagesToggle = useCallback(
+    (e) => {
+      view.setGridFitImageToWidth(e.target.checked);
+    },
+    [view],
```
nick-skriabin, line 40: @bmartel here's the actual fix. view dep was missing. the rest is to just align the naming and disable old behavior by default

bmartel: Can we just make the fix verbatim, I don't quite see the need for all of these other changes.

yyassi-heartex: if we can point to a specific change and say that it addresses the issue at hand then that should be the PR - we should get into the habit of breaking down our PRs to help delivery times by reducing complexity and review times

web/libs/datamanager/src/components/MainView/GridView/GridView.scss
```diff
@@ -52,7 +52,7 @@
         display: grid;
         grid-auto-flow: row;
         grid-template-rows: minmax(1rem, 1fr);
-      grid-template-columns: 100%;
```
yyassi-heartex: im surprised this change made it to develop - wouldnt have this triggered a lint error?

web/libs/datamanager/src/components/DataManager/Toolbar/GridWidthButton.jsx
nick-skriabin: ok

bmartel (approved):

web/libs/datamanager/src/components/MainView/GridView/GridView.scss
nick-skriabin: apparently, no

yyassi-heartex (approved):

nass600 (approved):

PR #7692, opened 2025-06-04
fix: BROS-29: Fix Collapse with Interactive View All

hlomzik:
This tag didn't have AnnotationMixin and couldn't use annotation to render content. That's important when DEV-3391 FF is ON.
`forceRender` fixes the issue when this flag is OFF and lazily rendered objects inside Panel don't enable isReady flag.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15566028771)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/timeseries/multichannel.cy.ts

web/libs/editor/src/tags/visual/Collapse.jsx
```diff
@@ -126,7 +126,7 @@ const HtxCollapse = observer(({ item }) => {
       {item.children
         .filter((i) => i.type === "panel" && (!isBulkMode || i.isIndependent))
         .map((i) => (
-          <Panel key={i._value} header={i._value}>
+          <Panel key={i._value} header={i._value} forceRender>
```
yyassi-heartex, line 129: forceRender will apply regardless of FF state - is that intended?

hlomzik: that's the parameter of antd component and we never had intention to lazily load content within it, should be safe to use it. theoretically someone might rely on lazy loading, but we never promised it, it's not very supported.

yyassi-heartex (approved):
approved but looks like the e2e test needs to be updated to match the new expectations

bmartel (approved):

nass600 (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15587198205)
> Successfully merged:  2 files changed, 19 insertions(+), 3 deletions(-)

PR #7696, opened 2025-06-04
feat: UTC-70: DeletedRow functionality

mcanu:
Adds DeletedRow model, to enable serializing model rows we need to delete but want to keep stored just in case.


label_studio/core/models.py
```diff
@@ -42,3 +44,39 @@ class AsyncMigrationStatus(models.Model):
...
         return f'(id={self.id}) ' + self.name + (' at project ' + str(self.project) if self.project else '')
+
+
+class DeletedRow(models.Model):
+    """
+    Model to store deleted rows of other models.
+    Useful for using as backup for deleted rows, in case we need to restore them.
+    """
+
+    model = models.CharField(max_length=255)   # tasks.task, projects.project, etc.
```
jombooth: one little suggestion here - we have some fairly long model paths already, and this field won't be populated based on user input. I think we can safely use TextField instead to get rid of the max_length, or just set it to something much larger like 1024.

jombooth (approved):
awesome, I particularly like the `bulk_serialize_and_create` class method you provided 🔥

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15451630590)

label_studio/core/models.py
```diff
@@ -42,3 +44,39 @@ class AsyncMigrationStatus(models.Model):
...
+    updated_at = models.DateTimeField(auto_now=True)
+
+    # optional fields for searching purposes
+    organization_id = models.IntegerField(null=True, blank=True)
+    project_id = models.IntegerField(null=True, blank=True)
+    user_id = models.IntegerField(null=True, blank=True)
+
+    @classmethod
+    def serialize_and_create(cls, model, **kwargs) -> 'DeletedRow':
+        data = json.loads(serializers.serialize('json', [model]))[0]
```
hakan458, line 69: this works for any model, and retains all fields, etc? Is there anything that would not be supported?

hakan458: Also have you tried deserializing?

jombooth: To this point, it would be great to test `Comment` specifically since that's our most immediate purpose for this. No need to write a unit test - just checking it locally is fine.

mcanu: I've added deserialization to tests

mcanu: @jombooth tested it with Comment and it works 👍

hakan458 (approved):

PR #7698, opened 2025-06-04
fix: BROS-81: UI. Enhanced control for enabling storage proxy

makseq:
This PR adds improved control for enabling storage proxies and provides more detailed descriptions to the toggles in storage settings.

|Old|New|
|---|---|
| img1 | img2 |
| image |image |
| image | image |


yyassi-heartex (approved):
i like the updated layout

PR #7702, opened 2025-06-05
fix: BROS-66: Fix Syncable in Interactive View All

hlomzik:
Another try to fix this.
Synced tags create SyncManager instance for every group in every annotation. With DEV-3391 FF OFF we had only one config tree and all tags in group were refering to the same manager.
With DEV-3391 FF ON we have separate config trees so we should distinguish managers.

BUT
History items also trigger `afterCreate()` from `SyncableMixin`, but we don't have annotation there.
One of the ways to fix this is to modify AnnotationMixin to retrieve HistoryItem in such cases.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15475837322)
> Successfully merged:  create mode 100644 label_studio/core/tests/test_models.py

Gondragos (approved):

Gondragos (approved):

web/libs/editor/src/mixins/Syncable.ts
```diff
@@ -136,7 +138,20 @@ const SyncableMixin = types
...
-      self.syncManager = SyncManagerFactory.get(self.sync, self.name);
+      let sync = self.sync;
+      let fallbackSync = self.name;
+
+      if (ff.isActive(FF_DEV_3391)) {
+        if (!self.annotationStore.initialized) return;
+
+        // different annotations have their own independent trees and should have independent
+        // sync managers; also history items are also independent, so should have the same
+        const postfix = `@${self.annotationOrHistoryItem?.id}`;
```
bmartel, line 149:
```
        const postfix = `@${self.annotationOrHistoryItem?.id}`;
```
If the annotationOrHistoryItem is null or undefined, won't this break? Can it be expected to exist?

bmartel (approved):

web/libs/editor/src/mixins/Syncable.ts
hlomzik: The worst case — tags won't sync because sync manager will point to a wrong tag (that's actually the current state with FF ON)
But in practice this should not happen — Syncable is a mixin for tags and `afterCreate()` will be called during `initRoot()` for annotation or history item. It could theoretically be invoke for the whole annotation store `initRoot()`, but this tree won't be used anyway.

PR #7746, opened 2025-06-11
fix: TRIAG-1357: Hotkey not working for Number Tag

luarmr:
Currently hotkeys are not working properly for input number elements.

The origin of the problem is that the method falls back to 0 when max is not defined. But when working on it, I realized that the `increaseValue` method does not:
- Respect minimum value (for example, when 0 is lower than the minimum value)
- Work when maximum is not defined
- Increase according to the step parameter
- Use the default value as the initial value

As extreme scenario to consider.  min=.1 max=.7 step=.1 defaultvalue=.5

A better fix would likely be to sanitize the value when the tag is created. However, since I couldn't identify the pattern and it appears that only this isolated method is affected, I decided not to overcomplicate the component.

tested on cases like:

```

<View>
  <!-- Test area for shortcuts -->
  <TextArea name="text1" toName="text" />
  <TextArea name="text2" toName="text" />
  
  <!-- Text content -->
  <Text name="text" value="$text" />
  
  <!-- Number variations -->
  <Header value="Basic number - no parameters" />
  <Number name="number_basic" toName="text" />
  
  <Header value="Number with min=5 - should not go below 5" />
  <Number name="number_min" toName="text" min="5" />
  
  <Header value="Number with max=10 - should not go above 10" />
  <Number name="number_max" toName="text" max="10" />
  
  <Header value="Number with step=2 - should increment by 2" />
  <Number name="number_step" toName="text" step="2" />
  
  <Header value="Number with defaultvalue=3 - should start at 3" />
  <Number name="number_default" toName="text" defaultvalue="3" />
  
  <Header value="Complete configuration - min=0, max=100, step=1, defaultvalue=2" />
  <Number name="number_complete" toName="text" min="0" max="100" step="1" defaultvalue="2" />
  
  <Header value="Number with hotkey='n' - press 'n' to increment" />
  <Number name="number_hotkey" toName="text" hotkey="n" />
  
  <Header value="Required number - must be filled" />
  <Number name="number_required" toName="text" required="true" />
  
  <Header value="Number with slider - visual slider control" />
  <Number name="number_slider" toName="text" slider="true" min="0" max="10" />
</View>
```

@hlomzik can you take a look to this?

### DIFF
--- a/web/libs/editor/src/tags/control/Number.jsx
+++ b/web/libs/editor/src/tags/control/Number.jsx
@@ -14,6 +14,7 @@ import ClassificationBase from "./ClassificationBase";
 import PerItemMixin from "../../mixins/PerItem";
 import { FF_LSDV_4583, isFF } from "../../utils/feature-flags";
 import { cn } from "../../utils/bem";
+import { safeNumber, positiveNumber } from "../../utils/number";
 
 /**
  * The Number tag supports numeric classification. Use to classify tasks using numbers.
@@ -171,15 +172,15 @@ const Model = types
       },
 
       increaseValue() {
-        if (self.number >= Number(self.max)) {
-          self.setNumber(0);
-        } else {
-          if (self.number > 0) {
-            self.setNumber(self.number + 1);
-          } else {
-            self.setNumber(1);
-          }
-        }
+        const min = safeNumber(self.min, 0);
+        const max = safeNumber(self.max, Number.POSITIVE_INFINITY);
+        const step = positiveNumber(self.step);
+        const defaultValue = safeNumber(self.defaultvalue, min);
+        const current = safeNumber(self.number, defaultValue);
+
+        const next = current + step > max ? min : current + step;
+
+        self.setNumber(next);
       },
 
       onHotKey() {
--- a/web/libs/editor/src/utils/number.js
+++ b/web/libs/editor/src/utils/number.js
@@ -1,0 +1,11 @@
+import { isDefined } from "./utilities";
+
+export const safeNumber = (v, fallback) => {
+  const n = Number(v);
+  return isDefined(v) && Number.isFinite(n) ? n : fallback;
+};
+
+export const positiveNumber = (v, fallback = 1) => {
+  const n = Number(v);
+  return Number.isFinite(n) && n > 0 ? n : fallback;
+};


yyassi-heartex (approved):

web/libs/editor/src/tags/control/Number.jsx
```diff
@@ -171,15 +172,15 @@ const Model = types
...
-        } else {
-          if (self.number > 0) {
-            self.setNumber(self.number + 1);
-          } else {
-            self.setNumber(1);
-          }
-        }
+        const min = safeNumber(self.min, 0);
+        const max = safeNumber(self.max, Number.POSITIVE_INFINITY);
+        const step = positiveNumber(self.step);
```
hlomzik, line 177: I'd just make it unsigned, so `-1` will be `1`, which feels the same here

web/libs/editor/src/tags/control/Number.jsx
```diff
@@ -171,15 +172,15 @@ const Model = types
...
+        const defaultValue = safeNumber(self.defaultvalue, min);
+        const current = safeNumber(self.number, defaultValue);
+
+        const next = current + step > max ? min : current + step;
```
hlomzik, line 181: should not we hit `max` value and stop there and not falling back to `min`?
```suggestion
        const next = current + step > max ? max : current + step;
```

web/libs/editor/src/tags/control/Number.jsx
luarmr: in the html component any negative convert to 1 (tested in chrome) , that was my thought process. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number

>Note: When the data entered by the user doesn't adhere to the stepping configuration, the user agent may round to the nearest valid value, preferring numbers in the positive direction when there are two equally close options.

hlomzik: according to @luarmr `step` in html number input should be positive and will be switched to `1` if negative

web/libs/editor/src/tags/control/Number.jsx
luarmr: The current flow is to loop, so I replicate the action. I agree with the current loop behavior because it always provides feedback.
The current bug looks like a bug because the input stays at 0.
Tell me if you feel strong here.

@hlomzik  thanks so much for go deep in the review, awesome!

luarmr: Btw in this one I am 50% so I will apply your take. Since will happen after several iterations

hlomzik (approved):

web/libs/editor/src/tags/control/Number.jsx
luarmr: We discussed offline and we both are 50% on one direction over another, so we'll keep it as is for now.

PR #7752, opened 2025-06-12
fix: PLT-796: Prevent deadlocks with batch updates

triklozoid:
When multiple users are simultaneously working on projects with a large number of tasks, PostgreSQL deadlocks were occurring during massive tasks update operations. These deadlocks happen when the _rearrange_overlap_cohort() method tries to update thousands of task records in a single database
   transaction while other users are also modifying tasks in the same project.

  - Split large UPDATE operations into batches (500 records)
  - Added retry logic with exponential backoff (up to 3 attempts)
  - Updated _rearrange_overlap_cohort() to use batch updates

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -2,6 +2,7 @@
 """
 import json
 import logging
+import time
 from typing import Any, Mapping, Optional
 
 from annoying.fields import AutoOneToOneField
@@ -28,7 +29,7 @@ from core.utils.common import (
 from core.utils.db import fast_first
 from django.conf import settings
 from django.core.validators import MaxLengthValidator, MinLengthValidator
-from django.db import models, transaction
+from django.db import OperationalError, models, transaction
 from django.db.models import Avg, BooleanField, Case, Count, JSONField, Max, Q, Sum, Value, When
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


label_studio/projects/models.py
```diff
@@ -451,6 +452,47 @@ def _update_tasks_states(
...
+                    break
+                except OperationalError as e:
+                    if 'deadlock detected' in str(e):
+                        retry_count += 1
+                        if retry_count >= max_retries:
+                            logger.error(
+                                f'Failed to update batch after {max_retries} retries. '
+                                f'Project: {self.id}, Batch: {i}-{i+len(batch_ids)}, Error: {e}'
+                            )
+                            raise
```
jombooth: I think it would be better for readability to move this to after the while loop exits, so that you don't have to compare retry count and max retries twice

label_studio/tests/test_batch_update_deadlock.py
```diff
@@ -0,0 +1,67 @@
...
+
+        call_count = 0
+
+        original_batch_update = project._batch_update_with_retry
+
+        def mock_batch_update(queryset, *args, **kwargs):
+            nonlocal call_count
+            call_count += 1
+            if call_count == 1:
+                # Simulate deadlock on first call
```
jombooth, line 40: 😍

jombooth (approved):

label_studio/projects/models.py
```diff
@@ -451,6 +452,47 @@ def _update_tasks_states(
         elif tasks_number_changed and self.overlap_cohort_percentage < 100 and self.maximum_annotations > 1:
             self._rearrange_overlap_cohort()
 
+    def _batch_update_with_retry(self, queryset, batch_size=500, max_retries=3, **update_fields):
```
jombooth, line 454: one thought I had: would it be worth moving this to some other utils module where it can be shared? this could be used for other models than just Task; the model class could be a parameter.

label_studio/projects/models.py
triklozoid: Fixed.

label_studio/projects/models.py
triklozoid: Fixed.

jombooth (approved):

PR #7760, opened 2025-06-13
fix: BROS-100: Flip rectangle regions without shift

hlomzik:
If you start resizing Rectangle region from left to right effectively flipping it and if you move the cursor very fast, the region will shift during flip, because Konva.js flips rectangle sides at some point.

Konva.js Transformer has parameter `flipEnabled` which was disabled. This parameter fixes the issue but it flips the height of the region and also rotates it at some cases.
We have to fix negative height and shift the region back.

Rotation is kept intact except for the two most common cases when it stays 0:
- when the region is flipped horizontally with no rotation, we fix the rotation back to 0.
- when the region is flipped vertically, rotation is still 0, we just flip the height.

All other cases might have rotated upside down, but sizes and position are correct, so that's acceptable.

Before:

https://github.com/user-attachments/assets/a2ca234a-d740-40ac-9762-1211d118adda

With `flipEnabled` but without fix, notice negative height and flipped rotation:

https://github.com/user-attachments/assets/cd1e8f6c-f829-4187-b2de-a71659e79c94

After the fix, height is positive:

https://github.com/user-attachments/assets/0651881a-7a77-4c1a-8db3-7d3831fcf913


hlomzik:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15643836190)

nick-skriabin (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15680751302)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/timeseries/multichannel.cy.ts

Gondragos (approved):

Issue #7761, opened 2025-06-14
Spectrogram Initial Draw


cloudmark:
**Describe the bug**
The spectrogram visualization fails to render initially when loading the annotation interface. The spectrogram only appears after user interaction (scrolling or zooming). This is a race condition where the initial draw call happens before the component is fully ready to render.

**To Reproduce**
1. Open Label Studio
2. Load an audio annotation task
3. The spectrogram visualization should be visible immediately
4. Instead, the spectrogram area appears empty
5. The spectrogram only renders after scrolling or zooming

**Expected behavior**
The spectrogram should render immediately when the annotation interface loads, without requiring user interaction.

**Screenshots**
[Video demonstration](https://www.loom.com/share/de21ba688961416bbbf24c7a30f2940b)

**Environment (please complete the following information):**
- OS: macOS
- Label Studio Version: [current version]

**Additional context**
The issue appears to be a timing problem where the initial draw call happens before the component is fully initialized. The fix involves deferring the initial draw call by a small amount (10ms) to ensure all necessary initialization is complete:

```typescript
setTimeout(() => this.draw(), 10);
```

This suggests that there might be some asynchronous initialization that needs to complete before the first draw can succeed. The current fix works by giving the component time to fully initialize before attempting to render.

**Technical Details**
- The bug occurs in the spectrogram renderer component
- The fix suggests that the component's initialization and first draw call are not properly synchronized
- The 10ms delay allows for any pending initialization to complete before the first render attempt

**Potential Long-term Solutions**
1. Implement proper initialization state tracking
2. Use a more robust initialization sequence
3. Consider using a proper initialization promise/event system instead of setTimeout
4. Add proper error handling for failed initial renders

cloudmark:
Closing, we handled this in https://github.com/HumanSignal/label-studio/pull/7788

PR #7762, opened 2025-06-14
fix: FIT-173: Follow up fix: Spectrogram initial render

cloudmark:
# Fix: Spectrogram Initialization Issue

## Reason for Change
Currently, the spectrogram visualization does not appear immediately upon initialization unless there is user interaction (scrolling or zooming). This is a visual bug that affects the user experience as the spectrogram data is loaded but not rendered until interaction occurs.

## Solution
We've implemented a deferred draw mechanism that triggers the initial render of the spectrogram shortly after initialization. This ensures the spectrogram is visible immediately without requiring user interaction.

## Technical Details
- Added a small delay (100ms) after initialization before triggering the first draw
- This delay allows the canvas and WebGL context to be fully ready
- The draw is triggered automatically without requiring user interaction
- Subsequent draws continue to work as before with user interactions

## Testing
1. Verified the spectrogram appears immediately upon initialization
2. Confirmed that subsequent interactions (zoom/scroll) continue to work normally
3. Tested across different browsers (Chrome, Firefox, Safari)
4. Verified performance impact is minimal

## Visual Evidence
- [Before the fix](https://www.loom.com/share/de21ba688961416bbbf24c7a30f2940b): Shows the spectrogram not appearing until user interaction
- [After the fix](https://www.loom.com/share/7a3ab437447043e58167fef39f9e6412): Demonstrates the spectrogram appearing immediately upon initialization

## Risks
- Low risk change as it only affects the initial render timing
- No impact on data processing or core functionality
- Small performance overhead from the additional draw call, but negligible

## Rollout Strategy
- This is a bug fix that can be deployed directly
- No feature flags or environment variables required
- Can be rolled out with normal deployment process

## Reviewer Notes
- Focus on the timing of the initial draw call
- Verify that the spectrogram appears immediately on page load
- Check that the solution doesn't introduce any performance issues
- Review the video evidence to confirm the fix works as expected

## Related Issue
https://github.com/HumanSignal/label-studio/issues/7761

## General Notes
This fix improves the user experience by ensuring the spectrogram is visible immediately upon initialization, making the visualization more intuitive and reducing potential user confusion about whether the data has loaded correctly.


makseq (approved):

web/libs/editor/src/lib/AudioUltra/Visual/Visualizer.ts
```diff
@@ -253,7 +253,7 @@ export class Visualizer extends Events<VisualizerEvents> {
     }
 
     this.invoke("initialized", [this]);
-    this.transferImage();
+    setTimeout(() => this.draw(), 10);
```
bmartel, line 256: Instead of binding to a timeout of 10ms, why not just leave it to the next animation frame?
```suggestion
    requestAnimationFrame(() => this.draw());
```

cloudmark: yes absolutely! great idea!

cloudmark:
Tested it out and this  does not work @bmartel , we need the timeout to defer a bit later, best to keep my changes for consistency

bmartel (approved):

PR #7770, opened 2025-06-16
feat: BROS-107: Visually hide Annotation ID on Annotation History empty state

ricardoantoniocm:
### Reason for change
In the Annotation History panel, the "Annotation ID" header was still visible even when there were no annotations to display. This change visually hides the header in the empty state to create a cleaner and more intuitive user interface, aligning with the intended design. The Annotation ID remains accessible to screen readers to maintain accessibility standards.

### Screenshots
**Before**
image

**After**
image

### Testing
1.  Navigate the Labeling UI, and open the History panel.
2.  Ensure there are no annotation changes (create a new one), so the panel displays its empty state.
3.  **Verify** that the "Annotation ID" header is not visible.
4.  Using browser developer tools, inspect the element to **confirm** that the header is still present in the DOM.
5.  Create a new annotation.
6.  **Verify** that the "Annotation ID" and its value are now visible in the history panel.

### General notes
This change improves the overall user experience by decluttering the interface in its empty state.

### DIFF
--- a/web/libs/editor/src/components/CurrentEntity/AnnotationHistory.tsx
+++ b/web/libs/editor/src/components/CurrentEntity/AnnotationHistory.tsx
@@ -17,7 +17,7 @@ import {
 } from "@humansignal/icons";
 import { Tooltip, Userpic } from "@humansignal/ui";
 import { Space } from "../../common/Space/Space";
-import { Block, Elem } from "../../utils/bem";
+import { Block, cn, Elem } from "../../utils/bem";
 import { humanDateDiff, userDisplayName } from "../../utils/utilities";
 import { EmptyState } from "../SidePanels/Components/EmptyState";
 import "./AnnotationHistory.scss";
@@ -142,7 +142,13 @@ const AnnotationHistoryComponent: FC<any> = ({
   if (shouldShowEmptyState) {
     return (
       <Block name="annotation-history" mod={{ inline, empty: true }}>
-        {sectionHeader && <div className="lsf-annotation-history__section-head sr-only">{sectionHeader}</div>}
+        {sectionHeader && (
+          <div
+            className={`${cn("annotation-history").elem("section-head").toString()}${showEmptyState ? " sr-only" : ""}`}
+          >
+            {sectionHeader}
+          </div>
+        )}
         {renderEmptyState ? renderEmptyState() : defaultEmptyState}
       </Block>
     );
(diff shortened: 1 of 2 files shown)


web/libs/editor/src/components/CurrentEntity/AnnotationHistory.tsx
```diff
@@ -142,7 +142,7 @@ const AnnotationHistoryComponent: FC<any> = ({
   if (shouldShowEmptyState) {
     return (
       <Block name="annotation-history" mod={{ inline, empty: true }}>
-        {sectionHeader && <Elem name="section-head">{sectionHeader}</Elem>}
+        {sectionHeader && <div className="lsf-annotation-history__section-head sr-only">{sectionHeader}</div>}
```
yyassi-heartex: This should do it
```suggestion
        {sectionHeader && <Elem name="section-head" rawClassName="sr-only">{sectionHeader}</Elem>}
```
Alternatively you can import `cn` and use it like this `cn('annotation-history').elem('section-head').toString()` then you can pass in sr-only as a classname with it too - this way  you dont need to manually put in the prefix (which could change in the future)

bmartel: This was to be my same feedback above ^.

bmartel: The only thing additional I will mention is that this will hide all sectionHeader content, regardless if there is history or not, which I don't believe we want.

hlomzik: just for the context this header displays id of current annotations, not history item

bmartel: That is correct, that is why hiding this ubiquitously seems incorrect.

ricardoantoniocm: Thank you all for the feedback. The latest code should address the issues. In my testing the Annotation header is visually hidden when the empty state is shown, but visible when it's not. @yyassi-heartex @bmartel @hlomzik

ricardoantoniocm: Ideally this annotation ID should exist in a different place that makes more sense on the application, so that it can be picked up by selenium tests directly and opening the history panel is avoided (this to me seems like a workaround that unfortunately adds unnecessary steps/delay to the tests).

bmartel: Oh I understand now, this change was to the **Empty State** version of this panel, not the general space. Ok, what you had originally would have been correct and passed selenium tests. Sorry I was confused about the change. 🙏

bmartel (approved):

web/libs/editor/src/components/CurrentEntity/AnnotationHistory.tsx
ricardoantoniocm: No worries!

yyassi-heartex (approved):

PR #7782, opened 2025-06-17
feat: BROS-86: Task Summary follow-ups

hlomzik:
Improve existing labeling types:
- move choice chips to a separate component and reuse it in `Choices`
- support `TextArea` with multiple texts
- fix per_region `Number`

Support new types of labeling:
- `DateTime`, just displaying the value
- `Taxonomy`, using the same chip component, aliases are not respected yet
- `Ranker`, display buckets and indices in their order, so it's easy to compare

Support data types in a simple shape:
- `Image`, `Audio`, `Video` as simple html tags
- `List`, `Paragraphs`, `Table` and `TimeSeries` as simple JSONs
- `Text`, `HyperText` as a text content
- `Pdf` as URL

Remove access to unavailable setting. If current user should not see agreement, it won't be available anyway.

Fix types a little.


web/libs/editor/src/components/TaskSummary/LabelingSummary.tsx
```diff
@@ -71,7 +71,7 @@ export const LabelingSummary = ({ annotations: all, controls, onSelect }: Props)
         const annotation = row.original;
 
         return (
-          <div className="flex gap-tight items-center cursor-pointer" onClick={() => onSelect(annotation)}>
+          <button type="button" className="flex gap-tight items-center cursor-pointer" onClick={() => onSelect(annotation)}>
```
nick-skriabin: ❤️

nick-skriabin (approved):

hlomzik:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15813307829)

PR #7790, opened 2025-06-18
fix: BROS-128: Correct multi-channel rendering and dynamic scaling

cloudmark:
This PR addresses and fixes several critical bugs within the `<TimeSeries>` component, enabling the correct rendering and dynamic scaling of multi-channel charts, particularly with large datasets. The changes ensure that the component is robust, performant, and behaves as expected when handling complex time series visualizations.

This solves issue [#7789](https://github.com/HumanSignal/label-studio/issues/7789).

### Reason for change

The `<TimeSeries>` component was failing to render data correctly when using a `<MultiChannel>` configuration or the `multiaxis="true"` attribute. The primary symptoms were a blank main chart view despite the overview scrubber being populated, and a Y-axis that would not scale dynamically on zoom.

This was caused by three distinct bugs:
1.  **Rendering Failure:** A typo (`this.line` instead of `channel.line`) in `TimeSeriesVisualizer.jsx` prevented the SVG path for channels from ever being drawn.
2.  **Dynamic Scaling Failure:** The Y-axis scaling logic for large datasets was flawed, using a CSS transform that did not update correctly. The component now forces a proper redraw of the path to ensure the Y-axis scales with the visible data.
3.  **Incorrect Component Invocation:** The `Channel.jsx` component was not correctly activating the `TimeSeriesVisualizer` for multi-channel views, leading to multiple, conflicting chart instances being rendered. The logic is now corrected to render a single, unified visualizer.

### Video

**Before:**
The main chart area is blank and does not render the channel data. The Y-axis is not scaled correctly.

https://www.loom.com/share/50fc240f1bae4d72ad421bdb9c2c8dfa

**After:**
All channels are rendered correctly in a single, unified chart. The Y-axis now scales dynamically and accurately as the user zooms and pans through the data.

https://www.loom.com/share/7d50d09befce4d19ad3da847ca2d6be7

### Rollout strategy

These changes are bug fixes to core functionality and can be rolled out directly. There is no feature flag associated with this fix.

### Testing

Verification was performed by:
1.  Creating a new project with the labeling configuration and task data specified in issue #7789.
2.  Observing the initial broken behavior (no channels rendered).
3.  Applying the code changes and confirming that all channels now render correctly.
4.  Zooming and panning on the timeline to verify that the Y-axis scales dynamically and smoothly as expected.

### Risks

[...]

Issue #7789, opened 2025-06-18
<TimeSeries> Doesn’t Render MultiChannel Data; Y-Axis Scaling Ignored on Zoom


cloudmark:
### **Describe the bug**

The `<TimeSeries>` component fails to correctly render data when multiple channels are nested within a `<MultiChannel>` tag. This issue is observable when using a large dataset loaded via a URL. While the overview scrubber at the bottom of the component renders the data preview correctly, the main chart area for the channels remains blank or shows nonsensical axes, indicating a failure in the main rendering logic.

Additionally, a related bug prevents the Y-axis from dynamically scaling on zoom when `fixedscale="false"` is set.

The root cause is a combination of flaws: the `<MultiChannel>` rendering path is broken, and the underlying `TimeSeriesVisualizer.jsx` component contains bugs related to path rendering and dynamic scaling.

### **To Reproduce**

1.  Create a new project in Label Studio.
2.  In the project settings, go to "Labeling Interface" and replace the default template with the following code:

    ```xml
    <View>
      <Labels name="label" toName="audio_sample" zoom="true" hotkey="ctrl+enter" showInline="true" choice="single">
        <Label value="category_a" background="#ff0d00" hotkey="1"/>
        <Label value="category_b" background="#00ff1e" hotkey="2"/>
      </Labels>

      <Audio name="audio_sample" value="$file"
             hotkey="space" sync="group_a"
             spectrogram="true"
             height="150"
             />

      <TimeSeries name="gyro_timeseries"
                  value="$gyro_data"
                  valueType="url"
                  timeColumn="relative_timestamp"
                  sync="group_a"
                  fixedScale="false"
                  overviewWidth="10%"
                  >
        <MultiChannel>
          <Channel column="gyro_x" strokecolor="#0047AB" legend="Gyro X" height="100"/>
          <Channel column="gyro_y" strokecolor="#FFBF00" legend="Gyro Y"  height="100"/>
          <Channel column="gyro_z" strokecolor="#DC143C" legend="Gyro Z" height="100"/>
          <Channel column="gyro_magnitude" strokecolor="#00FF7F" legend="Gyro Magnitude"/>
        </MultiChannel>
      </TimeSeries>
    </View>
    ```
3.  Go to the "Data Manager" tab and import a new task using the following JSON:

    ```json
    {
      "data": {
        "file": "https://storage.googleapis.com/labelstudio-uploads-public/recording_1749719922.m4a",
[...]

PR #7800, opened 2025-06-20
fix: BROS-128: Fix large data multichannel displaying

Gondragos:
This pull request fixes the handling of a large amount of data by `TimeSeriesVisualizer`.

Previously the same logic existed in the `ChannelD3`. And it was transferred to the new component with some changes for providing supporting MultiChannel mode of displaying. However, there were inaccuracies during the transfer and here are the fixes.

### Before
image
 
image

### After
image

image

### DIFF
--- a/web/libs/editor/src/components/TimeSeries/TimeSeriesVisualizer.jsx
+++ b/web/libs/editor/src/components/TimeSeries/TimeSeriesVisualizer.jsx
@@ -696,12 +696,16 @@ class TimeSeriesVisualizerD3 extends React.Component {
 
     let { series } = this.props;
 
+    if (this.optimizedSeries) {
+      channel.useOptimizedData = this.useOptimizedData;
+      series = this.optimizedSeries;
+    }
+
     series = series.filter((x) => {
       return x[column] !== null;
     });
 
     if (this.optimizedSeries) {
-      channel.useOptimizedData = this.useOptimizedData;
       channel.optimizedSeries = series;
     }
 
@@ -731,6 +735,7 @@ class TimeSeriesVisualizerD3 extends React.Component {
       .domain(d3.extent(values))
       .range([height - margin.max, margin.min]);
 
+    channel.x = this.x.copy();
     channel.plotX = this.x.copy();
     channel.y = y;
 
@@ -743,7 +748,7 @@ class TimeSeriesVisualizerD3 extends React.Component {
       .line()
       .defined((d) => d[time] >= range[0] && d[time] <= range[1])
       .y((d) => channel.y(d[column]))
-      .x((d) => this.x(d[time]));
+      .x((d) => channel.x(d[time]));
 
     const marker = this.defs
       .append("marker")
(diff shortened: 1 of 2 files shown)


bmartel (approved):

cloudmark:
Hey @Gondragos tested your changes and this fixes the timeline issue I identified as well.  

with this we can also close off this 

https://github.com/HumanSignal/label-studio/issues/7789

and you can close my PR here (because your fix is better because it keeps your optimisations)

https://github.com/HumanSignal/label-studio/pull/7790

Great stuff 🥳

cloudmark:
Hey @Gondragos ! I found an issue with the y-axis scaling when `fixedscale="false"`. The y-axis wasn't properly adjusting to the visible data range when zooming - it would show the full dataset range (e.g., -3.0 to 3.0) instead of dynamically scaling to the zoomed section, making the data appear compressed.

**Root cause:** `renderYAxis()` was being called before `setRangeWithScaling()`, so the axis was rendered with the initial domain instead of the calculated domain for the visible range.

**Fixed by:**
- Reordering the rendering calls so y-axis renders after domain calculations
- Improving the dynamic scaling logic with better bounds checking and null value handling
- Removing redundant axis rendering calls

The y-axis now properly scales to show tight ranges around visible data when zooming with `fixedscale="false"`.

You can cherry pick my commit 

https://github.com/HumanSignal/label-studio/compare/fb-BROS-128/ts-mc-uod...cloudmark:label-studio:fb-BROS-128/ts-mc-uod

or I can open a PR which goes to your branch

hlomzik (approved):

nick-skriabin (approved):

Gondragos:
Hey @cloudmark ! I'm glad to hear this PR helps with issues.

Unfortunately, I can’t reproduce the mentioned issue with `fixedscale="false"` on typical data. Please let me now if I need to use some specific data or the config instead.

Also, the root cause doesn't seem relevant. The diff you’ve provided as a fix contains removing 
```
    this.renderXAxis();
    this.renderYAxis();
```
 https://github.com/HumanSignal/label-studio/compare/fb-BROS-128/ts-mc-uod...cloudmark:label-studio:fb-BROS-128/ts-mc-uod#diff-59dc9fbf72d9df7b6716e0232d9891c4ec07fa4d04ccb3b5e36be9715d8727dfL892-L893
At the end of `setChannelRangeWithScaling` and adding it to the end of `setRangeWithScaling` right after calling `setChannelRangeWithScaling` 
https://github.com/HumanSignal/label-studio/compare/fb-BROS-128/ts-mc-uod...cloudmark:label-studio:fb-BROS-128/ts-mc-uod#diff-59dc9fbf72d9df7b6716e0232d9891c4ec07fa4d04ccb3b5e36be9715d8727dfR804-R805
Technically, it’s the same place from the perspective of the sequence of calls and it doesn't change the order. So it couldn’t cause the described issue.

Anyhow, it looks like these changes are out of the scope of the issue the PR is aimed at. So if there still is a bug, we can solve it in a separated PR just to avoid blocking this one.

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15821902060)
> Successfully merged:  14 files changed, 217 insertions(+), 37 deletions(-)

cloudmark:
Hey @Gondragos, I’ve opened issue #7815 to track the fixedScale="false" clipping bug—complete with the exact config and data files needed to reproduce it. The fix I proposed in my branch specifically addresses this issue.

PR #7802, opened 2025-06-20
fix: BROS-95: Incorrect draft is loaded during task history navigation in label stream

makseq:
This PR adds additional parameter `interaction=labelstream` in `api/tasks` when it's called from Label Stream.


hlomzik (approved):

makseq:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15813549101)

PR #7816, opened 2025-06-23
fix: FIT-255: Video settings on Labeling Interface Settings modal don't work

yyassi-heartex:
This pull request introduces two changes to the `SettingsModel` in `web/libs/editor/src/stores/SettingsStore.js`. These changes enhance the model's functionality by adding a new property and a method for toggling boolean properties.
before:
image

after:
image

Enhancements to `SettingsModel`:

* Added a new optional boolean property, `videoDrawOutside`, to the `SettingsModel`. This property can be used to control whether video elements are drawn outside a specific boundary.
* Introduced a new method, `toggleProperty(name)`, which allows toggling the value of a boolean property dynamically. This simplifies the process of flipping boolean states in the model.

### DIFF
--- a/web/libs/editor/src/stores/SettingsStore.js
+++ b/web/libs/editor/src/stores/SettingsStore.js
@@ -70,6 +70,8 @@ const SettingsModel = types
     videoHopSize: types.optional(types.number, 10),
 
     isDestroying: types.optional(types.boolean, false),
+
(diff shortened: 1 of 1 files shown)


hlomzik (approved):
we used a method we don't have in the code? wow!
that's very simple and important fix

PR #7821, opened 2025-06-24
fix: Correctly determine holdsState in onlyTime mode to prevent result removal

### LINKED ISSUE #7628: Datetime Tag with `only="time"` Parameter Fails to Display Selected Value
#### **Description**
When using the `datetime` tag  and setting the `only="time"` param, the selected time value fails to display correctly. Symptoms include:
- This value is updated only at the first time of the two selections/inputs, the subsequent changes will make the result empty;
- The exported dataset also lacks this field's value.

In addition, the following error appears in the browser console whenever the value is not properly displayed:
```
Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree.
(Object type: 'ClassificationArea', Path upon death: '/annotationStore/annotations/0/trackedState/areas/5hXc7ZWBTH', Subpath: '', Action: '/annotationStore/annotations/0/trackedState/areas/5hXc7ZWBTH.updateAppearenceFromState()').
Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

#### **Steps to Reproduce**
1. Add two `datetime` tags to config;
2. Add `only="time"` param;
3. Label a task and attempt to select or manually enter a time;
4. Observe whether the data exports correctly;
5. Open the browser console and check for related errors.

#### **Expected Behavior**
- Each change should trigger an update in both UI and model state;
- The exported result set needs to have correct datetime data
- No MobX-State-Tree warnings should appear in the console.

#### **Actual Behavior Screenshot**

Image
Image

xxxinsxxx:
### 🧠 Problem Summary 

In `only="time"` mode, the original logic of the `holdsState()` method was:

```js
if (self.onlyTime && !isDefined(self.time)) return false;
return isDefined(self.month) || isDefined(self.year);
```

During the **first time selecting a time**:
- `self.time` is assigned;
- `updateResult()` triggers `createPerObjectResult()`, successfully creating the result.

However, during the **second time modifying the time**:
- `self.time` is assigned;
- `updateResult()` triggers `self.result.area.setValue(self);`
- It enters setValue(tag)；
- The `tag.holdsState` is evaluated to determine whether to keep the result；
- Since `!isDefined(self.time)` is `false`, it falls through to `return isDefined(self.month) || isDefined(self.year)`；
- In `only="time"` mode, `month` and `year` are always `undefined`, so `holdsState` returns `false`；
- As a result, setValue(tag) calls `removeResult(result)`, **removing the result**, causing the bug where the input disappears after being set.

Changing the value again after that is just repeating this process.This behavior leads to an incorrect state evaluation that breaks expected functionality.

###### The involved files include: DateTime.jsx, ClassificationBase.js, AreaMixin.js

---

### Description:

In `only="time"` mode, the previous logic of `holdsState()` caused a regression where time updates would be discarded unexpectedly. The original condition:

```js
if (self.onlyTime && !isDefined(self.time)) return false;
return isDefined(self.month) || isDefined(self.year);
```

would fall through to checking `month` and `year`, which are always undefined in `only="time"` mode, leading to an incorrect `false` return value. This resulted in the result being removed during subsequent time changes.

This change simplifies and corrects the logic:

```js
if (self.onlyTime) return isDefined(self.time);
return isDefined(self.month) || isDefined(self.year);
```

Now, in `only="time"` mode, the state is solely determined by whether `self.time` is defined — aligning with the intended behavior and preventing unintended result removal.

---
### 🔍Problem Solved:
- fix #7628 

#### before:
before

#### after:
after

---

### 📁 Changed Files

**Path:** `libs/editor/src/tags/control/DateTime.jsx`

#### Code Changes:

```diff
 get holdsState() {
-  if (self.onlyTime && !isDefined(self.time)) return false;
-  return isDefined(self.month) || isDefined(self.year);
+  if (self.onlyTime) return isDefined(self.time);
+  return isDefined(self.month) || isDefined(self.year);
 },
```

---

### ✅ Testing Recommendation:

- Test time input on odd/even updates.
- Ensure results persist in `only="time"` mode.
- Validate behavior in other modes.
- Verify mixed-mode behavior.

### DIFF
--- a/web/libs/editor/src/tags/control/DateTime.jsx
+++ b/web/libs/editor/src/tags/control/DateTime.jsx
@@ -76,7 +76,7 @@ const Model = types
     },
 
     get holdsState() {
-      if (self.onlyTime && !isDefined(self.time)) return false;
+      if (self.onlyTime) return isDefined(self.time);
       return isDefined(self.month) || isDefined(self.year);
     },
 


robot-ci-heartex:

nick-skriabin (approved):

robot-ci-heartex:

robot-ci-heartex:

PR #7831, opened 2025-06-24
fix: FIT-264: Video settings hop size not working

yyassi-heartex:
This pull request makes a minor adjustment to the `TimelineComponent` in the `Timeline.tsx` file. The change involves renaming the `hopSize` property to `altHopSize` and ensuring it is properly assigned and passed to the component.
[screen-recorder-tue-jun-24-2025-11-34-38.webm](https://github.com/user-attachments/assets/e2b6fe48-815b-41ce-bee8-0919b2e104d9)

Key changes:

* [`web/libs/editor/src/components/Timeline/Timeline.tsx`](diffhunk://#diff-d4bff87ced2072a8362384fb79468b4f92dffc3aa0d908ae0b4a335d96ac8279L21-R22): Renamed `hopSize` to `altHopSize` and updated its assignment to maintain functionality.
* [`web/libs/editor/src/components/Timeline/Timeline.tsx`](diffhunk://#diff-d4bff87ced2072a8362384fb79468b4f92dffc3aa0d908ae0b4a335d96ac8279L131-R132): Updated the component to use the renamed `altHopSize` property instead of the previous `props.altHopSize`.


web/libs/editor/src/components/Timeline/Timeline.tsx
```diff
@@ -18,7 +18,8 @@ const TimelineComponent: FC<TimelineProps> = ({
   length = 1024,
   position = 1,
   framerate = 24,
-  hopSize = 1,
+  altHopSize = 1,
+  hopSize = altHopSize,
```
hlomzik, line 22: the only place simple `hopSize` is used is AudioNext:100 and we should not use it anymore. what do you think about unifying all these params to only one?

yyassi-heartex: ahh good to know - but that feels like it should be done as part of a separate ticket to remove AudioNext altogether - we've had multiple Audio versions for too long and its time to unify

nick-skriabin (approved):

bmartel (approved):

Issue #7815, opened 2025-06-23
Multi-channel with fixedScale="false" is not scaling properly


cloudmark:
**Describe the bug**
When `fixedScale="false"` is set on a `<TimeSeries>` display, the y-axis does **not** automatically rescale to fit the data. As shown in the video, the waveform is clipped at the top, and when zooming in further the signal remains chopped off instead of expanding to fill the view.

[[Watch the behavior on Loom](https://www.loom.com/share/70897de81e0e41bcb51e44b6563e6b3e)](https://www.loom.com/share/70897de81e0e41bcb51e44b6563e6b3e)

---

**To Reproduce**

1. Import this task JSON into Label Studio:

   ```json
   {
     "id": 8,
     "data": {
       "file": "https://storage.googleapis.com/labelstudio-uploads-public/recording_1749719922.m4a",
       "gps_data": "https://storage.googleapis.com/labelstudio-uploads-public/gps_1749719922.json",
       "gyro_data": "https://storage.googleapis.com/labelstudio-uploads-public/gyro_1749719922.json",
       "accel_data": "https://storage.googleapis.com/labelstudio-uploads-public/accel_1749719922.json",
       "magneto_data": "https://storage.googleapis.com/labelstudio-uploads-public/magneto_1749719922.json",
       "attitude_data": "https://storage.googleapis.com/labelstudio-uploads-public/attitude_1749719922.json"
     },
     "annotations": [],
     "predictions": []
   }
   ```
2. Use the following XML configuration in your Label Studio interface:

   ```xml
   <View>
     <Labels name="label" toName="audio_sample" zoom="true" hotkey="ctrl+enter" showInline="true" choice="single">
       <Label value="category_a" background="#ff0d00" hotkey="1"/>
       <Label value="category_b" background="#00ff1e" hotkey="2"/>
     </Labels>

     <Audio name="audio_sample" value="$file"
            hotkey="space" sync="group_a"
            spectrogram="true"
            height="150"/>

     <TimeSeries name="gyro_timeseries"
                 value="$gyro_data"
                 valueType="url"
                 timeColumn="relative_timestamp"
                 sync="group_a"
                 fixedScale="false"
                 overviewWidth="10%">
       <MultiChannel>
         <Channel column="gyro_x" strokecolor="#0047AB" legend="Gyro X" height="100"/>
         <Channel column="gyro_y" strokecolor="#FFBF00" legend="Gyro Y" height="100"/>
[...]

PR #7833, opened 2025-06-25
fix: BROS-147: Fix TimeSeries chart vertical shift

Gondragos:
A problem causing vertical displacement in TimeSeries charts has been fixed.
Large datasets and specific zoom levels caused an issue.  This sometimes shifted the chart beyond the visible area.

### Before

ts_v_shift_before

### After

ts_v_shift_after

## Related issues
https://github.com/HumanSignal/label-studio/issues/7815

### DIFF
--- a/.cursor/rules/cypress_tests.mdc
+++ b/.cursor/rules/cypress_tests.mdc
@@ -1,0 +1,739 @@
+---
+description: Writing and updating cypress integration tests
+globs: 
+alwaysApply: false
+---
+## Cypress Test Generation Rules for Label Studio
+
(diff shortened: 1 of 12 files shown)


yyassi-heartex (approved):

hlomzik (approved):
I would add a comment to the last part of your changes, because I thought that all previous changes already fixed the problem.
And tests would be good, even a simple test checking that there are colored pixels (not empty plot) for specific data and zoom.
But the fix is great and very precise!

hlomzik (approved):

PR #7834, opened 2025-06-25
fix: TRIAG-1495: Ensure fonts are being preloaded.

luarmr:
### Improve font preload effectiveness by adding `as="font"` and correct `type`

This PR updates the `<link rel="preload">` tags for `Figtree-Regular` and `Figtree-SemiBold` fonts to include:

- `as="font"` – correctly informs the browser of the resource type, enabling prioritized loading and proper caching behavior.
- `type="font/ttf"` – explicitly defines the MIME type of the font, which helps the browser understand and handle the resource correctly.

####   Why this matters
`as="font"` is required for the browser to correctly prioritize and preload fonts. Without it, preload may be ignored or downgraded.

`type="font/ttf"` ensures proper handling of the resource and may improve caching behavior.

`crossorigin="anonymous"` is retained to avoid CORS issues when fonts are reused from cache across pages or domains.

### Font path consistency: why we preload the un-hashed `.ttf`

In this setup, we are explicitly preloading the un-hashed font file:

```html
<link rel="preload" href="{{ settings.HOSTNAME }}/static/fonts/Figtree-Regular.ttf" ...>
```

This matches the font path currently referenced in the React app's CSS:

```css
@font-face {
  src: url("/static/fonts/Figtree-Regular.ttf");
}
```

Using the same path in both preload and runtime ensures the browser can reuse the preloaded resource and avoid duplicate requests.

### Alternative approach (for the future)
If we decide to rely on fingerprinted asset paths (e.g. Figtree-Regular.[hash].ttf), we would need to:

Inject the resolved font path into the HTML via Django (using {% static %}).

Dynamically load the font in JavaScript using the FontFace API and the injected path.

Ensure both the preload link and runtime font loading use the same hashed path.

This setup provides stronger cache busting guarantees and can be considered as a future enhancement.

```js
const font = new FontFace('Figtree', `url(${window.FIGTREE_FONT_URL})`, {
  weight: '400',
  style: 'normal',
  display: 'swap',
});
font.load().then(font => document.fonts.add(font));
```

####   References
[MDN: <link rel="preload">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-as)

####   How to test
 - Open Chrome DevTools → Network tab
 - Filter by .ttf
 - Reload the page
 - Confirm the font requests appear early and show Initiator: preload and Type: font
 
 #### Future
Considering using woff fonts and use FontFace API as mentioned in before.


hlomzik (approved):

nass600 (approved):

luarmr:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15880031976)
> Successfully merged:  3 files changed, 49 insertions(+), 2 deletions(-)

bmartel (approved):

PR #7837, opened 2025-06-25
fix: UTC-126: Data migration for FilterGroups associated with more than one view

mcanu:
After project duplication, the new View objects of the new project, still pointed to the old FilterGroup. Meaning that a change in filters in the new/old project affected the other one.
The project duplication bug was fixed but existing projects still have this data inconsistency.
This PR adds a data migration that copies the FilterGroup and Filters for each View.


hakan458 (approved):

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15903517685)

PR #7838, opened 2025-06-25
fix: BROS-146: Bitmask weird behavior leak

nick-skriabin:
This pull request introduces enhancements to the zoom and pan functionality in the image editor, focusing on improved detection of input methods (trackpad pinch vs. mouse wheel) and providing smoother zoom behavior. Additionally, it refines bitmask region handling to prevent unintended actions during drawing operations. Below are the most important changes grouped by theme:

### Zoom and Pan Enhancements:
* Added a `detectTrackpadPinch` method in `ImageView.jsx` to analyze zoom event patterns and distinguish between trackpad pinch gestures and mouse wheel input. This method uses event history and characteristics like frequency and delta values for detection.
* Updated the `handleZoom` method in `ImageView.jsx` to support smooth exponential zoom for trackpad pinch gestures and discrete zoom steps for mouse wheel events. This ensures a more natural zoom experience depending on the input method.
* Modified the `handleZoom` method in `Image.js` to implement separate zoom behaviors for trackpad pinch (smooth exponential zoom) and mouse wheel (discrete zoom steps). Added logic to clamp zoom levels and handle zoom restrictions effectively.

### Bitmask Region Handling:
* Refined logic in `ImageView.jsx` to ensure no actions are performed when clicking outside a bitmask region while a tool is selected. This prevents unintended behavior during drawing operations.
* Added a check in `ImageView.jsx` to skip highlight updates if no bitmask regions are present, improving performance and avoiding unnecessary operations.


web/libs/editor/src/components/ImageView/ImageView.jsx
```diff
@@ -583,12 +587,14 @@ export default observer(
...
-        const currentTool = item.getToolsManager().findSelectedTool().toolName;
+        const hasSelected = item.selectedRegions.some((r) => r.type === "bitmaskregion");
+        const isBitmask = ["BitmaskTool", "BitmaskEraserTool"].includes(
+          item.getToolsManager().findSelectedTool().toolName,
+        );
 
         // We want to avoid weird behavior here with drawing while selecting another region
         // so we just do nothing when clicked outside AND we have a tool selected
-        if (selectedRegion && currentTool === "BitmaskTool") {
+        if (hasSelected && currentTool === "BitmaskTool") {
```
hlomzik: `currentTool` is removed already in favor of simple boolean `isBitmask`, so let's fix the condition
```suggestion
        if (hasSelected && isBitmask) {
```

hlomzik (approved):

PR #7845, opened 2025-06-26
feat: UTC-137: Create a Typography component for Spark

nass600:
Introduced a new Typography atom to our Design System.

> Also restored Theme switch in sbDocs and applied it only to the preview content.

Screenshot 2025-06-26 at 10 35 03
Screenshot 2025-06-26 at 10 35 19


ricardoantoniocm:
We should consider calling this component Text, for brevity. Typography while correct seems quite verbose.

Also, I see some of the classes as `font-medium` and `font-body` which don't include the property, which in these two examples are `weight` and `family`. Is there a way to make tailwind generate them as `font-weight-medium` and `font-family-body` respectively? Otherwise these clases don't have semantic meaning as is.

ricardoantoniocm:
Suggestion: Each example under Usage at the top of the Docs page in Storybook should be accompanied with it's visual representation so that developers can visually see how the parameters affect the component.

yyassi-heartex (approved):

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15920091554)
> Successfully merged:  create mode 100644 web/libs/frontend-test/src/helpers/utils/SVGTransformUtils.ts

nass600:
> We should consider calling this component Text, for brevity. Typography while correct seems quite verbose.
> 
> Also, I see some of the classes as `font-medium` and `font-body` which don't include the property, which in these two examples are `weight` and `family`. Is there a way to make tailwind generate them as `font-weight-medium` and `font-family-body` respectively? Otherwise these clases don't have semantic meaning as is.

@ricardoantoniocm Tailwind doesn't allow to change this core classes, the only way is using workarounds via plugins that will generate new custom classes rather than overriding so not worth it

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15927770842)
> Successfully merged:  1 file changed, 4 insertions(+)

PR #7852, opened 2025-06-26
fix: BROS-87: Introduce unified zoom behavior for trackpads and mice

nick-skriabin:
Just moving the functionality under a feature flag.

Reason: buggy behavior.

### DIFF
--- a/web/libs/editor/src/tags/object/Image/Image.js
+++ b/web/libs/editor/src/tags/object/Image/Image.js
@@ -940,14 +940,14 @@ const Model = types
       if (val) {
         // Calculate the zoom change using exponential formula
         const zoomChange = Math.exp(-val * ZOOM_INTENSITY);
-        
+
         // Limit the maximum zoom change per event to prevent aggressive zooming
         // This prevents users from accidentally zooming too far with a single wheel event
         const limitedZoomChange = Math.max(
-          1 - MAX_ZOOM_CHANGE_PER_EVENT, 
-          Math.min(1 + MAX_ZOOM_CHANGE_PER_EVENT, zoomChange)
+          1 - MAX_ZOOM_CHANGE_PER_EVENT,
+          Math.min(1 + MAX_ZOOM_CHANGE_PER_EVENT, zoomChange),
         );
-        
+
         // Apply the limited zoom change
         const newZoom = clamp(self.currentZoom * limitedZoomChange, MIN_ZOOM, MAX_ZOOM);
         const zoomScale = newZoom;
--- a/web/libs/editor/src/tags/object/Image/Image.js
+++ b/web/libs/editor/src/tags/object/Image/Image.js
@@ -930,26 +930,45 @@ const Model = types
       self.resetZoomPositionToCenter();
     },
 
+    getInertialZoom(val) {
+      // Invert the delta value so that:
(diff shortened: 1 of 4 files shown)


bmartel (approved):

nass600 (approved):

nick-skriabin:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15908346263)

web/libs/editor/src/tags/object/Image/Image.js
```diff
@@ -931,25 +932,25 @@ const Model = types
...
+        // Limit the maximum zoom change per event to prevent aggressive zooming
+        // This prevents users from accidentally zooming too far with a single wheel event
+        const limitedZoomChange = Math.max(
+          1 - MAX_ZOOM_CHANGE_PER_EVENT, 
+          Math.min(1 + MAX_ZOOM_CHANGE_PER_EVENT, zoomChange)
+        );
+        
+        // Apply the limited zoom change
+        const newZoom = clamp(self.currentZoom * limitedZoomChange, MIN_ZOOM, MAX_ZOOM);
+        const zoomScale = newZoom;
```
yyassi-heartex: is `newZoom` being used anywhere? if not, lets clean up the unneeded variable
```suggestion
        const zoomScale = clamp(self.currentZoom * limitedZoomChange, MIN_ZOOM, MAX_ZOOM);
```

yyassi-heartex (approved):
minor suggestion but other wise looks good

web/libs/editor/src/tags/object/Image/Image.js
nick-skriabin: i don't think so. makes sense to remove, ty!

web/libs/editor/src/tags/object/Image/Image.js
```diff
@@ -931,25 +932,24 @@ const Model = types
...
-    handleZoom(val, mouseRelativePos = { x: self.canvasSize.width / 2, y: self.canvasSize.height / 2 }, pinch = false) {
+    handleZoom(val, mouseRelativePos = { x: self.canvasSize.width / 2, y: self.canvasSize.height / 2 }) {
       if (val) {
-        let zoomScale = self.currentZoom;
+        // Calculate the zoom change using exponential formula
+        const zoomChange = Math.exp(-val * ZOOM_INTENSITY);
+
+        // Limit the maximum zoom change per event to prevent aggressive zooming
+        // This prevents users from accidentally zooming too far with a single wheel event
+        const limitedZoomChange = Math.max(
```
hlomzik: you need to use `clamp()` here because you don't limit minimum value

hlomzik: basically you just need to run it two times, with ±MAX_CHANGE and with MIN/MAX

hlomzik (changes requested):
one important change regarding the goal of the PR, and the rest is good

nick-skriabin:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15913085611)

hlomzik (approved):

nass600 (approved):

PR #7853, opened 2025-06-26
fix: PLT-807: Dynamic batch size for export

triklozoid:
Large tasks with heavy data were causing memory issues during export due to fixed batches. Now batch size adapts  automatically: smaller batches for projects with large task data, larger batches for lightweight data.

  - Dynamic batch sizing: Added Project.get_task_batch_size() method that calculates optimal batch size based on actual task data size
  - Database optimization: Created PostgreSQL index task_proj_octlen_idx on (project_id, octet_length(data::text) DESC) for efficient data size queries
  - Configuration: Added MAX_TASK_BATCH_SIZE (default 1000) and TASK_DATA_PER_BATCH (default 50MB) settings
  - Feature flag: Protected changes with fflag_fix_back_plt_807_batch_size_26062025_short


label_studio/data_export/mixins.py
```diff
@@ -197,14 +198,20 @@ def get_export_data(self, task_filter_options=None, annotation_filter_options=No
             self.counters = {'task_number': 0}
             all_tasks = self.project.tasks
             logger.debug('Tasks filtration')
-            task_ids = (
+            task_ids = list(
                 self._get_filtered_tasks(all_tasks, task_filter_options=task_filter_options)
                 .distinct()
                 .values_list('id', flat=True)
+                .iterator(chunk_size=1000)
             )
```
jombooth, line 206: doesn't calling `list()` on the result effectively negate the benefit of using `iterator()`, since we have to materialize all the ids at once?

label_studio/core/settings/base.py
```diff
@@ -550,6 +550,8 @@
 
 # per project settings
 BATCH_SIZE = 1000
+MAX_TASK_BATCH_SIZE = int(get_env('MAX_TASK_BATCH_SIZE', 1000))
+TASK_DATA_PER_BATCH = int(get_env('TASK_DATA_PER_BATCH', 50 * 1024 * 1024))  # 50 MB in bytes
```
jombooth: I think comments defining these fields would be helpful

label_studio/data_export/mixins.py
```diff
@@ -197,14 +198,20 @@ def get_export_data(self, task_filter_options=None, annotation_filter_options=No
...
             base_export_serializer_option = self._get_export_serializer_option(serialization_options)
             i = 0
-            BATCH_SIZE = 1000
+
+            if flag_set('fflag_fix_back_plt_807_batch_size_26062025_short', user='auto'):
```
jombooth: It would be safer to use `self.project.created_by`

label_studio/data_export/mixins.py
triklozoid: We can't return an iterator because the code below doesn't support it. 
A list with an iterator consumes less memory than a list without it. 
And less memory than just iterating over a queryset.

label_studio/core/settings/base.py
triklozoid: Added comments.

label_studio/data_export/mixins.py
triklozoid: Changed to even more safer  `self.project.organization.created_by` :-)

jombooth (approved):

label_studio/data_export/mixins.py
jombooth: ah yup, that's what I was meaning 👍

triklozoid:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/15937251734)
> Successfully merged:  delete mode 100644 web/libs/ui/src/typography/sub.tsx

PR #7873, opened 2025-06-30
fix: TRIAG-1505: resolve MST reference error for userExtended in annotationStoreHistory

luarmr:
## Problem
Race condition when users navigate away during async operations:
```
Error: [mobx-state-tree] Failed to resolve reference 'XXXX' to type 'UserExtended'
```

## Root Cause
1. User triggers async operation (comment loading)
2. User navigates away → store gets destroyed  
3. Async callback tries to modify destroyed store
4. MobX throws error on "dead" store

## Solution
Improved `CommentsSdk` with robust error handling
**Try catch... Paranoid but safe**

## Benefits
- ✅ Eliminates MobX errors from race conditions
- ✅ Better architecture - SDK is store-agnostic
- ✅ Graceful degradation - comments show even if user enrichment fails
- ✅ More robust and testable

## Testing
1. Open Data Manager with comments
2. Load comments, then quickly navigate away (is rare, so low you connectivity speed or add a wait)
3. Check console - should see graceful handling instead of errors


hlomzik:
I'd say that title still needs formatting, because it went through some kind of normalization. Could you shorten it please and make more readable?
Code approach is good I think

bmartel (approved):

Gondragos (approved):

luarmr:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16035689906)
> Successfully merged:  create mode 100644 label_studio/users/validators.py

PR #7912, opened 2025-07-04
fix: fit-255: Video settings on Labeling Interface Settings modal don't work

yyassi-heartex:


nick-skriabin (approved):

yyassi-heartex:
this is an empty commit now - we can safely close this PR

PR #7950, opened 2025-07-09
fix: BROS-175: Add more resilience & improve error handling on saving annotations

niklub:
### Problems:
• Annotations are not saved due to backend errors, causing time or data loss.
• Errors are not shown to users.
• Blocking error modals harm the user experience.

### Solution:

Make sure annotation updates are not lost during errors by using frontend-level persistence.

Show clear error messages with calls to action (try again, contact support).

Prevent clutter by avoiding error modals that duplicate information.


web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -594,6 +595,31 @@ export class LSFWrapper {
     }
   };
 
+  /** @private */
+  showOperationToast(status, successMessage, errorAction, result) {
+    if (status === 200 || status === 201) {
+      this.datamanager.invoke("toast", { message: successMessage, type: "info" });
+    } else if (status !== undefined) {
+      const requestId = result?.$meta?.headers?.get("x-ls-request-id");
+      const supportUrl = requestId ? "https://support.humansignal.com/hc/en-us/requests/new?tf_37934448633869=" + requestId : "https://support.humansignal.com/hc/en-us/requests/new";
```
nick-skriabin: can we move the link to a constant and add a little description of what it is? also, the tf_ parameter is unclear here, looks like some magic value. deserves a a description too

nick-skriabin (changes requested):

web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -642,19 +665,22 @@ export class LSFWrapper {
...
-      this.datamanager.invoke("toast", { message: "There was an error updating your Annotation", type: "error" });
+    this.showOperationToast(status, "Annotation updated successfully", "Annotation is not updated", result);
 
     this.datamanager.invoke("updateAnnotation", ls, annotation, result);
 
     if (exitStream) return this.exitStream();
 
+    if (status >= 400 && status < 600) {
+      return;
+    }
```
hlomzik: just the first half of condition is enough for http codes
```suggestion
    if (status >= 400) return;
```

web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -938,12 +965,17 @@ export class LSFWrapper {
     }
 
     this.setLoading(false);
+    if (result?.$meta?.status >= 400 && result?.$meta?.status < 600) {
```
hlomzik: same here
```suggestion
    if (result?.$meta?.status >= 400) {
```

web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -594,6 +595,31 @@ export class LSFWrapper {
...
+  /** @private */
+  showOperationToast(status, successMessage, errorAction, result) {
+    if (status === 200 || status === 201) {
+      this.datamanager.invoke("toast", { message: successMessage, type: "info" });
+    } else if (status !== undefined) {
+      const requestId = result?.$meta?.headers?.get("x-ls-request-id");
+      const supportUrl = requestId ? "https://support.humansignal.com/hc/en-us/requests/new?tf_37934448633869=" + requestId : "https://support.humansignal.com/hc/en-us/requests/new";
+
+      this.datamanager.invoke("toast", { 
+        message: React.createElement("span", null,
```
hlomzik: can you use usual JSX here?

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -271,7 +271,7 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
         );
       }
 
-      if ((userGenerate && sentUserGenerate) || (!userGenerate && store.hasInterface("update"))) {
+      else if ((userGenerate && sentUserGenerate) || (!userGenerate && store.hasInterface("update"))) {
```
bmartel: This looks like it could be important, just not 100% sure what it fixes though. Is this a fix for the button states and which buttons can be included together?

hlomzik: Very good concern!

That's the fix proposed by me to fix a very long standing potential issue when we have both Submit and Update buttons on a screen. The reason: conditions are overlapping. And in this PR this case is not just potential: if we get an error during submit, but the annotation is still on a screen, both buttons are there!

The code is pure legacy and some parts of these conditions are not even relevant anymore. Supposedly! The main issue is that we have a lot of possible edge cases, so it's really hard to be sure how to change the code.

But here I think it's pretty robust, because we basically have `if (A && B) X` + `if (A && C) Y` and we always have at least X or Y, we don't have evidences of otherwise, but we do have evidences of both X and Y together, so adding this `else` should fix it. The only question left here for me is does it work when you submit it again? @niklub

web/libs/datamanager/src/sdk/lsf-sdk.js
niklub: done

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16203027352)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/skeleton/skeleton.tsx

nick-skriabin (approved):

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16219611771)
> Successfully merged:  delete mode 100644 label_studio/core/utils/openapi_extensions.py

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16249648888)
> Successfully merged:  create mode 100644 web/libs/app-common/src/pages/AccountSettings/sections/Hotkeys/utils.ts

hlomzik (approved):
I'd only ask to check non-http errors, like when we don't have internet at all. I hope in the worst case it will just act as before.

PR #7952, opened 2025-07-09
fix: BROS-171: Fix grid behavior

nick-skriabin:
The problem:
The combination of infinite scroll and react window doesn't handle dynamic grid changes very well. Specifically, when changing the column number from 5 to 10 we end up in a state where there's not enough items to fill the entire screen, hence no scrollbars appear. This lead to a situation when user is unable to view entire dataset and only limited to the first page because they can't scroll to initiate data load.

Solution:
Force data loading in such situations using a sideffect + custom logic to extend what infinite scroll currently does. It's not relying purely on the scroll behavior and it existence. Rather it will track how many items physically are there to trigger load when needed.

This pull request introduces improvements to the `GridView` component in `web/libs/datamanager/src/components/MainView/GridView/GridView.jsx`, focusing on optimizing data loading behavior and handling dynamic column count changes. Key updates include more robust logic for triggering data loads based on visible content, aggressive initial loading strategies, and adjustments to grid dimensions for better user experience.

### Data Loading Enhancements:
* Added multiple checks to trigger `loadMore` when there are insufficient items to fill the visible area, including edge cases such as very few items compared to column count or gaps between content and screen bottom. [[1]](diffhunk://#diff-fce86e3d740a69c1ffa62a1c99bbd6fe08a3cff824e71915acccc9cde8a0cf9aR191-R222) [[2]](diffhunk://#diff-fce86e3d740a69c1ffa62a1c99bbd6fe08a3cff824e71915acccc9cde8a0cf9aR246-R324)
* Introduced a custom `loadMore` function to bypass `InfiniteLoader` when needed, ensuring more reliable loading behavior.

### Dynamic Column Count Handling:
* Implemented logic in `useEffect` to detect changes in column count and trigger data loads if the new configuration results in insufficient visible items.

### Grid Dimension Adjustments:
* Modified `rowCount` to use `loadedRows` instead of `totalRows`, ensuring grid dimensions are based on loaded data rather than total data, thus avoiding overly long scrollbars. [[1]](diffhunk://#diff-fce86e3d740a69c1ffa62a1c99bbd6fe08a3cff824e71915acccc9cde8a0cf9aL153-R155) [[2]](diffhunk://#diff-fce86e3d740a69c1ffa62a1c99bbd6fe08a3cff824e71915acccc9cde8a0cf9aL240-R347)

### Code Improvements:
* Added `useEffect` and `useRef` imports to support new functionality for managing column count changes and aggressive loading strategies.

### DIFF
--- a/web/libs/datamanager/src/components/MainView/GridView/GridView.jsx
+++ b/web/libs/datamanager/src/components/MainView/GridView/GridView.jsx
@@ -190,36 +190,25 @@ export const GridView = observer(({ data, view, loadMore, fields, onChange, hidd
 
         // Calculate how many items are visible in the current view
         const visibleItemsCount = (visibleRowStopIndex - visibleRowStartIndex + 1) * columnCount;
-        
+
         // If we're showing items near the end of our loaded data, trigger loading
         // Use a threshold of 2 rows worth of items to trigger loading
         const threshold = Math.max(columnCount * 2, 8); // At least 8 items or 2 rows
-        
+
         // Check if we need to load more items
-        const shouldLoadMore = 
-          visibleItemStopIndex >= data.length - threshold &&
-          view.dataStore.hasNextPage;
+        const shouldLoadMore = visibleItemStopIndex >= data.length - threshold && view.dataStore.hasNextPage;
 
         // Also check if we don't have enough items to fill the visible area
         const hasEnoughItemsForVisibleArea = visibleItemStopIndex < data.length;
         const needsMoreItemsForDisplay = !hasEnoughItemsForVisibleArea && view.dataStore.hasNextPage;
-        
+
         // More aggressive check: if we have fewer items than columns, always load more
         const hasInsufficientItems = data.length < columnCount && view.dataStore.hasNextPage;
-        
+
         // Special case: if we have very few items compared to columns, be extra aggressive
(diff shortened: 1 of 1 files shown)


web/libs/datamanager/src/components/MainView/GridView/GridView.jsx
```diff
@@ -215,6 +232,70 @@ export const GridView = observer(({ data, view, loadMore, fields, onChange, hidd
...
+        loadMore?.();
+      }
+
+      // Special case: if we have fewer items than the column count itself, definitely load more
+      // This handles the case where there aren't enough items to even fill one row
+      if (data.length < currentColumnCount && view.dataStore.hasNextPage) {
+        loadMore?.();
+      }
+    }
+  }, [columnCount, data.length, view.dataStore.hasNextPage, view.dataStore.loading, loadMore, finalRowHeight]);
```
bmartel, line 266:
```suggestion
  }, [columnCount, data.length, view.dataStore.hasNextPage, loadMore, finalRowHeight]);
```
Do we want to be reacting to loading, your mention of:

> Note: We don't check !view.dataStore.loading here because we want to trigger loading

Makes me wonder if this could runaway?

bmartel (approved):

PR #7897, opened 2025-07-02
fix: BROS-156: Always show Fix+Accept when we have draft

hlomzik:
Any unsubmitted changes made by reviewer should be submitted via Fix+Accept.

Previously we only checked undo stack. But if you have draft already, but have no changes made in current session, you'll get Accept button only. Changes will be lost if you click it.

Now we also check for existence of draft.

Screenshot 2025-07-02 at 14 44 11

PR #7953, opened 2025-07-09
fix: BROS-172: Actually send a result in edge case for Fix+Accept

hlomzik:
When draft was created in previous session we see Fix+Accept button now (thanks to #7897), but the actual result is not sent along with approval. This fix adds the same condition from UI but to submission logic.


bmartel (approved):

hlomzik:
/git merge

> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16177486276)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/skeleton/skeleton.tsx

PR #7957, opened 2025-07-10
fix: FIT-363: Move hotkey management buttons to top of page

ricardoantoniocm:
### Reason for change

The actions on the hotkeys management page were previously displayed below the description, which was inconsistent with other page layouts. This PR moves the 'Import', 'Export', and 'Reset to Defaults' buttons to the top right of the Hotkeys title, aligning them with the established pattern for actions in the application header. This change improves visual harmony and user experience by making key actions more prominent and accessible.

The PR also includes implementing a confirm dialog for the Reset to Defaults function, to ensure consistency with how we handle destructive actions across the application.

#### Before
image

**Reset Hotkeys Modal Before**
image

#### After
image

**Reset Hotkeys Modal**
image

### Testing

Manual testing was performed to verify the following:

* **Hotkeys Page Layout:** Confirmed that the 'Import', 'Export', and 'Reset to Defaults' buttons are correctly positioned at the top right of the 'Hotkeys' title.
* **Button Functionality:**
    * **Import:** Verified that clicking 'Import' opens the import dialog and hotkeys can be successfully imported.
    * **Export:** Verified that clicking 'Export' triggers the download of the hotkey configuration.
    * **Reset to Defaults:** Verified that clicking 'Reset to Defaults' brings up the confirmation modal, and that confirming the action successfully resets hotkeys to their default values. Also confirmed that clicking 'Cancel' dismisses the modal without resetting.
* **Modal Styling:** Verified that the "Cancel" and "OK" (or "Reset to Defaults") buttons in the confirmation modal use default sizing as intended, following the removal of the `size="small"` prop.
* **Navigation:** Ensured that navigating to and from the Hotkeys section within Account Settings does not cause any layout or functionality issues.

### Reviewer notes

* The hotkey management logic for import, export, and reset has been moved into the new `HotkeysHeaderButtons` component. This involves some duplication of logic from the original `HotkeysManager` component.
* Please pay attention to the conditional rendering of the `HotkeysHeaderButtons` component to ensure it only appears when the 'hotkeys' section is active.


ricardoantoniocm:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16221500587)

bmartel (changes requested):
Why is there a poetry lock file change here? There shouldn't be any backend dependency changes, lets remove this change from the PR.

bmartel (commented):
Overall looks good, I will mention that this changeset is far larger than the stated goal, and though I think it is a good change it makes for a very hard to review PR because of the additional refactors. Just a note for next time 🙂

bmartel (approved):
Reverted all the backend deps changes to develop, approving based on everything going green in CI.

ricardoantoniocm:
@bmartel I can't reply to the other comment but in short, I was trying to satisfy github copilot's seemingly never ending suggestions.

But noted! Moving the buttons to the top required far more than I anticipated, like moving some of the functionality around, which then had its own ramifications.

Not sure what happened with the lock, that was unintentional.

Thank you for reviewing!

PR #7966, opened 2025-07-11
fix: FIT-362: Batching deletes in delete_tasks_without_signals to reduce memory usage

wesleylima:
Calling `.delete()` on a django queryset can reduce in high memory usage because each row to be deleted is loaded into an object so django can enforce `on_delete=CASCADE` constraints of other objects that have the deleted object as a ForeingKey.

This is the case for `Tasks`, it is a ForeignKey in at least 4 models in LSE and LS. We want to preserve the native django `.delete()` functionality, so to reduce memory usage, we do the delete operation in batches.

### DIFF
--- a/label_studio/core/utils/db.py
+++ b/label_studio/core/utils/db.py
@@ -86,21 +86,34 @@ def batch_delete(queryset, batch_size=500):
         int: Total number of deleted objects
     """
     total_deleted = 0
-    # Use iterator() to avoid loading all objects into memory at once
+
+    # Create a database cursor that yields primary keys without loading all into memory
+    # The iterator position is maintained between calls to islice
+    # Example: if queryset has 1500 records and batch_size=500:
+    # - First iteration will get records 1-500
+    # - Second iteration will get records 501-1000
+    # - Third iteration will get records 1001-1500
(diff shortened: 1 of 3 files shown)


label_studio/core/utils/db.py
```diff
@@ -71,3 +72,28 @@ def batch_update_with_retry(queryset, batch_size=500, max_retries=3, **update_fi
...
+
+    Returns:
+        int: Total number of deleted objects
+    """
+    total_deleted = 0
+    # Use iterator() to avoid loading all objects into memory at once
+    pks_to_delete = queryset.values_list('pk', flat=True).iterator(chunk_size=batch_size)
+    
+    # Delete in batches
+    while batch := list(itertools.islice(pks_to_delete, batch_size)):
```
bmartel: Not familiar with this syntax, how does this work?

wesleylima: Expanded the syntax and added comments 👍

bmartel (approved):

wesleylima:
/fm merge
> Command | Description
> --- | ---
> /fm sync <Optional branch override> | Sync upstream prs and merge with pull request base

PR #7979, opened 2025-07-14
ci: BROS-151: Add test for new UI, "simple init" FF and some hotkeys

hlomzik:
New tests:
- Label Studio init with FFs:
  - new UI (DEV-3873)
  - "simple init" (LEAP-443)
- New UI panels with some region interactions:
  - create and check relation
  - check relations hotkey
  - check draft state in annotation history

Also block one of the components for old UI (SidePanels) to improve coverage.


web/libs/editor/LSF.init.md
```diff
@@ -12,7 +14,7 @@ Flag:
 
 Components:
 
-- modern (FF_1170 + FF_3873): `SideTabsPanels` + `SidePanels/TabPanels` + `OutlinerTree` + `BottomBar`
+- modern (FF_1170 + FF_3873): `SidePanels/TabPanels/SideTabsPanels` (sic!!) + `OutlinerTree` + `BottomBar`
```
nick-skriabin, line 17: sick? :)

web/libs/editor/tests/integration/e2e/core/info_panels.cy.ts
```diff
@@ -0,0 +1,82 @@
...
+    cy.get("[class$=control_type_score]").should("have.text", "0.89");
+
+    // Change regions grouping by openining dropdown; it will be closed automatically
+    cy.contains("Manual").click();
+    cy.contains("Group by Label").should("be.visible");
+    // waiting for dropdown to be fully visible
+    // @todo better options?
+    cy.wait(200);
+    cy.contains("Group by Label").click();
+    cy.get("[class$=lsf-outliner-item__title").contains("Word1").should("be.visible");
```
nick-skriabin, line 38: not as the part of this PR, but we need aria-labels for such things 🤔

nick-skriabin (approved):

web/libs/editor/LSF.init.md
hlomzik: that's a common latin note used to state that the original spelling and wording are correct (and confusing)

web/libs/editor/tests/integration/e2e/core/info_panels.cy.ts
hlomzik: yeah, would be great! we have some helpers, but definitely not enough! added this to a next work list, thank you!

PR #7982, opened 2025-07-15
fix: UTC-168: Conditional Nesting Issue - Text Area

nass600:
Fixed submitting annotations that should not due to its parent visibility

Before:

https://github.com/user-attachments/assets/8b0e6266-839c-447c-bd95-791d4b7fbdd6

After:

https://github.com/user-attachments/assets/1a8aa336-ffdb-45cf-ac81-30462be6fa4d


yyassi-heartex (approved):

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16293029716)
> Successfully merged:  19 files changed, 241 insertions(+), 335 deletions(-)

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16296107991)
> Successfully merged:  1 file changed, 17 deletions(-)

web/libs/editor/src/regions/Result.js
hlomzik, line 1: I see that TextArea just doesn't use `VisibilityMixin`. I think that might be a proper fix. Can you try it?
This mixin already checks for parent's visibility

hlomzik (changes requested):

web/libs/editor/src/regions/Result.js
hlomzik: Hmmm, but your solution would work for all classification controls. Also I was concerned about check for required controls, which uses visibility params. But I checked and it actually works well, so the only thing to fix is serialization. And your solution covers it well! Approving.

hlomzik (approved):

web/libs/editor/src/regions/Result.js
nass600: Very good point @hlomzik , thanks!

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16301221600)
> Successfully merged:  2 files changed, 3 insertions(+), 2 deletions(-)

PR #7983, opened 2025-07-15
feat: Support snap to pixel for RectangleLabels

niklub:
Added ability to snap bounding box annotation to the pixel grid via:
```xml
<RectangleLabels name="bbox" toName="image" snap="pixel"/>
```

https://www.loom.com/share/a362e1747c2543e487be344ca20f4827?sid=05df6648-222e-4f80-8ede-64adc71a195f


web/libs/editor/src/regions/RectRegion.jsx
```diff
@@ -492,7 +578,46 @@ const HtxRectangleView = ({ item, setShapeRef }) => {
...
+
+        item.setPositionInternal(
+          topLeftPoint.x,
+          topLeftPoint.y,
+          snappedWidth,
+          snappedHeight,
+          rotation,
+        );
+      } else {
+        item.setPosition(x, y, width, height, rotation);
```
hlomzik: `setPosition()` already has snapping logic, so this part can be left as it was before

web/libs/editor/src/regions/RectRegion.jsx
```diff
@@ -467,13 +521,45 @@ const HtxRectangleView = ({ item, setShapeRef }) => {
...
```
hlomzik: The same here — logic is already in `setPosition()`

hlomzik:
We should test regions restoration via:
- load submitted annotation
- switch between annotations
- undo/redo
- history switching
- switching between images for MIG

Code looks good! Would be also nice to have screenshots/videos in PR to see it in action.

nick-skriabin (approved):

web/libs/editor/src/regions/RectRegion.jsx
```diff
@@ -478,6 +498,14 @@ const HtxRectangleView = ({ item, setShapeRef }) => {
       t.setAttr("scaleX", 1);
       t.setAttr("scaleY", 1);
 
+      // Sync Konva shape attributes back to computed canvas coordinates to cause a re-render
+      t.position({
```
hlomzik: better to wrap this with `if (self.control?.snap === "pixel")` to avoid unnecessary re-renders + add a comment that these canvas values are updated inside `setPosition()`
+apply this to the same code in `onDragEnd`

hlomzik (changes requested):

hlomzik (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16422955594)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/video/timeline_region_loop.cy.ts

PR #8005, opened 2025-07-18
fix: BROS-182: api/tasks timeouts with large number of annotations

makseq:
TaskAPI for a specific task works slowly. We don't need some fields (annotations_results, predictions_results) when we call TaskAPI, because annotations and predictions are already included inside of it. However, these fields slow down a lot. In this PR, we exclude them.


triklozoid (approved):

makseq:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16383227164)
> Successfully merged:  create mode 100644 label_studio/projects/migrations/0030_project_search_vector_index.py

PR #8009, opened 2025-07-18
fix: FIT-407: buttonVariant font size cleanup

yyassi-heartex:
This pull request includes a minor update to the `buttonVariant` function in `web/libs/ui/src/lib/button/button.tsx`. The change removes the `text-label-medium` class from the button styles to simplify the applied text styles.

before:
Screenshot 2025-07-18 at 2 32 41 PM

after:
Screenshot 2025-07-18 at 2 30 55 PM

* [`web/libs/ui/src/lib/button/button.tsx`](diffhunk://#diff-259a95d990638123fd08358c173b4a3fb80fa24ec5971af905c9a551db1f665cL72-R72): Removed the `text-label-medium` class from the button's style definitions in the `buttonVariant` function. This simplifies the text styling by relying on other existing classes.


bmartel (approved):
This has global implications, I am fairly certain this is the most correct base state of a button, just want to be sure this doesn't have any weird effects in the application.

PR #8010, opened 2025-07-18
fix: PLT-840: Store last_activity in redis

triklozoid:


label_studio/users/functions/last_activity.py
```diff
@@ -0,0 +1,560 @@
...
+        logger.info('Scheduled user activity sync job')
+        return True
+
+    except Exception as e:
+        logger.error('Failed to schedule activity sync: %s', e)
+        return False
+
+
+@job('low', timeout=7200)
+def full_activity_sync() -> dict:
```
bmartel: This is a manually invoked task?

triklozoid:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16381260988)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/core/info_panels.cy.ts

label_studio/users/functions/last_activity.py
triklozoid: I used it for testing, I'm not going to merge it.

label_studio/users/models.py
```diff
@@ -65,8 +66,35 @@ class UserLastActivityMixin(models.Model):
...
+        # Try to set in Redis first
+        redis_success = set_user_last_activity(self.id, current_time)
+
+        if not redis_success:
+            # Fallback to direct database update if Redis fails
+            self.last_activity = current_time
+            self.save(update_fields=['last_activity'])
+        else:
+            # Schedule sync if threshold is reached
+            schedule_activity_sync()
```
wesleylima, line 81: This flow will leave a lot of activity "unsynced" very often, right? Any time a bust of activity produces less than 50 updates, it won't save to the database.
For example, if I'm annotating pretty regularly, then take a break at 4:30 PM, I would expect the "unsyced" last activity to be at least very close to 4:30 PM. But after the break, if I do an action at  6:31 PM that doesn't produce a lot of activity (like opening a settings page), then walk away, it's not likely that this activity would've pushed me over the SYNC_THRESHOLD, so the database would still have ~4:30PM as my last activity, even though it really should be 6:31 PM.

Is this correct? Or is this being handled somewhere else?

I think this is still fine for 99.99% cases since Redis isn't likely to lose it's data (We may even have persistence enabled?) but it may be good to note.

Maybe a chron to force a "flush" of last_activty every day just in case?

label_studio/users/functions/last_activity.py
```diff
@@ -0,0 +1,481 @@
...
+            # Still clean up the batch set
+            cleanup_redis_activity_data(user_ids)
+            return {'success': True, 'processed': 0, 'errors': len(user_ids), 'message': 'No activity data found'}
+
+        # Sync to database
+        sync_result = _bulk_update_user_activities(activities)
+
+        if sync_result['success']:
+            # Clean up Redis data only if database sync was successful
+            cleanup_redis_activity_data(user_ids)
```
wesleylima, line 378: There's a race condition of sorts that could cause some activity_data to be deleted, right? 
if there is activty_data added to redis after `_bulk_update_user_activities` is called, but before we run `cleanup_redis_activity_data`, that activity data will be discarded

label_studio/users/models.py
triklozoid: Counter is global over all users, so "unsynced" state is possible only if nobody makes request at all.

label_studio/users/functions/last_activity.py
triklozoid: That's correct, but in my opinion, `last_activity` is not require 100% accuracy, the result of this race condition will be: "in some rare cases, last_activity is outdated for a few seconds", I think it's acceptable level of risk, comparing to the general performance improvement, we got from this feature.

wesleylima: I agree the tradeoff is worth it too! 👍

label_studio/users/models.py
wesleylima: Oh that makes sense! Should almost never stay unsynced then 👍

wesleylima (approved):

triklozoid:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16580822419)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/cloud-provider-s3.svg

makseq (approved):

PR #8027, opened 2025-07-22
fix: default.conf to include server SNI

### LINKED ISSUE #8008: Issue Loading URL with Persistent Storage using AWS(MinIO)
**Describe the bug**
A file that was successfully uploaded using `Import Data` on an arbitrary project cannot be retrieved as it errors out with the following error statement: `There was an issue loading URL from $image value`. 

It seems like the URL path for the file is not being correctly generated to retrieve files in Persistent Storage

**To Reproduce**
1. Properly functioning MinIO instance 
2. Configure Bucket as outlined [here](https://labelstud.io/guide/persistent_storage#Set-up-Amazon-S3)
3. Configure environment variables to use Persistent Storage as outlined [here](https://labelstud.io/guide/persistent_storage#Set-up-Amazon-S3)
   ```
      STORAGE_TYPE: s3
      STORAGE_AWS_ACCESS_KEY_ID: "minio_access_key"
      STORAGE_AWS_SECRET_ACCESS_KEY: "minio_secret_key"
      STORAGE_AWS_BUCKET_NAME: "bucket_name_here"
      STORAGE_AWS_REGION_NAME: ""
      STORAGE_AWS_ENDPOINT_URL: "https://minio.company.com"
   ```
4. Setup/Run server
5. Login and on the homepage click `Create Project`
6. At `Data Import` click `Upload Files` and upload an audio file
   - _NOTE: this error occurs for ALL file types. an audio file was specified in order to experience the error quicker_
7. Click `Save`
8. Choose any template
   - _NOTE: the template does not matter in reproducing this error_
9. Click on the task then click the play button

**Expected behavior**
A file that gets successfully uploaded to a Persistent Storage backend can get retrieved successfully using the correct path.

**Screenshots**
1. Improperly generated URL causing the error

   Image

2. Properly uploaded audio file in MinIO
   
   Image

 - OS: OS used in Label Studio's official Docker Image for version 1.20.0
 - Label Studio Version: 1.20.0

**Additional context**
- Related Issue: https://github.com/HumanSignal/label-studio/issues/4317

Using the Cloud Storage functionality to add a Source Cloud Storage seems to retrieve the file properly and thus if I do not use direct uploads (`Import Data`), it's fine. 
However the flow of having to:
1. Upload files to MinIO
2. Go into a project's `Settings` -> `Cloud Storage` -> `Add Source Storage`
3. `Sync storage`

feels a lot more cumbersome and clunky than just:
1. On the project page, click `import` -> choose the files for upload -> `Import` 

I understand that it's recommended to use a Cloud Storage and sync with external storage instead of a direct upload due to possibly overloading the workers that run Label Studio / Nginx. However, for simple, non-demanding uploads, using the direct upload should be the easiest way to get a file into one's project.

shkarface:
### Reason for change

This PR fixes [issue #8008](https://github.com/HumanSignal/label-studio/issues/8008), where TLS Server Name Indication (SNI) was not properly enabled when proxying requests over HTTPS. As a result, some upstream servers relying on SNI for correct routing were returning incorrect or failed responses.

### Solution

Added the following directives to the NGINX configuration to explicitly enable SNI and set the server name:

```nginx
proxy_ssl_server_name on;
proxy_ssl_name $download_host;
```

### DIFF
--- a/deploy/default.conf
+++ b/deploy/default.conf
@@ -164,6 +164,9 @@ http {
         proxy_set_header Cookie '';
         proxy_hide_header Content-Disposition;
         add_header Content-Disposition $upstream_http_content_disposition;
+        # Enable TLS SNI (important!)
+        proxy_ssl_server_name on;
+        proxy_ssl_name $download_host;
         # Stops the local disk from being written to (just forwards data through)
         proxy_max_temp_file_size 0;
         # Proxy the remote file through to the client


farioas:
I am concerned that this change may also affect private S3 endpoints

shkarface:
> I am concerned that this change may also affect private S3 endpoints

This is tested on both private and public S3 endpoints, the issue is that NGINX by default resolves a domain once and uses the domain to make the requests instead.

So something like `https://s3.temp.com/upload/some-key?query-params-here=value` becomes `https://IP_ADDRESS/PATH` when nginx makes the outgoing call. here are logs from one of our services:

`
2025/07/22 09:05:52 [error] 23#23: *82725 peer closed connection in SSL handshake while SSL handshaking to upstream, client: 10.244.5.74, server: , request: \"GET /storage-data/uploaded/?filepath=upload/5/f6879e50-e1272951-WhatsApp_Image_2025-06-17_at_11.08.22.jpeg HTTP/1.1\", upstream: \"https://10.10.80.85:443/label-studio/label-studio/upload/5/f6879e50-e1272951-WhatsApp_Image_2025-06-17_at_11.08.22.jpeg?AWSAccessKeyId=HIDDEN&Signature=HIDDEN%3D&Expires=1753261552\", host: \"label-studio.dev.krd\", referrer: \"https://label-studio.HIDDEN.COM/projects/5/data?tab=7\""
`

This part `https://10.10.80.85:443/` is the IP address of the name that is resolved

PR #8033, opened 2025-07-23
fix: UTC-198: Brightness controls are not working in the task interface

nass600:
Restored brightness functionality when Konva is used (Bitmask flag on). Note that it already works when not using Konva

Result:

https://github.com/user-attachments/assets/7784a49f-ca22-4835-93fe-2a27cddf1fdb

### DIFF
--- a/web/libs/editor/src/components/ImageView/ImageView.jsx
+++ b/web/libs/editor/src/components/ImageView/ImageView.jsx
@@ -19,7 +19,7 @@ import { useObserver } from "mobx-react";
 import ResizeObserver from "../../utils/resize-observer";
 import { debounce } from "../../utils/debounce";
 import Constants from "../../core/Constants";
-import { fixRectToFit } from "../../utils/image";
+import { fixRectToFit, mapKonvaBrightness } from "../../utils/image";
 import {
   FF_DEV_1442,
   FF_DEV_3077,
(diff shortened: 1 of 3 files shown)


web/libs/editor/src/utils/image.js
```diff
@@ -350,3 +350,19 @@ export function getActualZoomingPosition(
...
+
+/**
+ * Map brightness percentage to a value between -1 and 1 (Konva brightness value).
+ *
+ * 100% brightness is 0, 400% brightness is 0.8.
+ *
+ * @param {number} brightnessPercent
+ * @returns {number}
+ */
+export function mapKonvaBrightness(brightnessPercent) {
```
yyassi-heartex, line 362: im glad this was extracted to a function - would it be possible to add some unit tests that would show the expected output with some common inputs? this'll help make it more clear what the function does. especially since there are a bunch of magic numbers (300, 0.5, 0.8) and i dont have any context on what they mean

web/libs/editor/src/components/ImageView/ImageView.jsx
```diff
@@ -1281,32 +1281,50 @@ const EntireStage = observer(
...
-  }, [imageEntity?.downloaded]);
+  }, [imageEntity?.downloaded, imageEntity?.currentSrc]);
 
   const { width, height } = useMemo(() => {
     return {
       width: imageEntity.naturalWidth,
       height: imageEntity.naturalHeight,
     };
-  }, [imageEntity.naturalWidth, imageEntity.naturalHeight, item.stageWidth, item.stageHeight]);
+  }, [imageEntity.naturalWidth, imageEntity.naturalHeight]);
```
yyassi-heartex: with the removal item.stageWidth and item.stageHeight dependencies - it would mean this memo would re-fire when that happens - is that expected?
would a window resize potentially trigger these values to change requiring that we update this memoized value?

web/libs/editor/src/components/ImageView/ImageView.jsx
```diff
@@ -1281,32 +1281,50 @@ const EntireStage = observer(
...
+  useEffect(() => {
+    if (imageEntity?.downloaded && imageEntity.currentSrc) {
       const img = new window.Image();
-      img.src = ent.currentSrc;
-      img.width = Number.parseInt(ent.naturalWidth);
-      img.height = Number.parseInt(ent.naturalHeight);
-      return img;
+      img.crossOrigin = "anonymous";
+      img.src = `${imageEntity.currentSrc}?cb=${Date.now()}`; // avoid tainted cache
+      img.width = Number(imageEntity.naturalWidth);
```
hlomzik: the original code is so confusing... `naturalWidth` and `naturalHeight` are coming straight from the image and always numbers, I don't know why did we have this transform. let's add a comment
```suggestion
      // @todo `naturalWidth` and `naturalHeight` should always be numbers; check and remove transform
      img.width = Number(imageEntity.naturalWidth);
```

web/libs/editor/src/components/ImageView/ImageView.jsx
```diff
@@ -1281,32 +1281,50 @@ const EntireStage = observer(
...
+  // Load image with proper CORS and load event
+  useEffect(() => {
+    if (imageEntity?.downloaded && imageEntity.currentSrc) {
       const img = new window.Image();
-      img.src = ent.currentSrc;
-      img.width = Number.parseInt(ent.naturalWidth);
-      img.height = Number.parseInt(ent.naturalHeight);
-      return img;
+      img.crossOrigin = "anonymous";
+      img.src = `${imageEntity.currentSrc}?cb=${Date.now()}`; // avoid tainted cache
```
hlomzik: that might lead to problems if url already has query params.
plus why don't we trust cache here? image will be redrawn by canvas methods anyway, no?

hlomzik (changes requested):

web/libs/editor/src/components/ImageView/ImageView.jsx
nass600: It is as you mention, we do not need this casting so I will remove it

web/libs/editor/src/components/ImageView/ImageView.jsx
nass600: we do not need that anymore (remnants from my first fix) as Konva requires the `cache()` method in order to apply a change for the filters

web/libs/editor/src/components/ImageView/ImageView.jsx
nass600: Good catch, thanks

web/libs/editor/src/utils/image.js
nass600: Adding them

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16496387074)
> Successfully merged:  3 files changed, 9 insertions(+), 42 deletions(-)

yyassi-heartex (approved):
approving assuming all concerns from other reviewers have been addressed

hlomzik (approved):

nass600:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16499962183)
> Successfully merged:  1 file changed, 1 insertion(+), 1 deletion(-)

PR #8038, opened 2025-07-24
fix: BROS-220: Fix resize handling in Video tag

Gondragos:
Refactored `Frames.tsx` and `Minimap.tsx` to use the newly added `useResizeObserver` hook for timely width calculations. Added memoized references and hooks to have relevant callbacks. Introduced a new debounced `useResizeObserver` hook in the core library.

### Before:

[video_resize](https://github.com/user-attachments/assets/3bb0f7f7-78a6-42ba-9b1b-531ad3033046)

[timeline_resize](https://github.com/user-attachments/assets/6eaec82c-5d58-4756-af2a-44abc2409b7c)

[playing_zoom](https://github.com/user-attachments/assets/0c30845f-89c9-4f03-80e6-daf192e2984a)

### DIFF
--- a/web/libs/core/src/hooks/useResizeObserver.ts
+++ b/web/libs/core/src/hooks/useResizeObserver.ts
@@ -1,0 +1,140 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+
+export interface ResizeObserverSize {
+  width?: number;
+  height?: number;
+}
+
+export interface UseResizeObserverOptions<T extends ResizeObserverSize = ResizeObserverSize> {
+  /**
+   * Function to extract size data from ResizeObserver entries
+   * Default: returns the size of the first element
+   */
+  extractSize?: (elements: Element[], entries: ResizeObserverEntry[]) => T;
+  /**
+   * Debounce time in milliseconds to prevent excessive updates
+   * Default: 16ms (roughly 60fps)
+   */
+  debounceMs?: number;
+  defaultSize?: T;
+}
+
+export type ElementsOption = Element | Element[];
+
+const DEFAULT_DEBOUNCE_MS = 16;
+
+const defaultExtractSizeOneElement = <T extends ResizeObserverSize = ResizeObserverSize>(
+  _elements: Element[],
+  entries: ResizeObserverEntry[],
+): T => {
+  if (entries.length === 0) return {} as T;
+  const { width, height } = entries[0].contentRect;
+  return { width, height } as T;
+};
+
+const defaultExtractSizeMultipleElements = <T extends ResizeObserverSize = ResizeObserverSize>(
+  elements: Element[],
+  entries: ResizeObserverEntry[],
+): T => {
+  if (entries.length === 0) return {} as T;
+  const { width, height } = elements[0].getBoundingClientRect();
+  return { width, height } as T;
+};
+
+/**
+ * Hook for observing resize changes on elements
+ * Prevents "ResizeObserver loop limit exceeded" errors through debouncing
+ *
+ * @param elements - Element or array of elements to observe
+ * @param options - Configuration options
+ * @returns Current size data that triggers re-renders
+ */
+export function useResizeObserver<T extends ResizeObserverSize = ResizeObserverSize>(
+  elements: ElementsOption,
+  options: UseResizeObserverOptions<T> = {},
+): T {
+  const elementArray = Array.isArray(elements) ? elements : [elements];
+  const {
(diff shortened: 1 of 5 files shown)


yyassi-heartex (approved):
About time we got a hook for resize observer! great work!

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16500029474)
> Successfully merged:  1 file changed, 1 insertion(+), 1 deletion(-)

web/libs/core/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @param elements - Element or array of elements to observe
+ * @param options - Configuration options
+ * @returns Current size data that triggers re-renders
+ */
+export function useResizeObserver<T extends ResizeObserverSize = ResizeObserverSize>(
+  elements: ElementsOption,
+  options: UseResizeObserverOptions<T> = {},
+): T {
+  const elementArray = Array.isArray(elements) ? elements : [elements];
```
hlomzik:
```suggestion
  const elementsArray = Array.isArray(elements) ? elements : [elements];
```

web/libs/core/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export type ElementsOption = Element | Element[];
+
+const DEFAULT_DEBOUNCE_MS = 16;
+
+const defaultExtractSizeOneElement = <T extends ResizeObserverSize = ResizeObserverSize>(
+  _elements: Element[],
+  entries: ResizeObserverEntry[],
+): T => {
+  if (entries.length === 0) return {} as T;
```
hlomzik: it's only called when length is 1, so this line is redundant
```suggestion
```

web/libs/core/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,140 @@
...
+  const { width, height } = entries[0].contentRect;
+  return { width, height } as T;
+};
+
+const defaultExtractSizeMultipleElements = <T extends ResizeObserverSize = ResizeObserverSize>(
+  elements: Element[],
+  entries: ResizeObserverEntry[],
+): T => {
+  if (entries.length === 0) return {} as T;
+  const { width, height } = elements[0].getBoundingClientRect();
```
hlomzik, line 39: why methods are different for 1 and for many? could you add a link to MDN or another resource?

web/libs/core/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,140 @@
...
+            const newSize = extractSize(elementsRef.current as Element[], entries);
+            setSize((prevSize) => {
+              if (prevSize === newSize) return prevSize;
+              if (JSON.stringify(prevSize) !== JSON.stringify(newSize)) {
+                return newSize;
+              }
+              return prevSize;
+            });
+          } catch (error) {
+            console.warn("Error in resize observer callback:", error);
```
hlomzik:
```suggestion
            console.warn("Error in resize observer callback: ", error);
```

web/libs/editor/src/components/VideoCanvas/VideoCanvas.tsx
```diff
@@ -264,11 +264,13 @@ export const VideoCanvas = memo(
...
 
       if (playing) {
-        raf.current = requestAnimationFrame(handleAnimationFrame);
+        raf.current = requestAnimationFrame(handleAnimationFrameRef.current);
       } else {
         cancelAnimationFrame(raf.current!);
       }
     };
+    const handleAnimationFrameRef = useRef(handleAnimationFrame);
+    handleAnimationFrameRef.current = handleAnimationFrame;
```
hlomzik, line 273: this looks so confusing... are there other ways to have recursion in functional components?

Gondragos: Recursion works fine. It just can't use the right values inside the function without moving them to ref if they are changing during execution.
It's the same trick as when we move handlers from props to ref to decrease the necessity to invalidate caches of useMemo/useCallback.

web/libs/core/src/hooks/useResizeObserver.ts
Gondragos: With one element we get in entries exactly the element I expect to get, and it already has the right sizes.
With more than one element it provides only entries that exactly had changes. So it's harder to get the first element size if there is no guarantee that the first entry will be related to the first element.

PR #8046, opened 2025-07-25
feat: FIT-449: Select can now support partial option rendering to reduce dom footprint

yyassi-heartex:
This pull request introduces enhancements to the `Select` and `FilterDropdown` components, focusing on adding support for infinite scrolling and improving performance when rendering large option lists. The changes include updates to component props, rendering logic, and the integration of the `react-window` library for virtualization.

[screen-recorder-fri-jul-25-2025-13-29-40.webm](https://github.com/user-attachments/assets/d7fc414b-090c-46e6-ac78-61ec7a247866)

### Enhancements to `Select` Component:

* **Infinite Scroll Support:**
  - Added a new `useInfiniteScroll` prop to the `Select` component, enabling the use of infinite scrolling for option rendering. [[1]](diffhunk://#diff-567459e612a12d17fb4237f0381f79f4148ad291939afbf005a1c02e398fe942R81) [[2]](diffhunk://#diff-c04973e3a3095798cf15be72420af0687bc8cc748028e26fd47b27d4c3c37d5dR75)
  - Integrated `VariableSizeList` from `react-window` to virtualize the rendering of options when `useInfiniteScroll` is enabled, improving performance for large datasets.

* **Refactored Option Rendering:**
  - Extracted the logic for rendering options into a `useMemo` hook for better performance and maintainability.

### Enhancements to `FilterDropdown` Component:

* **Infinite Scroll Integration:**
  - Enabled infinite scrolling in the `FilterDropdown` component by setting the `useInfiniteScroll` prop to `true`.

### Dependency Update:

* **`react-window` Integration:**
  - Imported `VariableSizeList` from the `react-window` library to support virtualization in the `Select` component.


nass600:
@yyassi-heartex what about the requests? If I render first page 1, later page 2 (replace the dom) and scroll back up again, would I make a new request to page 1?

yyassi-heartex:
> @yyassi-heartex what about the requests? If I render first page 1, later page 2 (replace the dom) and scroll back up again, would I make a new request to page 1?

im working through a follow up to handle that part - this first part is just reducing the dom impact when we pass in a full list of options (no pagination)

bmartel:
@yyassi-heartex When searching, can we trigger a recalc of the virtual container so that it isn't displaying empty space? Or what is contributing to that height being retained?

web/libs/ui/src/lib/select/select.tsx
```diff
@@ -77,6 +78,7 @@ export const Select = forwardRef(
       selectedValueRenderer,
       selectFirstIfEmpty,
       renderSelected,
+      useInfiniteScroll = false,
```
bmartel: I would rename this to:


```suggestion
      isVirtualList = false,
```

web/libs/datamanager/src/components/Filters/FilterDropdown.jsx
```diff
@@ -56,6 +56,7 @@ export const FilterDropdown = observer(
         searchable={true}
         triggerClassName="whitespace-nowrap"
         searchFilter={searchFilter}
+        useInfiniteScroll={true}
```
bmartel:
```suggestion
        isVirtualList={true}
```

web/libs/ui/src/lib/select/select.tsx
```diff
@@ -281,72 +352,22 @@ export const Select = forwardRef(
...
-                        disabled={option?.disabled}
-                        style={option?.style}
-                        multiple={multiple}
-                        onSelect={() => {
-                          _onChange(optionValue, isOptionSelected);
-                        }}
-                      />
-                    );
-                  })}
+                  {useInfiniteScroll ? (
```
bmartel:
```suggestion
                  {isVirtualList ? (
```

web/libs/ui/src/lib/select/types.ts
```diff
@@ -72,6 +72,7 @@ export type SelectProps<T, A extends SelectOption<T>[]> = {
   onSearch?: (value: string) => void;
   selectFirstIfEmpty?: boolean;
   renderSelected?: (selectedOptions?: A[number][], placeholder?: string) => React.ReactNode | string;
+  useInfiniteScroll?: boolean;
```
bmartel:
```suggestion
  isVirtualList?: boolean;
```

yyassi-heartex:
> @yyassi-heartex When searching, can we trigger a recalc of the virtual container so that it isn't displaying empty space? Or what is contributing to that height being retained?

the library we use asks for a value for the height which it uses but im sure we can do something with that

web/libs/ui/src/lib/select/select.tsx
```diff
@@ -281,72 +352,22 @@ export const Select = forwardRef(
...
-                        }}
-                      />
-                    );
-                  })}
+                  {isVirtualList ? (
+                    <VariableSizeList
+                      itemData={renderedOptions}
+                      itemSize={() => 40}
+                      itemCount={renderedOptions.length}
+                      height={200}
```
bmartel: Can this be adapted to being a non hard-coded height? I'm thinking that is what is causing the list sizes smaller than this height to have empty space, particularly on searching but I would imagine it would happen anytime the options would be only a few.

yyassi-heartex: yeah i hard coded it for now but I'm wondering if we should have it as a prop and put the responsibility on the wrapper component to define how high that block should be - this way if we want to dynamically update - thats fine - if we want to set it to something that makes sense for that component - thats fine too

web/libs/datamanager/src/components/CellViews/Annotators/Annotators.jsx
```diff
@@ -72,6 +80,79 @@ export const Annotators = (cell) => {
   );
 };
 
+export const InfiniteVariantSelect = observer(
+  ({ filter, schema, onChange, multiple, value, placeholder, disabled, id }) => {
+    if (!schema) return <></>;
+    const { items } = schema;
```
bmartel:
```suggestion
export const InfiniteVariantSelect = observer(
  ({ filter, onChange, multiple, value, placeholder, disabled, id }) => {
```

bmartel: This doesn't look used, and having a conditional return above hooks violates the rules of hooks which would likely throw errors.

web/libs/datamanager/src/components/CellViews/Annotators/Annotators.jsx
```diff
@@ -72,6 +80,79 @@ export const Annotators = (cell) => {
   );
 };
 
+export const InfiniteVariantSelect = observer(
```
nass600: This component should probably be a a molecule in app-core (`UserSelect` or `UserDropdown` or similar) holding the extra logic of the user while rendering the select given we will need this component in Filters, DM action modals and more places potentially

web/libs/datamanager/src/components/CellViews/Annotators/Annotators.jsx
yyassi-heartex: result of copying the original VariantSelect

yyassi-heartex: done

web/libs/datamanager/src/components/CellViews/Annotators/Annotators.jsx
yyassi-heartex: great callout! I refactored it to UserSelect but kept it with in the web/libs/datamanager library because the underlying hook is using the datamanger store.apiCall. this will make it reusable with other parts of datamanger but we cant really use it outside of that.
once all the underlying apiCalls are standardized and moved into a common place we should be able to move it up into a more core spot.

bmartel (approved):

nass600 (approved):

PR #8054, opened 2025-07-28
fix: BROS-254: Fix schema updates in "Build docs" CI

hlomzik:
`create-docs` script should update both tags documentation and autocomplete schema. It still does it, but commit step ignores new schema. This fix adds this file back to git commit step.

### DIFF
--- a/web/libs/core/src/lib/utils/schema/tags.json
+++ b/web/libs/core/src/lib/utils/schema/tags.json
@@ -69,6 +69,13 @@
         "required": false,
         "default": 32
       },
+      "spectrogram": {
+        "name": "spectrogram",
+        "description": "Determines whether an audio spectrogram is automatically displayed upon loading.",
+        "type": ["true", "false"],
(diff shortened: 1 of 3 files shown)


web/libs/core/src/lib/utils/schema/tags.json
```diff
@@ -69,6 +69,13 @@
         "required": false,
         "default": 32
       },
+      "spectrogram": {
```
niklub, line 72: can it be automated along with doc generation?

hlomzik: that's what the fix does

hlomzik: the last commit was created by automation: https://github.com/HumanSignal/label-studio/pull/8054/commits/fa5007ba0cdbe5d980c7522e3e5dc6ea63b4f882

niklub (approved):

PR #8081, opened 2025-08-01
fix: FIT-497: User display name sometimes not correctly showing for datamanager users

bmartel:
Before:

Screenshot 2025-08-01 at 3 37 23 PM

After:

Screenshot 2025-08-01 at 3 36 48 PM


web/libs/core/src/lib/utils/helpers.ts
```diff
@@ -19,7 +19,14 @@ export const isDefined = <T>(value: T | undefined | null): value is T => {
...
   if (!user) return "";
-  const { firstName, lastName, username, email } = user;
+  let { firstName, lastName, first_name, last_name, username, email } = user;
+
+  if (first_name) {
+    firstName = first_name;
+  }
+  if (last_name) {
+    lastName = last_name;
+  }
```
luarmr, line 29: interesting, took me a while

luarmr (approved):

yyassi-heartex (approved):

PR #8085, opened 2025-08-04
fix: BROS-274: Update hotkeys.ts

deppp:
adding the lookup for a custom hotkey

PR #8088, opened 2025-08-05
fix: BROS-274: Fix app crash when no hotkeys found

hlomzik:
Screenshot 2025-08-05 at 09 36 50

Screenshot 2025-08-05 at 09 37 15

Follow-up for https://github.com/HumanSignal/label-studio/pull/8085

### DIFF
--- a/web/libs/datamanager/src/sdk/hotkeys.ts
+++ b/web/libs/datamanager/src/sdk/hotkeys.ts
@@ -33,7 +33,7 @@ export const useShortcut = (
 
   // Check for custom shortcut in app settings
   const customMapping = window.APP_SETTINGS?.lookupHotkey?.(`data_manager:${actionName}`);
-  if (customMapping !== undefined) {
+  if (customMapping) {
     // Explicitly use the custom key even if it's null
     shortcut = customMapping.key;
   }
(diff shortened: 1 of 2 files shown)


web/libs/datamanager/src/sdk/hotkeys.ts
```diff
@@ -33,7 +33,7 @@ export const useShortcut = (
 
   // Check for custom shortcut in app settings
   const customMapping = window.APP_SETTINGS?.lookupHotkey?.(`data_manager:${actionName}`);
-  if (customMapping !== undefined) {
+  if (customMapping) {
     // Explicitly use the custom key even if it's null
     shortcut = customMapping.key;
```
niklub, line 38: I believe the correct fix should be 
```
if (customMapping !== undefined) {
   shortcut = customMapping?.key
}
```

niklub (approved):

web/libs/datamanager/src/sdk/hotkeys.ts
hlomzik: `lookupHotkey()` might return `null`, TypeScript should complain about this, so checking just for undefined is wrong

PR #8108, opened 2025-08-06
feat: BROS-194: Custom tags

hlomzik:
## Summary

Enable platform support for custom tags and add backend validation for Chat data. This PR provides backend validation and a feature-flag switch that allows downstream apps to register custom tags.

## Key highlights in this PR

- **Allow to register custom tag**
  - Introduce/enable custom-tag registration through the editor registry so downstream apps can wire a tag's `model`, `view`, and `region` in one call.
  - Relevant: `web/libs/editor/src/core/Registry.ts` (`addCustomTag` hooks into `addTag`, `addObjectType`, `addRegionType`).

- **Include registered items in all required places for a new tag to work**
  - Registration wires the tag across the internal registries: models/views, object type mapping, and region type detection.
  - Deserialization is adjusted to avoid mis-detecting regions vs. classifications (see `omitValueFields` below), ensuring registered region types are detected correctly.

- **Utilize `hideable` param (previously unused in some UI paths)**
  - Respect `region.hideable`/`item.hideable` when rendering visibility controls in Outliner and Details panel.
  - Relevant: `web/libs/editor/src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx`, `web/libs/editor/src/components/SidePanels/DetailsPanel/RegionItem.tsx`.

- **Simplify `omitValueFields` to only address Text vs TextArea confusion**
  - Narrowed logic to only strip `value.text` when it's an array (TextArea case) under `FF_CUSTOM_TAGS`, preventing conflicts that caused wrong region/classification detection.
  - Relevant: `web/libs/editor/src/stores/Annotation/Annotation.js` (`omitValueFields`).

## Other Changes

- **Backend**
  - Extend `TaskValidator` data types to include `Chat` so tasks with chat data validate correctly.

- **Feature flags**
  - Enable `FF_CUSTOM_TAGS` override in development/OSS builds to allow experimentation and downstream registration of custom tags. No LS UI exposure in this PR.

## Affected Files (high-level)

- `label_studio/tasks/validation.py`
- `web/libs/core/src/lib/utils/feature-flags/ff.ts`
- `web/libs/editor/src/core/Registry.ts`
- `web/libs/editor/src/stores/Annotation/Annotation.js`
- `web/libs/editor/src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx`
- `web/libs/editor/src/components/SidePanels/DetailsPanel/RegionItem.tsx`

## Feature Flags

- `FF_CUSTOM_TAGS`: enabled by default in OSS/dev via override.

## Why

- Allow downstream applications to register and use custom tags. Add backend support for Chat-type task data so such data validates correctly when present. Editor changes are limited to plumbing for registration, honoring `hideable`, and deserialization safety; no new user-facing features in LS.

## Testing

1. Backend validation
   - Import/create tasks where labeling config includes a Chat object and task data contains chat content.
   - Verify tasks pass validation and can be imported.

## Rollout / Risk

- Additive backend and internal plumbing for custom tags; no new LS user-facing features.
- No migrations required.

## Docs

- Custom tag registration is intended for downstream applications. No user-facing LS documentation changes are included here.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16887546298)
> Successfully merged:  create mode 100644 web/libs/editor/src/lib/AudioUltra/Visual/Renderer/ResizeRenderer.ts

web/libs/editor/src/stores/Annotation/Annotation.js
```diff
@@ -80,14 +81,21 @@ const hotkeys = Hotkey("Annotations", "Annotations");
  * @param value {Object} object to fix
  * @returns {Object} new object without value fields
  */
-function omitValueFields(value) {
+const omitValueFields = ff.isActive(ff.FF_CUSTOM_TAGS) ? (value) => {
+  // @todo describe that we only omit `text` from TextArea
```
bmartel: Is this true? The original code looks like it is possibly omitting far more 🤯

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16962215812)
> Successfully merged:  create mode 100644 docs/themes/v2/source/images/project/project_members_dashboard.png

web/libs/editor/src/stores/Annotation/Annotation.js
hlomzik: Yeah, we needed it only because of this confusion. I added some info into commit message (https://github.com/HumanSignal/label-studio/pull/8108/commits/60ca64c07c31f4f05d3a8b5d299b95d1f96f2996), but will write a comment here as well.

web/libs/editor/src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
```diff
@@ -424,16 +424,17 @@ const RootTitle: FC<any> = observer(
...
-            item={item}
-            entity={props.entity}
-            regions={props.children}
-            type={props.type}
-            collapsed={collapsed}
-            hasControls={hasControls && isArea}
-            toggleCollapsed={toggleCollapsed}
-          />
+          {item.hideable && (
+            <RegionControls
```
Gondragos, line 428: I'm not sure we really want to remove all controls under this condition. There's functionality like `"Copy Region Link"`, and I don't yet understand whether it requires removal or will operate as intended if retained.

web/libs/editor/src/core/Registry.ts
```diff
@@ -111,6 +132,13 @@ class _Registry {
   getPerRegionView(tag: string | number, mode: string | number) {
     return this.perRegionViews[tag]?.[mode];
   }
+
+  addCustomTag(tag: string, definition: CustomTag) {
+    this.addTag(tag.toLowerCase(), definition.model, definition.view);
+    this.addObjectType(definition.model);
```
Gondragos, line 138: It looks like we won't be able to add something different from the object tag to the Custom tags.
Is it expectable behaviour?

web/libs/editor/src/regions/Result.js
```diff
@@ -60,9 +61,65 @@ const Result = types
...
+      "choices",
+      "datetime",
+      "number",
+      "taxonomy",
+      "textarea",
+      "rating",
+      "pairwise",
+      "videorectangle",
+      "ranker",
+      ...Registry.customTags.map((t) => t.resultName),
```
Gondragos: This might be unnecessary here.

web/libs/editor/src/regions/Result.js
```diff
@@ -32,7 +33,7 @@ const Result = types
     // object tag
     to_name: types.late(() => types.reference(types.union(...Registry.objectTypes()))),
     // @todo some general type, maybe just a `string`
-    type: types.enumeration([
+    type: ff.isActive(ff.FF_CUSTOM_TAGS) ? types.late(() => types.enumeration([
```
Gondragos: It might be easier to use an approach like this for such a large list:
```
(ff.isActive(ff.FF_CUSTOM_TAGS) ? type => types.late(() => type) : type => type)(types.enumeration([...]))
```
At least it's better to avoid some how duplication of the list itself. A new `const` is suitable as well.

web/libs/editor/src/regions/Result.js
```diff
@@ -87,6 +144,7 @@ const Result = types
       bitmasklabels: types.maybe(types.array(types.string)),
       taxonomy: types.frozen(), // array of arrays of strings
       sequence: types.frozen(),
+      ...Object.fromEntries(Registry.customTags.map((t) => [t.resultName, types.maybe(t.result)])),
```
Gondragos: This might be unnecessary here.

web/libs/editor/src/regions/Result.js
```diff
@@ -60,9 +61,65 @@ const Result = types
...
+      "textarea",
+      "rating",
+      "pairwise",
+      "videorectangle",
+      "ranker",
+      ...Registry.customTags.map((t) => t.resultName),
     ]),
     // @todo much better to have just a value, not a hash with empty fields
-    value: types.model({
+    value: ff.isActive(ff.FF_CUSTOM_TAGS) ? types.late(() => types.model({
```
Gondragos: The definitions of `model` for `value` are essentially the same.  They may only require a condition for adding custom tags, but in general we may avoid the duplicating.

web/libs/editor/src/stores/Annotation/Annotation.js
Gondragos: We have a point in the code where the serialized result is deserialized, and at this point all result data is duplicated both in value and at the top level. Because in RichText we decided to store the text in the serialized result, and both RichText region and TextArea use the `"text"` field to store their data, when passing it to the top level, we have to filter the `"text"` field to avoid type conflicts. After determining the type, we add this value "manually" using `applyAdditionalDataFromResult`.

However, yes, this is the only detected case at the moment.

(I really hope I remember the reason of this function correctly)

web/libs/editor/src/regions/Result.js
hlomzik: oh, you are right! we should not use custom tags without flag for feature flags! will fix this in a sec

web/libs/editor/src/regions/Result.js
hlomzik: yeah, improved it, thanks!

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17028520741)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/control_tags/text-with-dual-taxonomy.cy.ts

Gondragos (approved):

PR #8113, opened 2025-08-07
fix: BROS-263: Fix image aspect ratio when cached

nick-skriabin:
When caching images we must use `pixelRatio: 1` to preserve image's actual size.

By default Konva uses devicePixelRatio which will be 2 on Retina displays (might be higher for higher density displays).

https://konvajs.org/api/Konva.Node.html#cache


hlomzik (approved):
Oh, things like this are tough to spot! Thank you!
Add a link to Konva docs/SO post if possible

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16860969445)
> Successfully merged:  create mode 100644 label_studio/tests/data_import/test_streaming_import.py

PR #8130, opened 2025-08-11
feat: FIT-493: Make left column of User Assigner infinitely paginated and searchable

yyassi-heartex:
This pull request makes a minor adjustment to the `Select` component in `select.tsx` to standardize the height of the dropdown list. Instead of dynamically sizing the dropdown based on the number of available items, it now always renders space for five items.

* UI consistency: The dropdown list in the `Select` component now always renders space for five items, regardless of the number of available options. This change improves consistency in the user interface.


borisheartex:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/16933827651)

PR #8144, opened 2025-08-13
fix: ROOT-72: unblock docker image by removing transitional package reference

jombooth:
When it rains, it pours - this package isn't in the newer Linux image we seem to have started using during Docker builds.

https://askubuntu.com/a/1520333

Also, stop trying to get nginx from nginx.org for now - there is no build for `trixie` listed here: http://nginx.org/packages/debian/dists/


jombooth:
didn't work, falling back to slim-bookworm instead.

PR #8148, opened 2025-08-13
fix: BROS-255: Fix bbox flipping

Gondragos:
With this PR flipping will:
- work with zoomed image
- work for horizontal flip even when rotation is not equal to 0
- not create 2 steps in history

Also the related test is adjusted


niklub:
cursor review

web/libs/editor/src/regions/RectRegion.jsx
```diff
@@ -333,6 +335,34 @@ const Model = types
...
     },
 
+    beforeSetPosition(x, y, width, height, rotation) {
+      // Konva flipping fix
+      if (height < 0) {
+        let flippedBack;
+        // If height is negative, it means it was flipped. We need to correct it
+        // Negative height also means that it was changed.
+        // In that case the difference between rotation and current rotation may be only 0 (or 360) and 180 degrees.
+        // However, as it's not an integer value, we had to check the difference to be sure,
```
hlomzik, line 345: can we just check for a rounded value so it will be more precise and readable? like `if (Math.round(deltaRotation) === 180)`

web/libs/editor/src/regions/RectRegion.jsx
```diff
@@ -362,8 +393,14 @@ const Model = types
...
 
         // Calculate snapped dimensions
-        const snappedWidth = bottomRightPoint.x - topLeftPoint.x;
-        const snappedHeight = bottomRightPoint.y - topLeftPoint.y;
+        let snappedWidth = bottomRightPoint.x - topLeftPoint.x;
+        let snappedHeight = bottomRightPoint.y - topLeftPoint.y;
+
+        // Ensure at least 1 pixel in size after snapping
+        const minPixelWidth = self.parent?.zoomedPixelSize?.x ?? 0;
+        const minPixelHeight = self.parent?.zoomedPixelSize?.y ?? 0;
```
hlomzik: should not it be 1 here?
```suggestion
        const minPixelWidth = self.parent?.zoomedPixelSize?.x ?? 1;
        const minPixelHeight = self.parent?.zoomedPixelSize?.y ?? 1;
```

hlomzik (approved):

Gondragos:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17030291842)
> Successfully merged:  create mode 100644 web/libs/editor/tests/integration/e2e/control_tags/text-with-dual-taxonomy.cy.ts

PR #8062, opened 2025-07-29
feat: UTC-210: Prediction validation

robot-ci-heartex:
This PR was automatically generated via Follow Merge. [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/16887182584)
Please ensure that all linked upstream Pull Requests are merged before proceeding with this one.
- https://github.com/HumanSignal/label-studio-sdk/pull/496

### Prediction creation fully Covered with Validation

**UI Import (data_import/api.py - ImportPredictionsAPI)**
* Validated: Uses li.validate_prediction() with error collection
* Error Handling: Returns validation errors instead of raising exceptions

**PredictionAPI (tasks/serializers.py - PredictionSerializer)**
* Validated: Has validate() method that uses li.validate_prediction()
* Error Handling: Raises ValidationError with detailed error messages

**Task Bulk Import (tasks/serializers.py - BaseTaskSerializerBulk.add_predictions())**
* Validated: Uses li.validate_prediction() with error collection
* Error Handling: Returns validation errors instead of raising exceptions

**ML Backend Predictions (ml/models.py - MLBackend.predict_tasks())**
* Validated: Uses PredictionSerializer.is_valid() which includes our validation
* Error Handling: Raises ValidationError if validation fails

**Storage Import (io_storages/base_models.py - ImportStorage.add_task())**
* Validated: Uses PredictionSerializer.is_valid() which includes our validation
* Error Handling: Collects all validation errors which will be visible to user under storage with new status "Completed with errors"

Example errors:

**UI import** 
image

**Cloud Sync**
image
image

**Prediction API**
image

PR #8165, opened 2025-08-15
feat: UTC-210: Gating predictions validation raising errors behind fflag

mcanu:
Gates most of the features related to predictions validations introduced in https://github.com/HumanSignal/label-studio/pull/8062 behind `fflag_feat_utc_210_prediction_validation_15082025` fflag.

### DIFF
--- a/label_studio/data_import/api.py
+++ b/label_studio/data_import/api.py
@@ -267,7 +267,7 @@ class ImportAPI(generics.CreateAPIView):
 
         if preannotated_from_fields:
             # turn flat task JSONs {"column1": value, "column2": value} into {"data": {"column1"..}, "predictions": [{..."column2"}]
-            raise_errors = flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=self.request.user)
+            raise_errors = flag_set('fflag_feat_utc_210_prediction_validation_15082025', user='auto')
             parsed_data = reformat_predictions(parsed_data, preannotated_from_fields, project, raise_errors)
 
         # Conditionally validate predictions: skip when label config is default during project creation
@@ -292,7 +292,7 @@ class ImportAPI(generics.CreateAPIView):
                 for error in validation_errors:
                     error_message += f'- {error}\n'
 
-                if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=self.request.user):
+                if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user='auto'):
                     raise ValidationError({'predictions': [error_message]})
                 else:
                     logger.error(f'Prediction validation failed ({len(validation_errors)} errors):\n{error_message}')
@@ -523,7 +523,7 @@ class ImportPredictionsAPI(generics.CreateAPIView):
         for i, item in enumerate(self.request.data):
             # Validate task ID
             if item.get('task') not in tasks_ids:
-                if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=self.request.user):
+                if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user='auto'):
                     validation_errors.append(
                         f'Prediction {i}: Invalid task ID {item.get("task")} - task not found in project'
                     )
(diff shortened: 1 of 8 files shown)


label_studio/data_import/functions.py
```diff
@@ -73,10 +74,13 @@ def async_import_background(
...
                 error_message += f'- {error}\n'
 
-            project_import.error = error_message
-            project_import.status = ProjectImport.Status.FAILED
-            project_import.save()
-            return
+            if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=user):
+                project_import.error = error_message
+                project_import.status = ProjectImport.Status.FAILED
+                project_import.save()
```
jombooth: looks like this change predates the PR, but I think we should use `update_fields` here for sure

label_studio/data_import/api.py
```diff
@@ -519,10 +523,17 @@ def _create_legacy(self, project):
         for i, item in enumerate(self.request.data):
             # Validate task ID
             if item.get('task') not in tasks_ids:
-                validation_errors.append(
-                    f'Prediction {i}: Invalid task ID {item.get("task")} - task not found in project'
-                )
-                continue
+                if flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=self.request.user):
```
jombooth: Don't we want `org.created_by` here like our usual strategy for flag_set?

jombooth: (here and other call locations)

mcanu: I'm pretty sure there is no need to use the organization created_by. We set the organization attribute [here](https://github.com/HumanSignal/label-studio/blob/292201317f9d9be09af1548cd3aba090d552f6c5/label_studio/core/feature_flags/utils.py#L8) based on that, and then we can target accordingly in Launch Darkly.
I see it used like this in several other places too, maybe I'm missing something

jombooth: ah, got it. The place where it's most important to pass organization.created_by is in async code where we don't have a request. To be consistent here, assuming this code only runs sync, can we use `user='auto'`?

mcanu: Sorry I don't think I'm following. If we use `user='auto'` here we will lose the per org targeting functionality. This particular line is run sync so there is no problem.

I did have to use the organization.created_by user in the [ImportStorage.add_task](https://github.com/HumanSignal/label-studio/blob/2d0e90678f1c1dc0a146f64a3e087e06cbda6f26/label_studio/io_storages/base_models.py#L503) function.

mcanu: oh, my bad, I didn't know that auto did [this](https://github.com/HumanSignal/label-studio/blob/292201317f9d9be09af1548cd3aba090d552f6c5/label_studio/core/feature_flags/base.py#L94-L98). I'll change it.

jombooth: I don't think that's true - if we pass `auto` we'll [pull the user off the request like the current implementation does](https://github.com/HumanSignal/label-studio/blob/d91109a07127545651339f7707644efbd546d562/label_studio/core/feature_flags/base.py#L94-L98). Basically I just think we have too many patterns for this and would like to standardize on just user auto for sync contexts, and created_by for async contexts

mcanu: Updated!

label_studio/data_import/functions.py
```diff
@@ -48,7 +48,8 @@ def async_import_background(
 
     if project_import.preannotated_from_fields:
         # turn flat task JSONs {"column1": value, "column2": value} into {"data": {"column1"..}, "predictions": [{..."column2"}]
-        tasks = reformat_predictions(tasks, project_import.preannotated_from_fields, project)
+        raise_errors = flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=user)
```
jombooth: please don't hate me but I think this one needs to be
```suggestion
        raise_errors = flag_set('fflag_feat_utc_210_prediction_validation_15082025', user=project.organization.created_by)
```

jombooth: (here and below)

mcanu: no worries! I'll change it

jombooth (approved):
🙏

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17000078841)

Issue #7175, opened 2025-03-06
Unable to start Label Studio in an airgapped environment


DrissiReda:
I'm trying to start label studio in a kubernetes cluster, however it keeps trying to check the latest version then fails and crashes. 

What I've tried: 
- Run 1.16.0 in an offline kubernetes cluster: crashes on start.
- Run 1.16.0 with the same configuration in an online server: works.
- Tried 1.15.0 and 1.14.0 on the offline cluster: crashes on start.
- Tried 1.13.1 on the offline cluster: works.

Is there any way to disable latest version check? It makes no sense in an offline cluster.

<details>
<summary>Relevant logs</summary>

```bash
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "  File \"/label-studio/label_studio/core/settings/label_studio.py\", line 49, in <module>"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "  File \"/usr/local/lib/python3.12/importlib/__init__.py\", line 90, in import_module"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "    check_for_the_latest_version(print_message=True)"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "  File \"/label-studio/label_studio/core/utils/common.py\", line 394, in check_for_the_latest_version"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "    from label_studio.core.utils.common import collect_versions"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "  File \"/label-studio/label_studio/core/utils/common.py\", line 420, in <module>"}
{"timestamp": "06/Mar/2025:09:19:09 +0000", "source": "uwsgi", "levelname": "INFO", "message": "    application = get_wsgi_application()"}
[...]

PR #8185, opened 2025-08-19
feat: allow to disable version check

ParminderParmar:
Reason for change

In airgapped (offline) environments, the application attempts to check for the latest version online and fails with a network error. This causes unnecessary errors in logs and potential confusion for users.

This PR introduces an environment variable ALLOW_LATEST_VERSION_CHECK (default: True) that allows administrators to disable the version check when running in restricted environments.

Relevant issue: [#7175](https://github.com/HumanSignal/label-studio/issues/7175)

Code change
```python
LATEST_VERSION_CHECK = get_bool_env('ALLOW_LATEST_VERSION_CHECK', True)
``` 

Screenshots
N/A (no visible UI changes).

Rollout strategy
- Controlled via the ALLOW_LATEST_VERSION_CHECK environment variable.
- Default behavior remains unchanged (True), so existing deployments are not affected.
- Airgapped users can explicitly set ALLOW_LATEST_VERSION_CHECK=False to disable online version checks.

Testing
- Verified in an airgapped server by setting ALLOW_LATEST_VERSION_CHECK=False → no more failed requests.
- Verified in a normal environment with default setting (True) → version check continues to work.

Risks
- Minimal. Default behavior is preserved.
- Only risk is that some users may disable version checking unintentionally, but this is explicit via env var.

General notes
This PR provides a safe and configurable way to run Label Studio in offline/airgapped environments without spurious errors.
It addresses a common pain point for enterprise users operating without internet access.


robot-ci-heartex:

robot-ci-heartex:

PR #8238, opened 2025-08-22
fix: BROS-352: Revert order of returned params in `useCopyText()`

hlomzik:
This order was not supported by all usages.

image (11)


bmartel (approved):
Are we sure these are the only affected usages?

bmartel:
I checked the other usages, this will work. We need to address the inconsistent usage of this, tuples should always be [state, setState] when creating hooks like this.

bmartel:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17166365180)
> Successfully merged:  3 files changed, 15 insertions(+), 9 deletions(-)

PR #8241, opened 2025-08-22
fix: Prevent duplicate ImageEntity IDs in React StrictMode

luarmr:
## Description

PLAYGROUND ERRORS

Fixes a MobX State Tree reference resolution error that occurs when using React StrictMode during development. The error manifests as:

```
[mobx-state-tree] Cannot resolve a reference to type 'ImageEntity' with id: 'image#0@L55lJ' unambiguously, there are multiple candidates: /annotationStore/annotations/0/root/children/1/imageEntities/0, /annotationStore/annotations/0/root/children/1/imageEntities/1
```

## Root Cause

React StrictMode intentionally double-mounts components in development to help detect side effects. This caused the `createImageEntities()` function to be called twice during component mounting:

1. **First mount**: Creates ImageEntity with `id: "image#0@L55lJ"`
2. **Second mount**: Tries to create another ImageEntity with the same `id: "image#0@L55lJ"`
3. **Result**: Duplicate IDs in the MobX State Tree, causing reference resolution to fail

The issue became more prominent after the April 2024 changes that added annotation ID postfixes to ImageEntity IDs (`${self.name}#${index}@${self.annotation.id}`), making the IDs more complex and the duplicate detection more strict.

## Solution

Added `self.imageEntities.clear()` at the beginning of `createImageEntities()` to ensure any existing entities are removed before creating new ones. This prevents duplicate IDs regardless of how many times the function is called.

## Changes

- **Modified**: `libs/editor/src/tags/object/Image/Image.js`
  - Added `self.imageEntities.clear()` in `createImageEntities()` function
  - Added explanatory comment about React StrictMode double mounting

## Testing

- ✅ Verified the MobX State Tree error no longer occurs in development
- ✅ Confirmed image loading and display functionality remains intact
- ✅ No breaking changes to existing behavior

## Impact

- **Risk**: Low - only adds a defensive clear operation before entity creation
- **Backward compatibility**: Full - no API or behavior changes

## Related Issues

This fix addresses the underlying cause rather than requiring workarounds like disabling React StrictMode or reverting UI component changes that may have exacerbated the mounting cycles.


luarmr:
@hlomzikI think you will be interested in this one. It's related to the one we fixed recently where we thought the FF was the culprit. Maybe it wasn't.

If you agree with this, feel free to merge after approve. The playground deploy will occur just merging.

nick-skriabin (approved):

yyassi-heartex (approved):

luarmr:
/git merge 
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17244561659)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/inbox.svg

PR #8266, opened 2025-08-25
feat: UTC-235: Add prediction model versions to member dashboard

hakan458:


label_studio/ml_models/models.py
```diff
@@ -201,12 +202,16 @@ def delete_predictions(self):
         # to delete all dependencies where predictions are foreign keys.
         Annotation.objects.filter(parent_prediction__in=prediction_ids).update(parent_prediction=None)
         try:
-            from stats.models import PredictionStats
+            from stats.models import PredictionPairStats, PredictionStats
```
mcanu, line 205: Weird that we do this in LSO, but this was already happening for PredictionStats

mcanu (approved):

PR #8272, opened 2025-08-26
feat: FIT-587: FSM Architecture setup

bmartel:
This pull request introduces a new Finite State Machine (FSM) framework to the Label Studio codebase. The FSM system is designed for high-performance, time-series optimized state management using UUID7, and provides a generic, extensible infrastructure for tracking entity states and transitions. The changes include core model definitions, registry and transition systems, documentation, and Django app configuration. Additionally, the FSM framework is integrated into the main application settings with initial configuration.

**FSM Framework Introduction and Integration**

* Added the `fsm` app to the `INSTALLED_APPS` list in `base.py`, integrating the FSM framework into Label Studio's Django settings.
* Introduced FSM cache TTL configuration (`FSM_CACHE_TTL`) for controlling state cache duration.

**Core FSM Implementation**

* Implemented the core FSM model (`BaseState`) in `fsm/models.py`, featuring UUID7 primary keys for natural time ordering, audit trail fields, transition metadata, and optimized query methods for state history and time-range queries.
* Added initial state choices module (`fsm/state_choices.py`) for defining available states per entity type.

**Framework Documentation and App Structure**

* Added extensive documentation in `fsm/README.md`, detailing architecture, usage patterns, extension points, and performance characteristics of the FSM system.
* Created Django app configuration (`fsm/apps.py`) and package initialization (`fsm/__init__.py`) for the FSM framework. [[1]](diffhunk://#diff-b0bbdfca422af48284839dab1ab9ccb553fae0b4b4d3a2d0b3a8f1d68f103757R1-R13) [[2]](diffhunk://#diff-f699e3a7ad883bd25aaf854d487e2c1035d2c0b729a6eab3969f71defd15ab86R1-R12)
* Established migration and test package structure for future development (`fsm/migrations/__init__.py`, `fsm/tests/__init__.py`). [[1]](diffhunk://#diff-e46226c21c60230f56b6eafe1e0854fd24d92576a3cda5d10314d0b6044bcb5fR1) [[2]](diffhunk://#diff-56c3cc9bf9cd1c73e27ce972b2b564e16af7f42507d93516069de78ce50ac981R1)


label_studio/fsm/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+The framework provides generic REST endpoints:
+
+```
+GET /api/fsm/{entity_type}/{entity_id}/current/     # Current state
+GET /api/fsm/{entity_type}/{entity_id}/history/     # State history  
+POST /api/fsm/{entity_type}/{entity_id}/transition/ # Execute transition
+```
+
+Extend the base viewset for your application:
```
bmartel:
```suggestion
Extend the base viewset
```

label_studio/fsm/tests/test_api_usage_examples.py
```diff
@@ -0,0 +1,792 @@
...
+                    'assignment_details': result_data,
+                },
+                'timestamp': datetime.now().isoformat(),
+            }
+
+            # Validate API response
+            self.assertTrue(api_response['success'])
+            self.assertEqual(api_response['data']['new_state'], 'ASSIGNED')
+            self.assertEqual(api_response['data']['assignment_details']['assignee_id'], 123)
+            self.assertEqual(api_response['data']['assignment_details']['priority'], 'high')
```
mcanu: To stay consistent with the rest of the tests lets use native `assert`

label_studio/fsm/tests/test_api_usage_examples.py
mcanu, line 1: Lets rename this to test_api.py, the `APIIntegrationExampleTests` is already a logic separation. Or a `test_api.example.py` or `test_api_example.py` can also work.

We could also create a base class for new endpoints tests to inherit and have base assertions, but out of scope for this first approach.

label_studio/fsm/tests/test_edge_cases_error_handling.py
mcanu, line 1: `test_error_handling.py`

label_studio/fsm/tests/test_uuid7_utils.py
mcanu, line 1: `test_utils.py`

label_studio/fsm/apps.py
```diff
@@ -0,0 +1,46 @@
...
+        """Set up signal handlers for automatic state creation"""
+        try:
+            from django.conf import settings
+
+            # Only set up signals if enabled in settings
+            if getattr(settings, 'FSM_AUTO_CREATE_STATES', False):
+                logger.info('FSM signal handlers registered')
+
+        except Exception as e:
+            logger.error(f'Failed to set up FSM signals: {e}')
```
mcanu: Is this doing something?

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,434 @@
...
+        logger.debug('Cleared state model registry')
+
+    def get_all_models(self) -> Dict[str, Type['BaseState']]:
+        """Get all registered models."""
+        return self._models.copy()
+
+    def mark_initialized(self):
+        """Mark the registry as initialized."""
+        self._initialized = True
+        logger.info(f'State model registry initialized with {len(self._models)} models')
```
mcanu: Is this being used?

mcanu: We can remove it to simplify

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,434 @@
...
+class TransitionRegistry:
+    """
+    Registry for managing declarative transitions.
+
+    Provides a centralized way to register, discover, and execute transitions
+    for different entity types and state models.
+    """
+
+    def __init__(self):
+        self._transitions: Dict[str, Dict[str, Type['BaseTransition']]] = {}
```
mcanu:
```suggestion
        self._transitions: Dict[str, Dict[str, 'BaseTransition']] = {}
```
I think you can do this like this, please check

mcanu: No need to use `Type`

mcanu: Same for the rest of the `Type` usage

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,434 @@
...
+            class_name = transition_class.__name__
+            if class_name.endswith('Transition'):
+                class_name = class_name[:-10]  # Remove 'Transition' suffix
+
+            # Convert CamelCase to snake_case
+            name = ''
+            for i, char in enumerate(class_name):
+                if char.isupper() and i > 0:
+                    name += '_'
+                name += char.lower()
```
mcanu: I would make name required, just to be explicit about it and make it easier to search when troubleshooting.
Example, I see in the db a `start_task` transition and I want to search where its defined, I prefer not to remember this "magic" name creation.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+            if current_state == 'COMPLETED':
+                # Task is finished
+                pass
+        """
+        cache_key = cls.get_cache_key(entity)
+
+        # Try cache first
+        cached_state = cache.get(cache_key)
+        if cached_state is not None:
+            logger.debug(f'Cache hit for {entity._meta.label_lower} {entity.pk}: {cached_state}')
```
mcanu: Lets do this info and try to make these logs as structured as possible. I'd love to get some Grafana dashboard going with Cache Hit/Miss rates.

mcanu: Also log the miss.

We can revisit this if we see we are flooded with logs.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+            return None
+
+        try:
+            entity_field = f'{entity._meta.model_name}'
+            current_state = (
+                state_model.objects.filter(**{entity_field: entity})
+                .order_by('-id')  # UUID7 natural ordering
+                .values_list('state', flat=True)
+                .first()
+            )
```
mcanu: lets use `StateModel.get_current_state` here

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+
+        Returns:
+            Latest BaseState instance or None if no states exist
+        """
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            return None
+
+        entity_field = f'{entity._meta.model_name}'
+        return state_model.objects.filter(**{entity_field: entity}).order_by('-id').first()
```
mcanu: Same here, lets leave the access to the StateModel to itself.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            raise StateManagerError(f'No state model found for {entity._meta.model_name}')
+
+        current_state = cls.get_current_state(entity)
+
+        logger.info(
+            f'Transitioning {entity._meta.label_lower} {entity.pk}: '
+            f'{current_state} → {new_state} (transition: {transition_name})'
+        )
```
mcanu: Same here with structured logging. I can see a dashboard per Model state so we transactions being used. We can probably use the db for that, but I don't think we have any dashboards coming straight from the database.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+
+        try:
+            with transaction.atomic():
+                # INSERT-only approach - no UPDATE operations needed
+                # Get denormalized fields from the state model itself
+                denormalized_fields = {}
+                if hasattr(state_model, 'get_denormalized_fields'):
+                    denormalized_fields = state_model.get_denormalized_fields(entity)
```
mcanu: lets make `get_denormalized_fields` a default method in the base model and return an empty dict

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+
+        Returns:
+            List of state records ordered by most recent first
+        """
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            return []
+
+        entity_field = f'{entity._meta.model_name}'
+        return list(state_model.objects.filter(**{entity_field: entity}).order_by('-id')[:limit])
```
mcanu: We already have this method in the State model. Also I wouldn't cast it to a list immediately, that would execute a query and not make it lazy in case it isn't used downstream

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+            end_time: End of time range (defaults to now)
+
+        Returns:
+            List of states within the time range
+        """
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            return []
+
+        return list(state_model.get_states_in_range(entity, start_time, end_time or datetime.now()))
```
mcanu: Same here with list

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+    def invalidate_cache(cls, entity: Model):
+        """Invalidate cached state for an entity"""
+        cache_key = cls.get_cache_key(entity)
+        cache.delete(cache_key)
+        logger.debug(f'Invalidated cache for {entity._meta.label_lower} {entity.pk}')
+
+    @classmethod
+    def warm_cache(cls, entities: List[Model]):
+        """
+        invalidate_cacheWarm cache with current states for a list of entities.
```
mcanu:
```suggestion
        Warm cache with current states for a list of entities.
```

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+
+    @classmethod
+    def invalidate_cache(cls, entity: Model):
+        """Invalidate cached state for an entity"""
+        cache_key = cls.get_cache_key(entity)
+        cache.delete(cache_key)
+        logger.debug(f'Invalidated cache for {entity._meta.label_lower} {entity.pk}')
+
+    @classmethod
+    def warm_cache(cls, entities: List[Model]):
```
mcanu, line 363: Love this thinking

label_studio/fsm/transition_utils.py
```diff
@@ -0,0 +1,316 @@
...
+        entity: The entity to get transitions for
+
+    Returns:
+        Dictionary mapping transition names to transition classes
+    """
+    entity_name = entity._meta.model_name.lower()
+    return transition_registry.get_transitions_for_entity(entity_name)
+
+
+def get_valid_transitions(entity: Model, user=None, validate: bool = True) -> Dict[str, Type[BaseTransition]]:
```
mcanu: Do we need the `validate` argument? We have the `get_available_transitions` for this.

label_studio/fsm/transition_utils.py
```diff
@@ -0,0 +1,316 @@
...
+
+    This provides a similar interface to StateManager.transition_state
+    but uses the declarative system.
+    """
+    return execute_transition(
+        entity=entity, transition_name=transition_name, transition_data=transition_data, user=user
+    )
+
+
+class TransitionBuilder:
```
mcanu: Is this something we see ourselves using? I'd limit the number of ways we can create/execute transactions so the code is more maintainable.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+
+        except Exception as e:
+            # On failure, invalidate potentially stale cache
+            cache_key = cls.get_cache_key(entity)
+            cache.delete(cache_key)
+            logger.error(
+                f'State transition failed for {entity._meta.label_lower} {entity.pk}: '
+                f'{current_state} → {new_state}: {e}'
+            )
+            raise StateManagerError(f'Failed to transition state: {e}') from e
```
mcanu, line 301: We aren't returning False if it failed, we are raising an error. Might be okay to raise an error and not return anything from this function. We'll need to update `BaseTransition.execute` that expects a boolean.

mcanu (commented):
Amazing work!! 😍

I'm not sure when I need to do a declarative transaction or not, lets leave that clear in the Readme. Or even limit to always wanting to create declarative transactions.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,385 @@
...
+                    triggered_by=user,
+                    context_data=context or {},
+                    reason=reason,
+                    organization_id=organization_id,
+                    **denormalized_fields,
+                )
+
+                # Update cache with new state
+                cache_key = cls.get_cache_key(entity)
+                cache.set(cache_key, new_state, cls.CACHE_TTL)
```
niklub: I wonder whether it's safer to write cache in the callback upon a successful commit, like
```
    transaction.on_commit(
        lambda: cache.set(cache_key, new_state, cls.CACHE_TTL)
    )
```

label_studio/fsm/apps.py
bmartel: It used to but I refactored a bunch of stuff as I didn't like the way this method of extending worked, I removed all of these in favour of just standard python OOP and decorators, and signals are not being used at this time, but even if they were, I'd prefer it if it was simply just a matter of standard Django signal usage. (tbh that is all this was doing, just in a more glorified manner 😆)

label_studio/fsm/transition_utils.py
bmartel: Good catch, this is another refactor left-over, need to clean these up and consolidate.

label_studio/fsm/transition_utils.py
bmartel: Yea I personally don't see it, will remove it as honestly this is a more Java-esque pattern, not at all pythonic. Another concept that is leftover from all of the dabbling and thought process.

label_studio/fsm/state_manager.py
bmartel: Good catch, I was pondering a concept to make the execution log or raise (reason being that in some circumstances, for example if we use this to backfill states, we wouldn't want it to fail hard, just log and move on). So I think in that regard we may want to tool the behaviour accordingly, and in the case of log only mode it should return whether it succeeded.

label_studio/fsm/transition_utils.py
bmartel: We do need a validate argument as there are two modes of retrieval, one which gets available states from my current (validate=True) and the default is all available states for the entity (validate=False). I removed this extra function and consolidated the approach.

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17446383898)

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17446529166)

label_studio/core/settings/base.py
```diff
@@ -891,3 +892,7 @@ def collect_versions_dummy(**kwargs):
 # Data Manager
 # Max number of users to display in the Data Manager in Annotators/Reviewers/Comment Authors, etc
 DM_MAX_USERS_TO_DISPLAY = int(get_env('DM_MAX_USERS_TO_DISPLAY', 10))
+
+# Base FSM (Finite State Machine) Configuration for Label Studio
+FSM_CACHE_TTL = 300  # Cache TTL in seconds (5 minutes)
+FSM_AUTO_CREATE_STATES = False
```
hakan458: Dont see any usages of this here or in LSE , is this future planned work?

label_studio/fsm/models.py
```diff
@@ -0,0 +1,168 @@
...
+
+    class Meta:
+        abstract = True
+        # UUID7 provides natural ordering, reducing index requirements
+        ordering = ['-id']  # Most recent first
+        get_latest_by = 'id'
+
+    def __str__(self):
+        entity_id = getattr(self, f'{self._get_entity_name()}_id', 'unknown')
+        return f'{self._get_entity_name().title()} {entity_id}: {self.previous_state} → {self.state}'
```
hakan458, line 91: curious why include previous_state if this is just the BaseState, and not a transition right?

label_studio/fsm/models.py
```diff
@@ -0,0 +1,168 @@
...
+        Check if this is a terminal state (no outgoing transitions).
+
+        Override in subclasses with specific terminal states.
+        """
+        return False
+
+    def _get_entity_name(self) -> str:
+        """Extract entity name from model name (e.g., TaskState → task)"""
+        model_name = self.__class__.__name__
+        if model_name.endswith('State'):
```
hakan458, line 116: When adding states to the platform will you / should we be following this pattern always?

label_studio/fsm/models.py
```diff
@@ -0,0 +1,168 @@
...
+        entity_field = f'{cls._get_entity_field_name()}'
+        queryset = cls.objects.filter(**{entity_field: entity})
+        return UUID7Field.filter_by_time_range(queryset, start_time, end_time).order_by('id')
+
+    @classmethod
+    def get_states_since(cls, entity, since: datetime):
+        """Get all states since a specific timestamp"""
+        entity_field = f'{cls._get_entity_field_name()}'
+        queryset = cls.objects.filter(**{entity_field: entity})
+        return UUID7Field.filter_since_time(queryset, since).order_by('id')
```
hakan458, line 160: Just want to say the usage of UUID7 is amazing and makes things so much easier / more performant 👏🏼 great choice

label_studio/fsm/models.py
bmartel: States keep the previous state for quick access without additional queries to know what the previous state was. (This is so that flows which can move to multiple possible next states, we can know at a glance which the previous was).

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,420 @@
...
+        Register a state model for an entity type.
+
+        Args:
+            entity_name: Name of the entity (e.g., 'task', 'annotation')
+            state_model: The state model class for this entity
+            denormalizer: Optional function to extract denormalized fields
+        """
+        entity_key = entity_name.lower()
+
+        if entity_key in self._models:
```
hakan458, line 127: when/why would we overwrite an existing state model? In what scope will this persist? E.g. can I have a different state model in the app vs rqworker for `task` ?

label_studio/fsm/models.py
bmartel: This is not a transition, just a denormalized optimization.

label_studio/fsm/models.py
bmartel: Yes, this is a great callout! The preferred method is convention over configuration and to follow as it documents to name your states according to the models they anchor to. If the model is an inherited proxy model which would extend the state choices, then you override this in method and return the string key that is intended. (You will see this in PRs for the model states and choices implementation).

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,420 @@
...
+    Registry for managing declarative transitions.
+
+    Provides a centralized way to register, discover, and execute transitions
+    for different entity types and state models.
+    """
+
+    def __init__(self):
+        self._transitions: Dict[str, Dict[str, 'BaseTransition']] = {}
+
+    def register(self, entity_name: str, transition_name: str, transition_class: 'BaseTransition'):
```
hakan458, line 247: just curious does the system allow to register multiple transitions, with the same current state and target state? would that be good practice? Im sure you have thought of some examples

label_studio/fsm/models.py
bmartel: This was a suggestion from @wesleylima 🚀

label_studio/fsm/registry.py
bmartel: No this is strictly an extension mechanism to allow for LSO -> LSE model states, as they differ greatly.

label_studio/fsm/state_manager.py
hakan458: Yeah that might be a lot of logs 😆

label_studio/fsm/registry.py
bmartel: Currently no, but that is a really neat idea 🤔. Will have to think more on that, but generally speaking I believe it could be made to work like that. Might be a more declarative way to go about the multi option transitions.

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,422 @@
...
+        # Try cache first
+        cached_state = cache.get(cache_key)
+        if cached_state is not None:
+            logger.debug(f'Cache hit for {entity._meta.label_lower} {entity.pk}: {cached_state}')
+            return cached_state
+
+        # Query database using state model registry
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            logger.warning(f'No state model found for {entity._meta.model_name}')
```
hakan458: Maybe add something like "when attempting to get current state". Otherwise unclear where this is coming from

label_studio/fsm/state_manager.py
hakan458: +1 , wonder if there is a reason why it wasnt used

label_studio/core/settings/base.py
bmartel: Yea I took this out entirely, or so I thought 😆. Nice catch

label_studio/fsm/state_manager.py
```diff
@@ -0,0 +1,422 @@
...
+                new_state='IN_PROGRESS',
+                transition_name='start_annotation',
+                user=request.user,
+                context={'assignment_id': assignment.id},
+                reason='User started annotation work'
+            )
+        """
+        state_model = get_state_model_for_entity(entity)
+        if not state_model:
+            raise StateManagerError(f'No state model found for {entity._meta.model_name}')
```
hakan458: + " ... when transitioning state" or similar

label_studio/fsm/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+# Get current state
+current_state = StateManager.get_current_state(order)
+
+# Get state history
+history = StateManager.get_state_history(order, limit=10)
+
+# Bulk operations for performance
+orders = Order.objects.all()[:1000]
+states = StateManager.bulk_get_current_states(orders)
```
hakan458: does this exist in LSO? Or only LSE?

bmartel: LSE, I'll remove this from here.

label_studio/fsm/state_manager.py
bmartel: I can structure this and move to level info 👍

label_studio/fsm/registry.py
hakan458: How is it enforced then, that we cannot register multiple transitions with the same state?

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17496376340)

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17497319557)

mcanu (approved):

label_studio/fsm/registry.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        transition_class = self.get_transition(entity_name, transition_name)
+        if not transition_class:
+            raise ValueError(f"Transition '{transition_name}' not found for entity '{entity_name}'")
+
+        # Create transition instance with provided data
+        transition = transition_class(**transition_data)
+
+        # Get current state information
+        from fsm.state_manager import StateManager
```
wesleylima: Is it useful to have these import ins the function call?

bmartel: Checking this, we seem to have an import cycle between state_manager and this file due to this method, I'll take note and see if this can get refactored so we don't require this. At this point I believe it was the easiest way to have the functionality culminate.

bmartel: I'll take a stab at refactoring this and update the tests accordingly. Nice catch!

wesleylima (approved):

PR #8296, opened 2025-08-28
docs fix

luarmr:
this is the fix in docs. I will upload the change in the generator


luarmr:
Forced since is only docs

PR #8301, opened 2025-08-29
fix: BROS-325: Don't show hidden per-regions in Info panel

hlomzik:
We store conditional per-regions in results even if the condition is not met the moment for better UX, so user can switch between different options and see previous choices for all of them.
But hidden choices won't be submitted.
And they also should not be visible in Info panel to not confuse the user. This fix is dedicated to hide them.

Sample config:
```xml
<View>
  <Text name="text" value="$text"/>

  <Labels name="is_correct" toName="text">
    <Label value="correct" background="#CFE1FC"/>
    <Label value="incorrect" background="#C04AC6"/>
  </Labels>

  <View visibleWhen="region-selected" whenLabelValue="incorrect">
    <Header size="4">Why?</Header>
    <Choices name="reason" toName="text"
             perRegion="true" whenLabelValue="incorrect">
      <Choice value="Good reason"/>
      <Choice value="No reason"/>
    </Choices>
  </View>
</View>
```
with data like `{"text":"Any text"}`

### Before

https://github.com/user-attachments/assets/2fc08595-b5c1-41ef-ae0f-476f2f5b903b

### After

https://github.com/user-attachments/assets/cfb299eb-ebc6-443a-b1d3-ff9be0380730


Gondragos (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17372451416)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/inbox.svg

PR #8304, opened 2025-08-29
fix: PLT-870: Stream tasks from big file on storage

triklozoid:


triklozoid:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17402677478)
> Successfully merged:  create mode 100644 web/libs/editor/src/tools/Vector.js

label_studio/io_storages/utils.py
```diff
@@ -181,6 +194,48 @@ def _error_wrapper(exc: Optional[Exception] = None):
     _error_wrapper()
 
 
+def _load_tasks_json_lso_generator(blob: bytes, key: str):
```
mcanu, line 197: In the current implementation we sometimes use bulk_create to persist the object in the database, shouldn't we do it too in this implementation?

mcanu (approved):

label_studio/io_storages/utils.py
triklozoid: That's not a django model, we create objects later.

PR #8325, opened 2025-09-03
fix: UTC-248: Annotator Performance workspace dropdown not visible when workspaces name length is too large

nass600:
Before:

Screenshot 2025-09-03 at 12 43 21
Screenshot 2025-09-03 at 12 45 11

After:
Screenshot 2025-09-03 at 12 43 44
Screenshot 2025-09-03 at 12 44 19


yyassi-heartex (approved):
looks way better, great work!

nass600:
/selenium test
> See '/help'

bmartel (approved):

PR #8359, opened 2025-09-05
chore: ROOT-94: read from static_build, eliminate unnecessary static file

jombooth:
Upon further investigation, it seems like keeping static + static_build in the package is a good idea. This allows for install from archive links + collectstatic after to work as expected (as happens in LSE). Compression also means that having copies of the same files in static + static_build has a negligible effect on the wheel/archive size of the package.

However, we keep a couple good changes from this diff:
- eliminate code that reads from static/ instead of static_build/
- remove an old copy of ReDoC that's sitting around in the repo


bmartel (approved):
Makes sense, keeps things predictable

PR #8365, opened 2025-09-08
fix: BROS-372: Custom tag: Fix decodeHtmlEntities

niklub:


niklub:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17549830519)

hlomzik (approved):
decoding is nice, but code change is not clear — one code is replaced with something very different. but we need to ship it I guess?

PR #8374, opened 2025-09-09
chore: BROS-362: Disable smoothing automatically when Bitmask is used

nick-skriabin:
This pull request refactors the image smoothing logic in the editor by introducing a new computed property, `smoothingEnabled`, which dynamically determines whether smoothing should be applied based on the presence of "bitmask" regions. All references to the previous `smoothing` property in rendering logic and settings initialization have been updated to use `smoothingEnabled` instead, ensuring consistent behavior across the application.

**Image smoothing logic update:**

* Added a computed property `smoothingEnabled` to the `Image` model, which disables smoothing if any annotation region is of type "bitmask"; otherwise, it uses `self.smoothing` extracted from the tag attributes. (`web/libs/editor/src/tags/object/Image/Image.js`)

**Component updates for smoothing logic:**

* Updated all relevant components (`DrawingRegion`, `ImageLayer`, and `StageContent`) to use `item.smoothingEnabled` instead of `item.smoothing` for the `imageSmoothingEnabled` prop and related logic. (`web/libs/editor/src/components/ImageView/ImageView.jsx`) [[1]](diffhunk://#diff-9e437cca91d1dff57dcf03b57c1a642bff8c6b5513277e2c28cc04e7094d582aL121-R121) [[2]](diffhunk://#diff-9e437cca91d1dff57dcf03b57c1a642bff8c6b5513277e2c28cc04e7094d582aL1330-R1331) [[3]](diffhunk://#diff-9e437cca91d1dff57dcf03b57c1a642bff8c6b5513277e2c28cc04e7094d582aL1451-R1460)
* Modified the constructor logic to initialize smoothing settings using `item.smoothingEnabled` instead of `item.smoothing`. (`web/libs/editor/src/components/ImageView/ImageView.jsx`)


yyassi-heartex (approved):
Change makes sense, but I noticed that yarn tests were failing. Reran it since it's likely flakey tests.

PR #8381, opened 2025-09-09
fix: BROS-374: BROS-376: Fix hidden dropdowns in Preview

hlomzik:
Modals in LSO has `z-index: 10000`. So in order for dropdowns to be visible we have to increase their z-index to be greater than 10000.
Fixed dropdowns:
- Taxonomy dropdown
- Autocomplete dropdown


Gondragos (approved):

hlomzik:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17588628134)

Gondragos (approved):

PR #8388, opened 2025-09-10
feat: Improved polygon drawing

cozeybozey:
Reasons for change: In this pull request I added the code that allows users to draw polygons by simply holding down the mouse button. The user can simply draw a polygon as if he/she is drawing in paint. Once the user releases the mouse button next to the starting point the polygon will close. Because drawing a polygon this way adds lots of polygon points, I added a package called simplify-geometry that simplifies the completed polygon. The simplifying maintains the same shape of the polygon, but removes lots of unnecessary points.

Screenshots: Nothing that's visible to the user changed.

Rollout strategy: The code works alongside the previous system, so I don't a specific strategy for rollout is necessary.

Testing: I tested it extensively and so did some other annotators I work with and they have not run into any issues as far as I am aware. They also much prefer the new system compared to the old system of drawing polygons. 

Risks: One risk is that the quality of the polygons might not be as good as drawing the points individually. Since this system relies on a good simplification from the simplify-geometry package. But so far that doesn't seem to be much of an issue.

Reviewer notes: Pretty much all the changes for just the improved polygon drawing are in DrawingTool.js. There's also a lot of changes that are related to touchscreen drawing. I am less sure of this code, since it was a bit more difficult to understand, so maybe the reviewer can look at that.

General notes: I don't really have any general notes.


cozeybozey:
Hello, I was wondering if any of the reviewers can take a look at this pr.

robot-ci-heartex:

robot-ci-heartex:

PR #8394, opened 2025-09-10
fix: BROS-402: Properly exit tool when switching or exiting via Esc

nick-skriabin:
This PR covers some of the edge cases when exiting a tool. Specifically:

- when switching tool mid-flight
- when being in a mixed state of selected tool without selected label


web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
```diff
@@ -253,10 +253,27 @@ export const KonvaVector = forwardRef<KonvaVectorRef, KonvaVectorProps>((props,
...
+    // Create a hash-like string for each point to detect changes efficiently
+    rawInitialPoints
+      .map((p) => {
+        if (Array.isArray(p)) {
+          return `${p[0]},${p[1]}`;
+        }
+        // For BezierPoint objects, include key properties
+        return `${p.x},${p.y},${p.isBezier ? "b" : "r"},${p.id || ""}`;
+      })
+      .join("|"),
```
hlomzik: can you also move this calculation to a memo maybe?

hlomzik: or at least maybe use more optimized and standardized approach with `JSON.stringify()`?

hlomzik (approved):

PR #8412, opened 2025-09-11
fix: UTC-269: Cant convert Export if previous conversions failed

mcanu:
Allowing new conversions if previous Export conversions failed

- Also added `TEST_ENVIRONMENT=1` to pytest.ini so we don't try to send ContextLogs to https://tele.labelstud.io [here](https://github.com/HumanSignal/label-studio/blob/a8b839fc34fe0b63bcb7ccdadfdefa11734a0f1c/label_studio/core/utils/contextlog.py#L217-L218) during testing.


hakan458 (approved):
crazy we didnt handle this before!

triklozoid (approved):

PR #8440, opened 2025-09-16
docs: BROS-351: Chat templates and Tag page updates

caitlinwheeless:
- Add 5 new template pages
- Add new template gallery page
- Adjust template gallery ordering
- Update chat tag page


docs/source/tags/chat.md
```diff
@@ -34,23 +33,45 @@ To use an LLM, you need to do two things:
...
+
+You can show options only when a message from a specific role is selected. 
+
+To do this, wrap the controls in a `<View>` block and set the `visibleWhen` and `whenRole` parameters. This ensures the controls only appear  when the annotator selects a message from that role. 
+
+For example, the annotator will only see the following choices when they select a message from the user role:
+
+```xml
+<View visibleWhen="region-selected" whenRole="user">
+  <Choices>
```
hlomzik: These are required params, especially `perRegion`, because non-per-region controls should not be placed inside `visibleWhen="region-selected"` blocks
```suggestion
  <Choices name="choices" toName="chat" perRegion="true">
```

docs/source/templates/chat_eval.md
```diff
@@ -0,0 +1,158 @@
...
+        messageroles="user,assistant"
+        editable="true" />
+</View>
+```
+
+## About this labeling configuration
+
+### Style
+
+The styling is applied through `<View>` tags, which are rendered as HTML `<div>` tags.
```
hlomzik: This phrase is not relevant here, we don't have styles in `View` tag (and it's single here)
```suggestion
```

hlomzik (approved):

PR #8442, opened 2025-09-16
fix: UTC-283: Display Order by Start Media Time depending on the labeling config rather than regions

nass600:
Detecting Start Media time mode by labeling config rather than regions. Additionally fixed, sort and sortingOrder not being up to date when changing tasks

Before:

https://github.com/user-attachments/assets/9a0f1f9a-31a5-42fc-97e8-cc7cd9b1f6ab

After:

https://github.com/user-attachments/assets/21610e89-5c30-4066-9dd9-640cef5c041c


yyassi-heartex (approved):
the change looks good, but im wondering if the change should have a FF and it seems to be less a bug and more a feature request.

nass600:
@yyassi-heartex I think the feature is very very small to add a feature flag tbh plus there is an outside of scope fix here

PR #8449, opened 2025-09-16
fix: FIT-688: Closing a temporary DM tab makes the page crash

yyassi-heartex:
This pull request addresses a race condition in the tab deletion logic within the `TabStore` in `web/libs/datamanager/src/stores/Tabs/store.js`. The main improvement is the addition of a mechanism to handle the deletion of virtual views more reliably, ensuring the correct view is selected after deletion.

[screen-recorder-tue-sep-16-2025-16-26-48.webm](https://github.com/user-attachments/assets/c984bd7c-5e41-48ea-aed0-634694d951ae)

**Tab deletion and selection improvements:**

* Added logic to handle deletion of virtual views asynchronously using `setTimeout`, which helps resolve a race condition between deleting a view and selecting a new one.
* Refactored the conditional structure in `deleteView` to ensure that autoselection and opener logic are handled before the virtual view check, improving code clarity and correctness.


web/libs/datamanager/src/stores/Tabs/store.js
```diff
@@ -174,7 +174,14 @@ export const TabStore = types
         yield getRoot(self).apiCall("deleteTab", { tabID: view.id });
       }
 
-      destroy(view);
+      if (view.virtual && newView) {
+        // this helps resolve the race condition between the view being deleted and the new view being selected
+        setTimeout(() => {
+          self.deleteView(view, { autoselect: false });
+        }, 0);
```
bmartel: Could we not just put the destroy always to be on the next tick, it is hard to understand the intent of this with a recursive function call.

web/libs/datamanager/src/stores/Tabs/store.js
```diff
@@ -174,7 +174,14 @@ export const TabStore = types
...
+      } else {
+        destroy(view);
+      }
```
bmartel:
```suggestion
      // this helps resolve the race condition between the view being deleted and the new view being selected
      setTimeout(() => {
        destroy(view);
      });
```

web/libs/datamanager/src/stores/Tabs/store.js
```diff
@@ -156,9 +156,9 @@ export const TabStore = types
     }),
 
     deleteView: flow(function* (view, { autoselect = true } = {}) {
-      if (autoselect && self.selected === view) {
-        let newView;
+      let newView;
 
+      if (autoselect && self.selected === view) {
```
bmartel:
```suggestion
      if (autoselect && self.selected === view) {
        let newView;
```

web/libs/datamanager/src/stores/Tabs/store.js
yyassi-heartex: I tried that but it was throwing this error `Cannot modify 'View[]@/viewsStore/views', the object is protected and can only be modified by using an action.`

web/libs/datamanager/src/stores/Tabs/store.js
yyassi-heartex: yes, i tried exactly that. i know the recursive call is yucky - but this was the only way i can consistently remove it with out weird mobx errors. I suspect this issue has been there for a very long time (possibly since feature was initially put in) the real issue is that Mobx should be able to handle this correctly

web/libs/datamanager/src/stores/Tabs/store.js
yyassi-heartex: since I check for newView on line 177 (outside this if statement) i needed to move the declaration to before the if. otherwise we get `ReferenceError: newView is not defined`

yyassi-heartex:
/selenium test
> See '/help'

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17795618990)
> Successfully merged:  delete mode 100644 web/libs/datamanager/src/components/Common/Form/Elements/RadioGroup/RadioGroup.scss

web/libs/datamanager/src/stores/Tabs/store.js
bmartel: The thing I don't understand is how calling the function recursively differs in how it resolves other than it is hard to read?

bmartel (approved):

PR #8456, opened 2025-09-17
Fix: Enable RQ_QUEUES configuration via environment variables (#8455) 

### LINKED ISSUE #8455: [bug] RQ_QUEUES configuration ignores environment variables and defaults to empty dict
The RQ_QUEUES configuration in `label_studio/core/settings/label_studio.py` is hardcoded to an empty dictionary {} instead of using
  environment variables, which prevents proper Redis Queue configuration for job processing.

To Reproduce
  1. Set up Label Studio with Redis Queue configuration via environment variables
  2. Try to configure Redis connection parameters through RQ_QUEUES environment variable
  3. Observe that the configuration is ignored because `RQ_QUEUES = {}` doesn't read from environment variables
  4. Redis Queue jobs fail to process or use incorrect connection settings

Expected behavior
  RQ_QUEUES should read from the RQ_QUEUES environment variable with a proper default value, similar to:

```python  
RQ_QUEUES = json.loads(os.getenv("RQ_QUEUES", "{}"))
```

This would allow environment-based configuration while providing the same default Redis Queue settings as defined in base.py.

Environment:
  - OS: Linux
  - Label Studio Version: 1.20.0
Additional context
  - The base configuration in `label_studio/core/settings/base.py:346-371` properly defines RQ_QUEUES with default Redis settings
  - However, `label_studio/core/settings/label_studio.py:31` overrides this with a hardcoded empty dict, preventing both environment variable
  configuration and Redis Queue functionality
  - Other settings in the same file properly use `get_env()` or `os.getenv()` for environment variable support
  - This affects the ability to configure RQ connection parameters for production deployments

seeker-jie:
## Reason for change
**Problem:** RQ_QUEUES was hardcoded to an empty dictionary `{}` in `label_studio/core/settings/label_studio.py`, preventing Redis Queue
configuration via environment variables and breaking background job processing.

**Solution:** Changed `RQ_QUEUES = {}` to `RQ_QUEUES = json.loads(get_env("RQ_QUEUES", "{}"))` to enable environment variable
configuration while maintaining backward compatibility.

## Rollout strategy
This change uses an existing environment variable pattern (`get_env`) and maintains backward compatibility with empty dict default. No
feature flags required - the change is safe to deploy immediately.

## Testing
**Manual verification:**
- [x] Verified RQ_QUEUES defaults to `{}` when `RQ_QUEUES` environment variable is not set
- [x] Verified RQ_QUEUES correctly parses JSON configuration from `RQ_QUEUES` environment variable
- [x] Tested that existing deployments without the env var continue working unchanged

**Unit tests:** Not required for this change as it's a simple Django settings configuration modification. Existing RQ functionality tests
will verify the configuration works correctly.

## Risks
**Low risk:**
- Backward compatible change - existing deployments unaffected
- Uses established `get_env` pattern consistent with other settings in the same file
- JSON parsing could theoretically fail with malformed input, but this follows the same pattern as other env var configurations

## Reviewer notes
- Single line change in settings file
- Follows existing code patterns in the same file (see other `get_env` usage)
- Addresses issue raised in #8455
- No additional tests needed - this is a configuration-level change

## General notes
This fix enables proper Redis Queue configuration for production deployments where Redis might be running on different hosts/ports or
require authentication, while maintaining the current behavior for existing installations.

Fixes #8455

### DIFF
--- a/label_studio/core/settings/label_studio.py
+++ b/label_studio/core/settings/label_studio.py
@@ -28,7 +28,7 @@ SESSION_COOKIE_SECURE = get_bool_env('SESSION_COOKIE_SECURE', False)
 
 SESSION_ENGINE = 'django.contrib.sessions.backends.signed_cookies'
 
-RQ_QUEUES = {}
+RQ_QUEUES = json.loads(get_env("RQ_QUEUES", "{}"))
 
 SENTRY_DSN = get_env('SENTRY_DSN', 'https://68b045ab408a4d32a910d339be8591a4@o227124.ingest.sentry.io/5820521')
 SENTRY_ENVIRONMENT = get_env('SENTRY_ENVIRONMENT', 'opensource')


farioas:
This functionality is not currently available in Label Studio, is not documented, and is not in use.

PR #8461, opened 2025-09-17
fix: BROS-418: UX improvements and fixes

nick-skriabin:
This PR addresses the following issues:

- Inability to select another region when one region is already selected AND the stage is zoomed in (coordinate system mismatch)
- Inability to continue drawing a fresh region after undo/ctrl-z
- One of the recent changes broke the multi-point transformations (move points, rotate, scale). This PR addresses this issue
- Inability to exit drawing by clicking on the last added point. This only worked for a fresh region, but not for existing ones.

https://github.com/user-attachments/assets/c49e71a2-d8e4-4dfb-a49b-5c308ff68ee9


web/libs/editor/src/tags/control/Vector.js
```diff
@@ -86,7 +87,14 @@ const Model = types
...
         hotkeys.addNamed("vector:undo", () => {
-          if (self.annotation?.selected && self.annotation.isDrawing) self.annotation.undo();
+          if (self.annotation?.selected && self.annotation.isDrawing) {
+            self.annotation.undo();
+            const tms = ToolsManager.allInstances();
+            const tool = tms.flatMap((tm) => tm.findSelectedTool())[0];
+
+            const area = tool?.commitDrawingRegion();
+            area && self.annotation.selectArea(area);
+          }
```
hlomzik: add comments describing what's going on here please. why do we search for tool? especially after undo? why do we commit region after undo? what undo is this? just not finished region as I know, right?

web/libs/editor/src/components/ImageView/ImageView.jsx
```diff
@@ -605,7 +605,7 @@ export default observer(
         return reg.isHovered?.() ?? false;
       });
 
-      if (hoveredRegion) {
+      if (hoveredRegion && !evt.defaultPrevented) {
```
hlomzik, line 600: where and how it can be prevented?

niklub:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17823935304)

web/libs/editor/src/components/ImageView/ImageView.jsx
nick-skriabin: anywhere in the call chain by using `evt.preventDefault()`

in this case it's prevented in the VectorRegion component in KonvaVector's onClick handler

web/libs/editor/src/tags/control/Vector.js
nick-skriabin: not necessary anymore, i found simpler solution

hlomzik (approved):
Approving this to unblock fixes, but feels like interactions can be improved:
- `pointSelectionHandled` can be dropped in favor of correct event propagation stop
- methods can be made static and shorter to reduce performance hit

niklub:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17918552766)

PR #8489, opened 2025-09-22
fix: Update LS deps

makseq:


makseq:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/17901465607)

PR #8513, opened 2025-09-23
feat: UTC-300: Configurable permission for managing user personal token

mcanu:
### Backend:
- New `users_token_any` controls what users can access Legacy and JWT tokens
- Removed inconsistent permission checking for JWTSettings. Now doing it how it works across the codebase, by checking permissions in the API based on request method.
- Added `permission_required` to JWT token views that were only checking if the user was authenticated, but not the permission.


mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18041315276)

web/apps/labelstudio/src/app/App.jsx
```diff
@@ -70,6 +71,7 @@ const App = ({ content }) => {
             <ConfigProvider key="config" />,
             <RoutesProvider key="rotes" />,
             <ProjectProvider key="project" />,
+            <AuthProvider key="auth" />,
```
bmartel: This should be put in the outermost layer of the providers, just below JotaiProvider, as it is possible to need access to its values from one of the other providers in the future, and it is only itself dependent on QueryClientProvider and JotaiProvider from what I can see.

bmartel (approved):

hakan458 (approved):

web/apps/labelstudio/src/app/App.jsx
nass600: Good observation, thanks

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18133474297)

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18139982858)

PR #8515, opened 2025-09-23
fix: BROS-449: Fix wrong content block height on home page (#8777)

Gondragos:
This is a try to fix dependencies of cherry-picked PR


web/libs/editor/src/Component.jsx
```diff
@@ -26,4 +26,4 @@ export class LabelStudio extends Component {
   render() {
     return this.state.initialized ? <App store={this.store} /> : null;
   }
-}
+}
```
Gondragos:
```suggestion
}
```

Gondragos:
This is no longer needed.

PR #8534, opened 2025-09-25
chore: FIT-671: Add state transitions to the current application flows

bmartel:
This pull request introduces significant improvements to the context management and finite state machine (FSM) infrastructure in the codebase, with a focus on making context propagation more robust for asynchronous jobs, enhancing FSM reliability and configurability, and introducing new state choices for annotation drafts. The most notable changes include the addition of a `CurrentContext` class for thread-local context management, improved context capture for async jobs, a more flexible state transition registration decorator, and optimizations for FSM state transitions and cache handling.

**Context Management and Async Job Improvements:**

* Introduced the `CurrentContext` class in `core/current_request.py` to provide a unified, thread-local context management interface, including methods to set/get user, organization, and request, as well as utilities for FSM feature toggling and context clearing. All context access is now routed through this class, replacing direct thread-local usage.
* Enhanced async job handling in `core/redis.py` to automatically capture and serialize relevant context (user, organization, custom job data) and inject it into job metadata, ensuring context is preserved for background tasks. [[1]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R85-R118) [[2]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R163-L146)
* Updated job execution logic to cleanly separate context management for async (Redis) and sync jobs, and improved argument handling for job submission. [[1]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55L115-R153) [[2]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R193) [[3]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R203-R204)

**FSM Infrastructure Enhancements:**

* Added a feature flag check to enable or disable FSM logic dynamically, using the new context management utilities for user/organization scoping. FSM state retrieval now respects this flag, and logging includes organization context. [[1]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R12-R13) [[2]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R58-R64) [[3]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R90-R104) [[4]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4L109-R140)
* Improved the `transition_state` method in `fsm/state_manager.py` with optimistic concurrency control using cache-based locks, prevention of redundant same-state transitions, and correct mapping of entity fields for state model records. Organization context is now consistently set and logged. [[1]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R176) [[2]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R193) [[3]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R210-R280) [[4]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R295-R301)

**FSM State and Transition Registration:**

* Extended the `register_state_transition` decorator to support trigger metadata for transitions, allowing fine-grained control over when transitions are triggered (on create, update, or specific fields).
* Introduced a new `AnnotationDraftStateChoices` enum for draft annotation lifecycle states, supporting a more granular and independent workflow for annotation drafts.

**Other Notable Improvements:**

* Corrected annotation deletion logic to count items before deletion, ensuring accurate reporting of deleted annotations.

These changes collectively improve the reliability, observability, and flexibility of both context propagation and FSM-driven workflows across the application.

---

**Most Important Changes:**

_Context Management & Async Job Context Propagation:_
- Added `CurrentContext` for unified thread-local context management, replacing direct `_thread_locals` usage and providing methods for user, organization, and request handling.
- Async job submission now auto-captures and serializes context (user, organization, job data) for use in background jobs, ensuring proper context propagation. [[1]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R85-R118) [[2]](diffhunk://#diff-2e8881789c63cd7d2fa057833d32a8429037cc6263a9b306679bb15fec088b55R163-L146)

_FSM Infrastructure & Transition Handling:_
- FSM state retrieval and transitions now respect a feature flag and include optimistic concurrency control, prevention of redundant transitions, and improved logging with organization context. [[1]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R58-R64) [[2]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R90-R104) [[3]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4L109-R140) [[4]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R176) [[5]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R193) [[6]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R210-R280) [[7]](diffhunk://#diff-757222088adfaa313e095fd7d461743038d7cf0963ca8ca443ba186cd0e6a9f4R295-R301)
- Enhanced `register_state_transition` decorator to support triggers on create, update, or specific fields, enabling more flexible FSM workflows.

_FSM State Choices:_
- Added `AnnotationDraftStateChoices` to support a full draft annotation lifecycle, independent from main annotation states.

_Annotation Deletion:_
- Fixed annotation deletion logic to count items before deletion for accurate reporting.


bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18048942826)

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18099038603)

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18192838961)

label_studio/fsm/signals.py
```diff
@@ -0,0 +1,177 @@
...
+    # Try common user field patterns
+    for field in ['created_by', 'updated_by', 'completed_by', 'user']:
+        if hasattr(instance, field):
+            user = getattr(instance, field)
+            if isinstance(user, User):
+                return user
+    return None
+
+
+@receiver(post_save)
```
makseq: Please don't use signals, it's very hard to debug them and trace the order of their calls.

makseq (dismissed):

label_studio/fsm/signals.py
bmartel: What alternatives would you suggest to keep this loosely coupled?

wesleylima: You could use a base model or mixin to override the save and update methods, but I think that's not better than signals because it increases the surface area of changes, since we have to update every model definition (which may also cause some `noop` migrations to be generated) and we'll also have to patch in FSM into every `.save()` and `.update()` override that are already in place.

I think any other alternative would create _much_ tighter coupling, like adding a "service" layer between the API views and the model, which would also be a significant departure from the normal Django app architecture.

@makseq is right, signals are normally very hard to debug, but I think FSM right now has some very good logging and the data entities in FSM itself help the entire chain of execution to be very debuggable.

wesleylima (approved):

wesleylima (approved):

label_studio/fsm/signals.py
jombooth: I think the base model idea is a pretty good one - ie, subclass `models.Model` as `HsModel` and replace all `models.Model` references with it. I don't think this would generate any migrations, and it would avoid some of the pain associated with signals, which I can attest is very real (predictable backend testing, in particular, can become a real challenge when signals are in the picture)

label_studio/fsm/signals.py
```diff
@@ -0,0 +1,177 @@
...
+def handle_model_state_transitions(sender, instance, created, **kwargs):
```
jombooth: I think it would be nice to break this up a bit, from a large if..elif..elif... wrapped in a try-except. How about:

1) implementing the error handling in a decorator
2) implementing a handler for each model as a function, decorated with said decorator
3) using a "registry" mapping each model name to its handler function
4) when we are going to do a transition, pull the function from the registry and call it. Log appropriately if no handler is registered for the model in question

This would make it easier to test each state transition handler in a decoupled fashion, as well as to gradually add new ones and increase their complexity over time

label_studio/fsm/integrations.py
```diff
@@ -0,0 +1,372 @@
...
+    Check if Label Studio Enterprise is enabled/present.
+
+    Returns True if enterprise features are available, which means
+    this core integration should be disabled to avoid conflicts.
+    """
+    from django.conf import settings
+
+    # Check if this is Enterprise edition via Django settings
+    if getattr(settings, 'VERSION_EDITION', 'Community') != 'Community':
+        return True
```
jombooth: I recommend reusing `from core.utils.common import is_community`. We've generally framed this as community vs not so far, because technically, Starter Cloud uses the LSE codebase while not being "enterprise" per se

label_studio/core/redis.py
```diff
@@ -80,6 +82,40 @@ def redis_connected():
     return redis_healthcheck()
 
 
+def _is_serializable(value: Any) -> bool:
+    """Check if a value can be serialized for job context."""
+    return isinstance(value, (str, int, float, bool, list, dict, type(None)))
```
jombooth, line 87: In what sense do we mean serializable here - ie, into what format? it strikes that including `list` and `dict` in here could be risky depending on the notion of serializability we mean

label_studio/fsm/state_manager.py
```diff
@@ -190,100 +207,155 @@ def transition_state(
...
-                    previous_state=current_state,
-                    transition_name=transition_name,
-                    triggered_by=user,
-                    context_data=context or {},
-                    reason=reason,
-                    organization_id=organization_id,
-                    **denormalized_fields,
-                )
+            try:
+                with transaction.atomic():
```
jombooth: why is transaction.atomic() needed here? reading through, it looked like only one DB operation happens here (state_model.objects.create()). Very possible I'm missing something here

label_studio/fsm/signals.py
bmartel: Done 👍

label_studio/fsm/signals.py
bmartel: Done 👍, uses explicit modelled Transition objects instead.

label_studio/fsm/integrations.py
bmartel: Refactored away from this file entirely, and the approach is using registration as the core means to extension for all aspects of FSM.

label_studio/core/redis.py
bmartel: Great point, I was trying to cut down on accidental passing of values that were not expected, like classes etc. This was a means to ensure there were only simple values being passed, not complex objects.

label_studio/fsm/state_manager.py
bmartel: That is a great question, honestly after the multitude of iterations here, I don't see a need for it. This used to be differently handled in one incarnation I had, but since that point with different refactors and changes in approach, this does not look effective anymore.

label_studio/projects/__init__.py
```diff
@@ -1,2 +1,10 @@
 """This file and its contents are licensed under the Apache License 2.0. Please see the included NOTICE for copyright information and LICENSE for a copy of the license.
 """
+
+# Register FSM transitions for Project model
+# This import ensures transitions are registered when the app loads
+try:
+    from projects import transitions  # noqa: F401
+except ImportError:
+    # FSM may not be available in all configurations
+    pass
```
jombooth: I think we should definitely log if we go down the exception branch - this could easily be missed in a refactor of the projects module for example. As in the example from LSE, let's log in any places where we handle an ImportError

jombooth (approved):
LGTM once the pytests are green!

label_studio/projects/__init__.py
bmartel: Yea, I actually was wondering about this flow, I might revisit it as it seems like it should be part of the apps.py ready callback of a given app to avoid any loading funny business and to keep it inline with the Django lifecycle.

label_studio/core/redis.py
```diff
@@ -122,28 +160,26 @@ def start_job_async_or_sync(job, *args, in_seconds=0, **kwargs):
...
     if 'redis' in kwargs:
         del kwargs['redis']
+
     job_timeout = None
     if 'job_timeout' in kwargs:
         job_timeout = kwargs['job_timeout']
         del kwargs['job_timeout']
+
     if redis:
-        # Auto-capture request_id from thread local and pass it via job meta
```
triklozoid, line 134: It looks like this change breaks passing `request_id` into the job.

bmartel: It still does it, just using CurrentContext to handle all the thread_local related aspects. I had this solution on branch before this change for request_id, so I just merged it together. request_id is treated just as other job_data is, and is part of a dictionary of values that is set onto the job meta and restored from there on the job side.

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19040255769)

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19076424295)

label_studio/core/settings/base.py
```diff
@@ -216,13 +216,14 @@
...
+    'fsm',  # MUST be before apps that register FSM transitions (projects, tasks)
     'core',
     'users',
     'organizations',
     'data_import',
     'data_export',
-    'projects',
-    'tasks',
+    'projects',  # Registers FSM transitions - needs fsm loaded first
+    'tasks',  # Registers FSM transitions - needs fsm loaded first
```
bmartel:
```suggestion
    'projects',
    'tasks', 
```

label_studio/core/settings/base.py
```diff
@@ -216,13 +216,14 @@
...
     'drf_generators',
+    'fsm',  # MUST be before apps that register FSM transitions (projects, tasks)
     'core',
     'users',
     'organizations',
     'data_import',
     'data_export',
     'projects',
-    'tasks',
+    'tasks',
```
bmartel:
```suggestion
    'tasks',
```

label_studio/io_storages/base_models.py
```diff
@@ -665,10 +698,13 @@ def sync(self):
...
                 if not self.info_set_queued():
                     return
-                sync_job = queue.enqueue(
+                # Use start_job_async_or_sync to automatically capture and restore CurrentContext
+                # This ensures user_id, organization_id, and request_id are available in the worker
+                sync_job = start_job_async_or_sync(
                     import_sync_background,
                     self.__class__,
                     self.id,
+                    queue_name='low',
```
hakan458, line 683: can get rid of `queue = django_rq.get_queue('low')` above now to avoid confusion

bmartel: It is used  by the `is_job_in_queue` function, which is why I left it.  I was thinking in another pass (another PR) we would go through and scrub all the extras and rework things accordingly to do so.

label_studio/core/models.py
```diff
@@ -9,6 +10,569 @@
 logger = logging.getLogger(__name__)
 
 
+class HsModel(models.Model):
```
makseq: why do we need to save it in /core? and not in /fsm? or does it lead to circular imports? 
I would try keeping isolate fsm maximally

label_studio/core/models.py
```diff
@@ -9,6 +10,569 @@
...
+        Args:
+            *args: Positional arguments passed to super().save()
+            **kwargs: Keyword arguments passed to super().save()
+                     Special kwarg: skip_fsm=True to bypass FSM processing
+
+        Returns:
+            Whatever super().save() returns
+        """
+        # Check for explicit FSM skip flag
+        skip_fsm = kwargs.pop('skip_fsm', False)
```
makseq: I would make it True by default to ensure we don't affect existing code

label_studio/core/redis.py
```diff
@@ -122,28 +160,26 @@ def start_job_async_or_sync(job, *args, in_seconds=0, **kwargs):
...
-            request_id = getattr(_thread_locals, 'request_id', None)
-            if request_id:
-                # Store in job meta for worker access
-                meta = kwargs.get('meta', {})
-                meta['request_id'] = request_id
-                kwargs['meta'] = meta
-        except Exception:
-            # Fail silently if no request context
-            pass
+        # Async execution with Redis - wrap job for context management
```
makseq, line 175: try/except is missed here now

label_studio/io_storages/base_models.py
```diff
@@ -484,6 +485,8 @@ def add_task(cls, project, maximum_annotations, max_inner_id, storage, link_obje
                 cancelled_annotations=cancelled_annotations,
                 inner_id=max_inner_id,
             )
+            # Save with skip_fsm flag to bypass FSM during bulk import
+            task.save(skip_fsm=True)
```
makseq, line 489: I would keep skip_fsm=True by default and avoid modifying extra code

label_studio/io_storages/base_models.py
```diff
@@ -640,6 +643,36 @@ def _scan_and_create_links(self, link_class):
...
+                    backfill_state_for_entity(task, 'task', create_record=True)
+
+                logger.info(f'Storage sync: FSM states created for {len(task_ids)} tasks')
+            except ImportError:
+                # LSE not available (OSS), skip FSM sync
+                pass
+            except Exception as e:
+                # Don't fail storage sync if FSM sync fails
+                logger.error(f'FSM sync after storage sync failed: {e}', exc_info=True)
+
```
makseq, line 651: it's better to move to /fsm/import.py or /fsm/io_storages.py or maybe functions.py or whatever; and write here just one line that calls a function from /fsm/import.py (or whatever)

label_studio/tasks/models.py
```diff
@@ -773,13 +774,65 @@ def save(self, *args, update_fields=None, **kwargs):
...
     def delete(self, *args, **kwargs):
+        # Store task and project references before deletion
+
         result = super().delete(*args, **kwargs)
         self.update_task()
         self.on_delete_update_counters()
+
         return result
 
+    def _update_task_state_after_deletion(self, task, project):
```
makseq, line 785: I would move it to /fsm/tasks.py or /fsm/functions.py

makseq (changes requested):

label_studio/core/models.py
makseq: Also, I would name it FsmHistoryStateModel, it's much more clearer

makseq (dismissed):

label_studio/io_storages/base_models.py
makseq: It seems we can replace enqueue to start_job_async_or_sync, but take into account:
https://github.com/HumanSignal/label-studio/pull/4059/files#r1191788434

It means some checks, such as if if redis_connected(), will be called twice. Maybe something else will be affected... please double-check it.

label_studio/io_storages/base_models.py
bmartel: But then we will have to make more changes than just skipping one case, we'd have to tell the 99% of cases we want to capture to do so. I don't think that makes sense. That would mean more code

mcanu (approved):

label_studio/io_storages/base_models.py
bmartel: This is Feature Flagged and the system won't do anything at all if off. When on we need this to be handled ubiquitously otherwise we will miss states. This is required to be this way.

bmartel:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19137772912)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/enterprise-upgrade-overlay/index.ts

PR #8541, opened 2025-09-26
feat: Improve table json handling and add tests

niklub:
### Reason for change

This PR enhances the `<Table>` tag's ability to render JSON data by introducing more intelligent handling for JSON arrays. Previously, JSON objects were rendered as key/value pairs. This change extends that to:
1.  **JSON array of objects**: Each object is treated as a row, and columns are dynamically derived from the keys of the first object.
2.  **JSON array of primitives/mixed values**: Rendered as a key/value table where the array index is the "Name" and the value is the "Value".
This provides a more intuitive and flexible display for various JSON data structures.

### Screenshots

N/A (no direct UI changes, only data rendering logic)

### Rollout strategy

No specific rollout strategy is required; this is a direct code improvement.

### Testing

New e2e tests have been added in `web/libs/editor/tests/e2e/tests/table.test.js` to verify:
-   Rendering of a JSON array of objects with derived columns.
-   Rendering of a JSON array of primitives as a key/value table.

### Risks

No known risks to security or performance. The changes are localized to data processing within the `Table` component.

### Reviewer notes

Please pay close attention to the `data` and `columns` getters within `web/lib


web/libs/editor/src/tags/object/Table.jsx
```diff
@@ -35,12 +35,34 @@ const Model = types
       const { type } = parseTypeAndOption(self.valuetype);
 
       if (type === "json") {
-        return Object.keys(self._value)
+        const value = self._value;
+
+        // If JSON is an array of objects, treat each object as a row for a normal table
+        if (Array.isArray(value)) {
+          if (
```
hlomzik: This check can be moved to getter to reduce code duplication and avoid further errors if we need to edit this condition.

web/libs/editor/tests/e2e/tests/table.test.js
```diff
@@ -31,3 +31,72 @@ Scenario("Check if the table is sorted", async ({ I, LabelStudio, AtTableView })
...
+        <Choice value="Correct"/>
+        <Choice value="Incorrect"/>
+    </Choices>
+</View>
+`;
+
+  const listData = {
+    rows: [
+      { a: 1, b: 2 },
+      { a: 3, b: 4, c: 5 },
```
hlomzik, line 50: According to the code column "c" will be missing. We have to either fix it or explicitly check for this column in test, otherwise we have an undefined behavior.

hlomzik (approved):
It's good in general, so I'm approving it, but I'd advice to address these two comments

hlomzik:
Also this PR should be Follow Merged, recreate it with `fb-` prefix please, otherwise we'll get inconsistent state of LSE

niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18095367622)
> Successfully merged:  create mode 100644 docs/themes/v2/source/images/releases/2_29_adjust.mp4

PR #8547, opened 2025-09-28
feat: Create proper results for CustomInterface

hlomzik:
First try


niklub:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18078010277)
> Successfully merged:  create mode 100644 label_studio/tests/sdk/test_project_annotators_api.py

web/libs/editor/src/tags/custom/Custom.jsx
```diff
@@ -126,16 +126,6 @@ const Model = types
     createRegion(value, pid) {
       const r = CustomRegionModel.create({ pid, _value: value });
       self.regions.push(r);
-      // create a simple area/result in the annotation to integrate with store
-      try {
-        const areaValue = { custominterface: value };
-        const resultValue = { custominterface: value };
-        self.annotation.createResult(areaValue, resultValue, self, self.toname);
```
hlomzik: why did you remove the main call creating result and region?

web/libs/editor/src/tags/custom/Custom.jsx
```diff
@@ -249,12 +239,18 @@ const Model = types
...
+            } catch (e) {}
           }
         } else {
-          if (store && store.events) {
-            store.events.invoke("submitAnnotation", store, annotation).catch(() => {});
+          if (store && store.events && typeof store.events.invoke === "function") {
+            try {
+              const res = store.events.invoke("submitAnnotation", store, annotation);
+              if (res && typeof res.catch === "function") res.catch(() => {});
+            } catch (e) {}
```
hlomzik: we have to remove all these calls completely! result update should not submit annotation, that's totally wrong. also it looks like it doesn't work anyway.

web/libs/editor/src/tags/custom/index.ts
```diff
@@ -11,7 +11,9 @@ Registry.addCustomTag("CustomInterface", {
   view: (CustomInterfaceComponent as unknown) as any,
   result: types.frozen(),
   resultName: "custominterface",
-  detector: (sn: any) => Boolean(sn?.value?.custominterface || sn?.custominterface),
+  // Do not detect regions by value for CustomInterface. Classification areas (global results)
+  // should deserialize as ClassificationArea, not CustomRegionModel, otherwise addResult is missing.
+  detector: () => false,
```
hlomzik: if there are problems with false detection of classification we should fix the problem, not break the detection completely.

PR #8573, opened 2025-10-02
docs: DOC-300: PDF beta page

caitlinwheeless:
New PDF page

New PDF sample for users to test with


docs/source/tags/pdf-beta.md
```diff
@@ -0,0 +1,81 @@
...
+    .pdf-actions{display:flex;justify-content:space-between}
+  </Style>
+  <Header value="Label any areas that you wish to correct" size="4"/>
+  <View className="pdf-actions">
+    <OcrLabels name="ocr" toName="pdf">
+      <Label value="Typo"/>
+      <Label value="Incorrect amount"/>
+      <Label value="Incorrect name"/>
+    </OcrLabels>
+  </View>
```
hlomzik: The idea here was to have labels on the left and per-region choices to the right. If we don't have them then better to remove the `View` and connected styles to make it simple.
But I think it would be better to add per-regions back to showcase possible use case.

caitlinwheeless: For this I'd leave it as simple as possible, but what you're talking about will make a really good template page in the docs!

hlomzik (approved):

Issue #8576, opened 2025-10-02
Web UI shows error on 206 response


EliRibble:
**Describe the bug**
I've created a task definition for processing audio transcripts. Label studio is getting audio data from a minio S3-compatible data store. My task includes a valid `s3://...` URL. When the web UI attempts to load the content it complains about a 206 response code from Minio. This is [a valid non-error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/206?utm_source=mozilla&utm_medium=devtools-netmonitor&utm_campaign=default)

**To Reproduce**
1. Deploy minio
2. Upload an audio file
3. Define a labeling interface with the following code:

```
<View>
  <Audio name="audio" value="$audio" zoom="true" hotkey="ctrl+enter" />
  <Header value="Check Transcription" />
  <TextArea name="transcription" toName="audio"
            rows="4" editable="true" maxSubmissions="1" value="$transcription"/>
</View>
```

4. Add a Cloud Storage configuration pointing to the minio bucket. Import Method should be "Files"
5. Create a task with the following body:

```
{
  audio: "s3://bucket-name/file.m4a",
  transcription: "some text that doesn't matter"
}
```

6. Load up the task.

**Expected behavior**
The audio file should load in the UI and be playable.

**Screenshots**

<img width="1581" height="515" alt="Image" src="https://github.com/user-attachments/assets/d303df72-fd8d-4720-8c4b-aadc144c6613" />

**Environment (please complete the following information):**
 - OS: MacOS
 - Label Studio Version 1.21.0

**Additional context**
The request made to minio is:
```
GET /label-studio-nidus-audio/ffda05fd-a999-4a1d-b043-0089d3241280-normalized.m4a?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=redact%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20251002T183831Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=redact HTTP/3
Host: s3.gleipnir.technology
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:143.0) Gecko/20100101 Firefox/143.0
Accept: audio/webm,audio/ogg,audio/wav,audio/*;q=0.9,application/ogg;q=0.7,video/*;q=0.6,*/*;q=0.5
Accept-Language: en-US,en;q=0.5
[...]

PR #8583, opened 2025-10-03
fix: Allow any non-error status on XHR response in audio media loader

EliRibble:
Addresses https://github.com/HumanSignal/label-studio/issues/8576

Without this a perfectly valid partial response status code like 206 will cause the entire promise to fail erroneously.

### DIFF
--- a/web/libs/editor/src/lib/AudioUltra/Media/MediaLoader.ts
+++ b/web/libs/editor/src/lib/AudioUltra/Media/MediaLoader.ts
@@ -183,7 +183,7 @@ export class MediaLoader extends Destructable {
       });
 
       xhr.addEventListener("readystatechange", () => {
-        if (xhr.readyState === 4 && xhr.status !== 200 && xhr.status !== 0) {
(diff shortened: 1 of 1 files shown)


bmartel (approved):

web/libs/editor/src/lib/AudioUltra/Media/MediaLoader.ts
```diff
@@ -183,7 +183,7 @@ export class MediaLoader extends Destructable {
       });
 
       xhr.addEventListener("readystatechange", () => {
-        if (xhr.readyState === 4 && xhr.status !== 200 && xhr.status !== 0) {
+        if (xhr.readyState === 4 && xhr.status >= 400 && xhr.status !== 0) {
```
bmartel, line 186:
```suggestion
        if (xhr.readyState === 4 && xhr.status >= 400) {
```

bmartel: It is redundant but won't hurt anything, will merge as is and follow up with this internally. Thanks @EliRibble!

PR #8630, opened 2025-10-10
fix: ROOT-212: Reduce ImportStorageLink COUNT calls during storage sync

matt-bernstein:
batch tasks_existed count during storage sync


matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18418010052)
> Successfully merged:  create mode 100644 label_studio/tasks/migrations/0058_task_precomputed_agreement.py

matt-bernstein:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18418013225)

label_studio/io_storages/base_models.py
```diff
@@ -553,9 +557,17 @@ def _scan_and_create_links(self, link_class):
...
-            if n_tasks_linked := link_class.n_tasks_linked(key, self):
-                logger.debug(f'{self.__class__.__name__} already has {n_tasks_linked} tasks linked to {key=}')
-                tasks_existed += n_tasks_linked  # update progress counter
+            if link_class.exists(key, self):
+                logger.debug(f'{self.__class__.__name__} already has tasks linked to {key=}')
+                if existed_count_flag_set:
+                    keys_for_existed_count.append(key)
+                    if len(keys_for_existed_count) >= settings.STORAGE_EXISTED_COUNT_BATCH_SIZE:
+                        tasks_existed += link_class.objects.filter(
+                            key__in=keys_for_existed_count, storaged=self.id
```
niklub: I believe key is not in the index, so this will become expensive checking for each object

label_studio/io_storages/base_models.py
```diff
@@ -539,12 +539,16 @@ def _scan_and_create_links(self, link_class):
         max_inner_id = (task.inner_id + 1) if task else 1
         validation_errors = []
 
-        # Check feature flag once for the entire sync process
+        # Check feature flags once for the entire sync process
         check_file_extension = flag_set(
             'fflag_fix_back_plt_804_check_file_extension_11072025_short', user=self.project.organization.created_by
         )
+        existed_count_flag_set = flag_set(
+            'fflag_root_212_reduce_importstoragelink_counts', user=self.project.organization.created_by
```
makseq: Please never use created_by, it will lead to weird behaviour because createdby depends on createdby active_organization. 

This week I introduced the new way - you can pass here organization directly. Check it in cursor rules about feature-flags

makseq (dismissed):

label_studio/io_storages/base_models.py
matt-bernstein: Yes, that is still true, each query is as expensive as it was before. But there are now STORAGE_EXISTED_COUNT_BATCH_SIZE = 1000x fewer queries. For storages being synced often so most keys in the storage already exist, this is the dominant cost of a sync operation. And this is the situation we are in now, with people running sync on a cronjob every 30 minutes forever (which you can check in datadog).

matt-bernstein: Adding a (storage, key) index would be a big improvement too. But adding an index to an existing large table is something Jo has warned against doing in the past, as it leads to difficult migrations. I'm not confident that doing it is a good idea without his input.

matt-bernstein: Along these lines, the next thing I will try is batching the `exists()` query as well, but that's a larger refactor, so I wanted to get this diff out first.

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18502705569)
> Successfully merged:  8 files changed, 206 insertions(+), 43 deletions(-)

label_studio/io_storages/base_models.py
```diff
@@ -553,9 +557,17 @@ def _scan_and_create_links(self, link_class):
             self.info_update_progress(last_sync_count=tasks_created, tasks_existed=tasks_existed)
 
             # skip if key has already been synced
-            if n_tasks_linked := link_class.n_tasks_linked(key, self):
-                logger.debug(f'{self.__class__.__name__} already has {n_tasks_linked} tasks linked to {key=}')
-                tasks_existed += n_tasks_linked  # update progress counter
+            if link_class.exists(key, self):
```
triklozoid: Have I understood it correctly: before we did `.count` request for every key and now we do `.exists` request for every key and also 1 request per batch for count?
So we will have even more requests than before, but we expect that `exists` is much lightweight?

matt-bernstein: Yes, exactly

triklozoid: But with the feature flag disabled we will have twice as many requests? (`.exists()` plus `.count()` for every key)
Maybe we can reimplement it in a safer way?

matt-bernstein: That's a good point, fixed it :+1:

matt-bernstein:
Results from silk on resyncing a storage with 2500 tasks:

ff ON:
548ms overall
37ms on queries
28 queries

ff OFF:
486ms overall
38ms on queries
28 queries

let me make sure the test was valid, and try other numbers of tasks

label_studio/io_storages/base_models.py
```diff
@@ -880,8 +897,8 @@ class ImportStorageLink(models.Model):
     row_index = models.IntegerField(null=True, blank=True, help_text='Parquet row index, or JSON[L] object index')
 
     @classmethod
-    def n_tasks_linked(cls, key, storage):
-        return cls.objects.filter(key=key, storage=storage.id).count()
```
jombooth, line 884: why not keep around this class method as well, but change the signature to `cls, key_set, storage`? This could be used to make the new code a bit DRYer

label_studio/io_storages/base_models.py
```diff
@@ -539,12 +539,16 @@ def _scan_and_create_links(self, link_class):
...
-            'fflag_fix_back_plt_804_check_file_extension_11072025_short', user=self.project.organization.created_by
+            'fflag_fix_back_plt_804_check_file_extension_11072025_short', organization=self.project.organization
+        )
+        existed_count_flag_set = flag_set(
+            'fflag_root_212_reduce_importstoragelink_counts', organization=self.project.organization
         )
 
         tasks_for_webhook = []
+        keys_for_existed_count = []
         for key in self.iter_keys():
```
jombooth: another optimization idea: could we implement a `batch_iter_keys` and then process multiple keys (at least for checking if already synced) in a single query?

jombooth: to reduce implementation overhead, this might not be necessary for every storage - we could just do it on the storage we're struggling with and have `.batch_iter_keys` raise NotImplemented on other storages for now

label_studio/io_storages/base_models.py
matt-bernstein: Since it's used in one place with `key=`, and another place with `keys__in=`, I didn't want to combine those, it seemed premature (even though `keys__in=[key]` is a valid use I felt it would increase rather than reduce confusion) - especially since we're going to be getting rid of one.

label_studio/io_storages/base_models.py
matt-bernstein: Could do that too, yep, but playing around with this idea in case there's still an easy win from better query planning, that'd be less risk

matt-bernstein:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18569736205)

matt-bernstein:
Silk results very similar to last time after adding batching

matt-bernstein:
As a note for the future, loading the storages page is quite slow due to the following GETs: `/storages, /storages/export, /storages/types` (most time is not spent in the DB though)

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18601468586)
> Successfully merged:  create mode 100644 label_studio/data_manager/migrations/0017_update_agreement_selected_to_nested_structure.py

label_studio/io_storages/base_models.py
```diff
@@ -539,88 +539,102 @@ def _scan_and_create_links(self, link_class):
         max_inner_id = (task.inner_id + 1) if task else 1
         validation_errors = []
 
-        # Check feature flag once for the entire sync process
+        # Check feature flags once for the entire sync process
         check_file_extension = flag_set(
-            'fflag_fix_back_plt_804_check_file_extension_11072025_short', user=self.project.organization.created_by
+            'fflag_fix_back_plt_804_check_file_extension_11072025_short', organization=self.project.organization
```
jombooth, line 544: any reason not to keep passing the `user` argument + `created_by`? I know that's the primary pattern we've used on the backend elsewhere

jombooth (approved):

label_studio/io_storages/base_models.py
matt-bernstein: Comment from Max https://github.com/HumanSignal/label-studio/pull/8630#discussion_r2422210742
new pattern introduced last week

jombooth: oh cool - yeah, that is absolutely a better design. thanks @makseq :)

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18655724511)
> Successfully merged:  create mode 100644 web/libs/ui/src/shad/components/ui/tabs.tsx

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18662394500)
> Successfully merged:  3 files changed, 21 insertions(+), 18 deletions(-)

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18683828059)
> Successfully merged:  6 files changed, 331 insertions(+), 8 deletions(-)

PR #8636, opened 2025-10-13
fix: BROS-464: Vector region selection

nick-skriabin:
This PR introduces new selection and transformation behavior to Vector tag.

Previously:
- Users weren't able to transform multiple Vectors at the same time
- UX for Vector selection was confusing -- users had to cmd or ctrl-click on the shape to select all the points and be able to transform the vector

Now:
- Users can select multiple shapes and transform them together
- Users enter the "transform" state by default when selecting Vector and can double-click on the shape to start interacting with individual points

A little demo showcasing the changes:

https://github.com/user-attachments/assets/2c7b1c25-922a-44ca-a330-8f181018eaf1


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18840357028)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/webhook.svg

web/libs/core/src/lib/utils/schema/tags.json
```diff
@@ -2763,7 +2756,7 @@
...
...>\n    <Label value=\"Nose\" />\n    <Label value=\"Mouth\" />\n  </VectorLabels>\n</View>\n```\n\n## Advanced Features\n\n### Path Breaking\nWhen a path is closed, you can break it at any segment:\n- **Alt + Click** on any segment of a closed path\n- The path breaks at that segment\n- The breaking point becomes the first element\n- The point before breaking becomes active\n\n### Skeleton Mode\n- **Purpose**: Create branching paths instead of linear sequences\n- **Behavior**: New points connect to the active point, not the last added point\n- **Use Case**: Tree structures, network diagrams, anatomical features\n\n### Bezier Curve Management\n- **Symmetric Control**: By default, control points move symmetrically\n- **Asymmetric Control**: Hold Alt while dragging to disconnect handles\n- **Control Point Visibility**: Control points are shown when editing bezier points\n\n### Multi-Selection Workflow\n1. **Build Selection**: Use Cmd/Ctrl + Click to add points\n2. **Transform**: Use transformer handles for rotation, scaling, translation\n3. **Batch Operations**: Apply transformations to all selected points\n4. **Clear**: Click outside or use programmatic methods\n\n## Props Reference",
```
hlomzik, line 2759: That's the full-scaled doc, not description. Move this text to other place (jsdoc, .md file, etc) and make description short and without markdown

web/libs/editor/src/components/ImageTransformer/ImageTransformer.jsx
```diff
@@ -214,10 +225,14 @@ export default class TransformerComponent extends Component {
           }}
           dragBoundFunc={this.dragBoundFunc}
           onDragEnd={() => {
+            // Call applyTransform for VectorRegion instances
```
hlomzik, line 228: Move this comment to the method itself. Why? Here it reads naturally and feels a good fit, you don't need to describe it using one specific tag. But the method itself could use a description saying that it's applied just to one specific tag and also with explanation why only for this one tag.

hlomzik (approved):
Too many parts with changed transformation logic, but it works well and smooth, approving!
There are issues with hover over unselected regions, but this can be solved later.
Also added small comments about comments :)

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19041793694)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/arrow-up-circle-filled.svg

PR #8637, opened 2025-10-14
fix: UTC-325: 'By Media Start Time' sort doesn't work with Video Object Tracking template

nass600:
Improved the detection to enable `Order by Media Start Time` to tuples rather than single tag types to avoid displaying the sorting option in label configs that cannot be sorted this way in the Region Panel of the Editor. Enabled tuples for now:

```typescript
const mediaStartTimeSupportedTags = [
  ["labels", "audio"],
  ["timelinelabels", "video"],
  ["timeserieslabels", "timeseries"],
];
```

### DIFF
--- a/web/libs/editor/src/components/SidePanels/OutlinerPanel/ViewControls.tsx
+++ b/web/libs/editor/src/components/SidePanels/OutlinerPanel/ViewControls.tsx
@@ -39,8 +39,14 @@ interface ViewControlsProps {
   onFilterChange: (filter: any) => void;
 }
 
+const mediaStartTimeSupportedTags = [
+  ["labels", "audio"],
+  ["timelinelabels", "video"],
+  ["timeserieslabels", "timeseries"],
+];
+
 export const ViewControls: FC<ViewControlsProps> = observer(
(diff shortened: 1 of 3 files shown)


web/libs/editor/src/components/SidePanels/OutlinerPanel/ViewControls.tsx
```diff
@@ -39,19 +39,27 @@ interface ViewControlsProps {
...
 
+const mediaStartTimeSupportedTags = [
+  ["labels", "audio"],
+  ["timelinelabels", "video"],
+  ["timeserieslabels", "timeseries"],
+];
+
 export const ViewControls: FC<ViewControlsProps> = observer(
-  ({ ordering, regions, orderingDirection, onOrderingChange, onGroupingChange, onFilterChange }) => {
+  ({ ordering, regions, orderingDirection, onOrderingChange, onGroupingChange }) => {
```
yyassi-heartex, line 48: if we're removing `onFilterChange` from here and `ViewControlsProps` isn't being used else where we should remove it from `ViewControlsProps` too

nass600: Good point, code cleaned

yyassi-heartex (approved):

PR #8638, opened 2025-10-14
fix: PLT-916: Killswitch for interaction timer

niklub:


web/apps/labelstudio/src/pages/DataManager/DataManager.jsx
```diff
@@ -33,7 +33,7 @@ const initializeDataManager = async (root, props, params) => {
     apiGateway: `${window.APP_SETTINGS.hostname}/api/dm`,
     apiVersion: 2,
     project: params.project,
-    polling: !window.APP_SETTINGS,
+    polling: window.APP_SETTINGS.polling,
```
hlomzik: if `window.APP_SETTINGS` can be missing, then we should add a null-check:
```suggestion
    polling: window.APP_SETTINGS?.polling,
```

hlomzik (approved):

PR #8656, opened 2025-10-16
fix: BROS-514: Data corruption during transformation

nick-skriabin:
This PR addresses the following two issues:

- Exception "Maximum update depth exceeded" when selecting a shape
- Data corruption during transformation due to shape deformation when bezier curves are used


niklub (approved):

niklub:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18691249234)

PR #8667, opened 2025-10-17
fix: BROS-526: Add pre-filled value for TextArea only when possible

hlomzik:
TextArea has a mechanism to add input value on annotation submit if this value was not added by user manually. This way we ensure that user didn't forget to add it. Pre-filled value will be submitted as well if user didn't touch the input.

But we should not add pre-filled value if `maxSubmissions` is reached. This fix checks for `showSubmit` which checks `maxSubmissions` under the hood.


bmartel (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18665239115)
> Successfully merged:  create mode 100644 web/libs/ui/src/shad/components/ui/tabs.tsx

PR #8678, opened 2025-10-20
docs: Add tutorials page

huguestennier:


docs/source/guide/ml_tutorials/bert_classifier.md
```diff
@@ -123,4 +123,4 @@ The following parameters are available for training:
 
 # Customization
 
-The ML backend can be customized by adding your own models and logic inside the `./bert_classifier` directory.
\ No newline at end of file
+The ML backend can be customized by adding your own models and logic inside the `./bert_classifier` directory.
```
luarmr, line 126: Why this lines are being flag as a change? Looks it happened in several parts.

luarmr (approved):
Looks good, but there's a string that shouldn't be there. I'm wondering if it's an encoding issue in the file. I you can recheck all is good there
Thanks for this!!

PR #8682, opened 2025-10-21
fix: UTC-325: By Media Start Time sort doesn't work with Video Object Tracking template

nass600:
Extended the allow list of tag combinations to trigger the sorting via `start media time` and handled the `videorectangleregion` data to sort accordingly

After:

https://github.com/user-attachments/assets/2e5bdea1-5925-4b57-8193-b87f64a0ccc8

### DIFF
--- a/web/libs/editor/src/stores/RegionStore.js
+++ b/web/libs/editor/src/stores/RegionStore.js
@@ -280,6 +280,15 @@ export default types
           }
         }
 
+        // Handle video rectangle regions - they have sequence with frames
+        if (region.type === "videorectangleregion" && region.sequence && region.sequence.length > 0) {
+          // Get the first keyframe's frame number
+          const firstKeyframe = region.sequence[0];
+          if (firstKeyframe && typeof firstKeyframe.frame === "number") {
+            return firstKeyframe.frame;
+          }
+        }
+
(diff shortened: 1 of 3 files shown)


web/libs/editor/src/stores/RegionStore.js
```diff
@@ -280,6 +280,15 @@ export default types
...
         }
 
+        // Handle video rectangle regions - they have sequence with frames
+        if (region.type === "videorectangleregion" && region.sequence && region.sequence.length > 0) {
+          // Get the first keyframe's frame number
+          const firstKeyframe = region.sequence[0];
+          if (firstKeyframe && typeof firstKeyframe.frame === "number") {
+            return firstKeyframe.frame;
+          }
+        }
```
hlomzik, line 282: I'd simplify code, now it's overcautious
```suggestion
        // Handle video rectangle regions - they have sequence with frames
        if (region.type === "videorectangleregion" && region.sequence?.[0]) {
          // Return the first keyframe's frame number
          return region.sequence[0].frame;
        }
```

hlomzik (approved):

nass600:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18685147480)

PR #8660, opened 2025-10-16
refactor: ECHO-415: BEM Migration in editor

luarmr:
# BEM Migration: Replace `BemWithSpecifiContext()` / `Block` / `Elem` with `cn()` Helper

## Summary

This PR migrates the editor codebase from the legacy `BemWithSpecifiContext()` wrapper components (`<Block/>`, `<Elem/>`) to the modern `cn()` helper function for BEM class name generation.

**Migration Status:**
- ✅ **62 files successfully migrated** with individual commits
- ⏳ **22 files remaining** (not skipped for safety - just pending migration) - make this pr review complicated
- 🚫 **0 files skipped** due to dynamic names or cross-file Elem overlaps

All remaining files are safe to migrate and will be completed in a follow-up PR.

### Conservative Safety Rules Applied

1. **No Dynamic Names**: Only migrated files where Block/Elem `name` attributes were literal strings
2. **No Cross-File Overlaps**: Verified no Elem literals appeared under multiple distinct Block literals across the repo
3. **Preserve All Behavior**: Maintained exact class string generation order and values
4. **Individual Commits**: Each file migrated in separate commit for easy review and rollback

### Transformation Patterns

**Before (Block):**
```jsx
<Block name="foo" mod={{ active }} mix={className} style={styles} {...props}>
  {children}
</Block>
```

**After (cn):**
```jsx
<div className={cn("foo").mod({ active }).mix(className).toClassName()} style={styles} {...props}>
  {children}
</div>
```

**Before (Elem):**
```jsx
<Elem name="bar" tag="span" mod={{ visible }} />
```

**After (cn):**
```jsx
<span className={cn("foo").elem("bar").mod({ visible }).toClassName()} />
```

**Before (Dynamic Tag with Block):**
```jsx
<Block name="button" tag={finalTag} mod={mods} mix={className} ref={ref} {...props}>
  {children}
</Block>
```

**After (cn with createElement):**
```jsx
import { createElement } from "react";

createElement(
  finalTag,
  {
    ref,
    ...props,
    className: cn("button").mod(mods).mix(className).toClassName(),
  },
  children,
)
```

**Before (Elem with Component Tag):**
```jsx
<Elem tag={Userpic} name="item" mod={{ active }} user={user} />
```

**After (cn with Component):**
```jsx
<Userpic className={cn("foo").elem("item").mod({ active }).toClassName()} user={user} />
```

### Common Components (15)
- ✅ Space.tsx
- ✅ Button.tsx
[...]

PR #8688, opened 2025-10-22
refactor: ECHO-415: Revert "BEM Migration in editor"

luarmr:
Reverts HumanSignal/label-studio#8660

Just in case


yyassi-heartex (approved):
hopefully we won't need it

PR #8691, opened 2025-10-22
refactor: ROOT-221: remove fflag_feat_all_leap_1430_per_field_comments_100924_short FF

luarmr:
Removes the `fflag_feat_all_leap_1430_per_field_comments_100924_short` feature flag and makes the new per-field comments system the default implementation.

Also modify the code since an e2e test started failing. We are running yarn e2e with this feature flag set to false 🤦
We should update the feature flag file to the latest values.

### Frontend (Open Source)
- **Comments.tsx** - Always use new `Comment/` components instead of conditional loading
- **OutlinerTree.tsx** - Always enable comment linking mode for regions
- **App.jsx** - Always render comments overlay
- **Comment.js** - Make `regionRef` and `classifications` always available
- **NodesConnector.js** - Always watch `containerWidth`, `containerHeight`, and `canvasSize`
- **feature-flags.ts** - Remove FF_PER_FIELD_COMMENTS export

### Cleanup
- Delete `OldComment/` directory (5 files) - legacy comment system no longer needed

## Impact
- Per-field comments with region linking is now the default and only implementation
- Reduced code complexity by removing dual system support
- All users now have access to enhanced comment features (classifications, region linking, overlay)

# E2E start failing:

Issue Explanation: Test Failure After Removing FF_PER_FIELD_COMMENTS

**What Changed**
When removing the FF_PER_FIELD_COMMENTS feature flag, we updated OutlinerTree.tsx to always enable comment linking functionality:

**Why the Test Failed**
The condition annotation.isLinkingMode returns true for any active linking mode:
LINK_COMMENT_MODE - for linking comments to regions (what we wanted)
CREATE_RELATION_MODE - for creating relations between regions (should be separate)

**The E2E test scenario**:
Test clicks "Create Relation" button → activates CREATE_RELATION_MODE
Test clicks a region to select it
Our code intercepted the click thinking it was for comment linking
Region selection was prevented → background color didn't change → test failed

**The Fix**
Check specifically for comment linking mode instead of any linking mode:
This ensures we only intercept region clicks when explicitly in comment linking mode, allowing relation creation and normal region selection to work as expected.

**TL;DR:** The feature flag previously masked a bug where we were checking isLinkingMode (any mode) instead of specifically checking for LINK_COMMENT_MODE. Once the flag was removed and the code always executed, it broke relation mode functionality by intercepting all linking mode clicks.

**Sentry errors?**

No, this specific bug wouldn't show up in Sentry because:

Why It's Not Visible in Sentry
No JavaScript Errors - The code executed successfully without throwing exceptions
Logic Bug, Not Runtime Error - The wrong code path was taken, but nothing crashed
Silent Behavioral Issue - Users would just experience unexpected behavior (can't select regions during relation creation)

What Users Would Experience
Instead of errors in Sentry, users might have reported:
"I clicked 'Create Relation', but when I click a region, nothing happens"
"Can't select regions after clicking the relation button"
"Region selection is broken in certain scenarios"


jombooth (approved):
This is great, really excited that we can remove it!

web/libs/editor/tests/e2e/tests/rich-text/rich-text-regions-displaying.test.js
```diff
@@ -209,8 +209,11 @@ Scenario("Displaying selected and highlighted regions", async ({ I, LabelStudio,
     }
   }
 
-  I.say("Unselect region");
-  I.pressKey("u");
+  I.say("Exit relation mode and unselect region");
```
hlomzik:
```suggestion
  I.say("Exit relation mode");
```

hlomzik (approved):

PR #8699, opened 2025-10-23
fix: FIT-851: The Assign Annotators/ Reviewers modal does NOT always open from Task Actions

yyassi-heartex:
This pull request makes a small change to the `AppStore` in the data manager. The change removes the use of `yield` when calling `actionCallback`, making the function call synchronous instead of asynchronous.

this cleans up the console from an error relating to yielding a function that returns an object


web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -751,7 +751,7 @@ export const AppStore = types
       }
 
       if (actionCallback instanceof Function) {
-        const result = yield actionCallback(actionParams, view);
+        const result = actionCallback(actionParams, view);
```
bmartel, line 749: Is this function sometimes a generator function, other times not? Or is it always a sync standard function?

bmartel: Because calling it without the yield if it ends up a generator function will cause other problems

bmartel: Could just check if its a generator function then yield if it is:

```ts
if (actionCallback.constructor.name === 'GeneratorFunction')
```

yyassi-heartex: it seems to be always a standard function. I went through all the options in the action menu and everything just pops up a modal  and seems to work. removing the yield removes the console mobx error - but I was unable to replicate the modal failing to open as reported int he ticket (which appears to only happen in selenium so likely related some other failure).

bmartel (approved):

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/18811457147)
> Successfully merged:  create mode 100644 web/libs/ui/src/assets/icons/webhook.svg

PR #8710, opened 2025-10-24
feat: UTC-349: Show tooltip for tasks over agreement selected threshold

hakan458:
Displays an info icon with tooltip for any tasks that have more annotations than the threshold set for Agreement (Selected) 
For these tasks we skip calculation and now will display helpful information for the user


web/libs/datamanager/src/components/CellViews/AgreementSelected.jsx
```diff
@@ -25,11 +27,20 @@ const formatNumber = (num) => {
 export const AgreementSelected = (cell) => {
   const { value, original: task } = cell;
 
-  const score = (
+  const threshold = window.APP_SETTINGS?.agreement_selected_threshold;
```
yyassi-heartex, line 30: I checked this in console in develop - it evaluates to undefined - how is this getting passed to the front?

hakan458: Its only set in LSE since thats where we use agreement selected and have this option in the backend, to send to the frontend
https://github.com/HumanSignal/label-studio-enterprise/pull/9068/files#diff-ffe366b713895be945d47dc72dfe05a2386164869c760bdd0def0d4ef9d6f8fcR72

yyassi-heartex: oh, i missed that! great!

yyassi-heartex (approved):

PR #8726, opened 2025-10-28
fix: UTC-358: Blank page appears instead of 404 page does NOT appear upon trying to navigate to an inexistent project

nass600:
Restoring redirection handling when receiving a 404 from the api


yyassi-heartex (approved):

bmartel (approved):

hlomzik (changes requested):
I'll block PR for further investigation. We had similar behavior before and it can be VERY confusing, so we should proceed with caution

PR #8733, opened 2025-10-29
docs: fix report bug link

luarmr:
we were not using the ling in our community templates to report a bug


bmartel (approved):

luarmr:
Broken link action is not related. and it is failing randomly. Let's look at this in the cooldown

PR #8746, opened 2025-10-31
fix: UTC-358: Blank page appears instead of 404 page does NOT appear upon trying to navigate to an inexistent project

nass600:
Adding missing icons


ricardoantoniocm (approved):
Approving, on the condition that icons that the suffix is changed to `-circle-filled` instead of just `-filled`.

PR #8762, opened 2025-11-05
fix: FIT-786: Making DM filters support multiple projects (optionally)

wesleylima:

### DIFF
--- a/label_studio/data_manager/prepare_params.py
+++ b/label_studio/data_manager/prepare_params.py
@@ -37,12 +37,24 @@ class SelectedItems(BaseModel):
 
 
 class PrepareParams(BaseModel):
-    project: int
+    project: Union[int, List[int]]  # Support both single project and multiple projects
     ordering: List[str] = []
     selectedItems: Optional[SelectedItems] = None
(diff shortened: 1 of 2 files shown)


label_studio/data_manager/prepare_params.py
```diff
@@ -37,13 +37,25 @@ class SelectedItems(BaseModel):
...
 
+    @property
+    def projects(self) -> List[int]:
+        """Get project IDs as a list, whether single or multiple were provided."""
+        if isinstance(self.project, list):
+            return self.project
+        return [self.project]
+
+    @property
+    def is_multi_project(self) -> bool:
```
yyassi-heartex, line 55: should this also be checking for the FF? current stuff shouldn't be getting a list at all right?

yyassi-heartex (approved):

label_studio/data_manager/prepare_params.py
wesleylima: it's more about how the `projects` attribute in this class is set. Without the ff, nothing is setting set.project to a list so we don't have to re-check the ff here

PR #8798, opened 2025-11-10
docs: Page for org permissions

caitlinwheeless:
There is a new page for Owners to customize permissions.

Also reorganized some pages

### DIFF
--- a/docs/source/guide/admin_permissions.md
+++ b/docs/source/guide/admin_permissions.md
@@ -44,6 +44,6 @@ Permissions that can be configured include the following:
 !!! note
     Permissions for Annotators, Reviewers, and Managers are already restricted to projects that they created (if a Manager) or have explicitly been granted access to. 
 
-!!! tip
+!!! info tip
     You can click **Reset to Defaults** to return all permissions to their default value. 
 
--- a/docs/source/guide/admin_permissions.md
+++ b/docs/source/guide/admin_permissions.md
@@ -1,0 +1,50 @@
+---
+title: Customize organization permissions
+short: Permissions
+tier: enterprise
+type: guide
+hide_menu: false
+order: 0
+order_enterprise: 362
+meta_title: Customize organization permissions
+section: "Manage Your Organization"
+parent_enterprise: "admin_settings"
+date: 2025-02-18 12:03:59
+---
+
+!!! note
+    This page is only visible to users in the Owner role. 
+
+You can use this page to customize certain permissions for roles. For information about Label Studio Enterprise roles and default permissions, see [User roles and permissions](admin_roles).
+
+Note the following:
+
+* Permissions for the Owner role are not configurable. 
+* You can use the options on this page to restrict permissions, but you cannot extend permissions. 
+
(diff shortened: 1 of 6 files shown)


docs/source/guide/admin_permissions.md
```diff
@@ -0,0 +1,50 @@
...
+
+!!! note
+    This page is only visible to users in the Owner role. 
+
+You can use this page to customize certain permissions for roles. For information about Label Studio Enterprise roles and default permissions, see [User roles and permissions](admin_roles).
+
+Note the following:
+
+* Permissions for the Owner role are not configurable. 
+* You can use the options on this page to restrict permissions, but you cannot extend permissions.
```
mcanu: I don't think this is valid, there might be permissions that are restricted to Admin and Manager, and through this page extend it to Annotator and Reviewer.

eurica: agreed, but I just don't love the word "extend". It's true if it means "create new permissions" but in the english sense of the word you can extend the ability to people.

Proposal: You can use the options on this page to restrict or expand which roles have permissions, but not create new roles.

docs/source/guide/admin_permissions.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Permissions for the Owner role are not configurable. 
+* You can use the options on this page to restrict permissions, but you cannot extend permissions. 
+
+    For example, if an action is typically available to Managers, Admins, and Owners, you can restrict it to Admins and Owners, but you cannot *extend* it to Annotators or Reviewers. 
+
+* Restrictions will also apply to the API. For example, if you restrict Managers' ability to configure cloud storage, they will not be able to complete those actions through the UI or the API. 
+
+
+Permissions that can be restricted include the following:
```
hakan458: restricted -> configured since what Marcel said above is correct

docs/source/guide/admin_permissions.md
```diff
@@ -0,0 +1,50 @@
...
+
+| Permission | Default Roles |
```
eurica, line 29: Love calling out the defaults! Can we mention you can "Reset to Defaults" which returns every perm to these?

hakan458, line 29: Is there any way we can build this list dynamically, based on a list coming from the code itself? I know we will forget to update this when adding new configurable permissions. I know we have this problem in other places, wondering if you have any ideas @caitlinwheeless

eurica: (I agree *we* will forget, but I doubt Caitlin will :)

docs/source/guide/admin_permissions.md
nass600: Technically yes, we can extend permissions once they are plugged in in this feature (done by tech) if the permissions before were more restrictive but I would not mention this to the user as it will confuse them. I would put something like: "You can use the options on this page to have a more fined-grained control on which roles can execute certain actions"

docs/source/guide/admin_permissions.md
caitlinwheeless: This will be in version 2!

caitlinwheeless: Done

mcanu (approved):

PR #8806, opened 2025-11-12
fix: FIT-986: Visual regression affecting Filters in Datamanager

yyassi-heartex:
This pull request makes targeted improvements to the UI component logic in `select.tsx` and cleans up imports in `Custom.jsx`. The most significant change is a refinement to how the Select dropdown calculates the number of visible items, ensuring more accurate rendering when options are grouped or filtered.

before:
Screenshot 2025-11-12 at 3 42 18 PM

after:
Screenshot 2025-11-12 at 3 42 40 PM

**UI component improvements:**

* Improved the logic for calculating the number of visible items in the Select dropdown by accurately counting items from `flatOptions` and handling both searching and non-searching states, including support for option groups.

**Code cleanup:**

* Removed unused imports from `Custom.jsx` to streamline the file.

### DIFF
--- a/web/libs/editor/src/tags/Custom.jsx
+++ b/web/libs/editor/src/tags/Custom.jsx
@@ -1,6 +1,3 @@
-// Function-based CustomInterface available only in Label Studio Enterprise
-import React from "react";
-
 import { types } from "mobx-state-tree";
 import { observer } from "mobx-react";
 import { EnterpriseBadge } from "@humansignal/ui";
--- a/web/libs/ui/src/lib/select/select.tsx
+++ b/web/libs/ui/src/lib/select/select.tsx
@@ -394,23 +394,8 @@ export const Select = forwardRef(
                         ref: infiniteLoaderRef,
                       }: { onItemsRendered: (params: any) => void; ref: any }) => {
                         // Calculate height based on actual item count from flatOptions
-                        // When searching, _options is already flattened; when not searching, we need to count children
-                        const actualItemCount =
-                          searchable && query.trim()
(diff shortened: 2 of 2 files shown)


web/libs/editor/src/tags/Custom.jsx
bmartel, line 1: Is this intended to remove? I would leave this for now, just in case. Since it is not part of this changeset.

yyassi-heartex: it was cleaned up as part of linting - i see the changed that made it redundant was merged 2 hours ago - not sure how linter check didnt fail

web/libs/ui/src/lib/select/select.tsx
```diff
@@ -393,8 +393,24 @@ export const Select = forwardRef(
                         onItemsRendered,
                         ref: infiniteLoaderRef,
                       }: { onItemsRendered: (params: any) => void; ref: any }) => {
-                        // Calculate height based on actual item count, max 5 items
-                        const actualItemCount = renderedOptions.length;
+                        // Calculate height based on actual item count from flatOptions
+                        // When searching, _options is already flattened; when not searching, we need to count children
+                        const actualItemCount =
```
bmartel: Seems a bit expensive of a check just to ensure the maxVisibleItems is between 1 and 5. I'm not 100% sure I follow the logic though.

yyassi-heartex: great catch! I cleaned it up - i blame claude for overthinking it

bmartel (approved):

PR #8810, opened 2025-11-12
fix: FIT-988: Add BemWithSpecificContext back to bem.tsx

ricardoantoniocm:
This pull request makes a minor update to the `web/apps/labelstudio/src/utils/bem.tsx` file, enhancing the module's export capabilities.

* Now exports both `cnb` (as `cn`) and `BemWithSpecificContext` from `@humansignal/core/lib/utils/bem`, allowing consumers to use additional BEM utility functions.


luarmr:
Gracias!!!

luarmr (approved):

PR #8819, opened 2025-11-13
chore: Move Audio File Decoder to humansignal dep

bmartel:
Moving the Audio File Decoder dependency to point to humansignal namespace.


web/package.json
```diff
@@ -254,11 +247,5 @@
     "on-headers": "1.1.0",
     "form-data": "4.0.4",
     "stylus": "0.0.1-security"
```
yyassi-heartex, line 249: what you're at it, can you remove the stylus dependency?

yyassi-heartex (approved):

PR #8821, opened 2025-11-13
feat: BROS-594: Add Distribution row to Task Summary

hlomzik:
## Aggregation/Distribution

It displays aggregations for choices, taxonomies, etc.:
- For number and rating it shows average number
- For labels and taxonomies it shows counts for every option
- For choices it shows distribution of every choice, including annotations where this choice was skipped

## Styles

A lot of styles were fixed in Task Summary: badges, table design, zebra rows, sticky column, usernames, etc.

## Autoselect of Summary tab

When you switch between tasks you'll end up in the same pane you've been before — either annotation view, Summary tab or Compare tab. It makes it easier to review multiple tasks.

We already had a memory about a tab in View All, but now we also remember was it annotation view or View All. Technically it's tricky because we exit View All in a lot of places where we select annotation. And we select all annotations (or one with FF_SIMPLE_INIT on) on LSF init, so it always resets the view. So now we only exit View All if we are already in View All and do an interaction with UI.

Screenshot 2025-11-13 at 18 02 44


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19347995250)
> Successfully merged:  create mode 100644 label_studio/fsm/urls.py

hlomzik:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19348703150)
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19348703150)

web/libs/editor/src/components/TaskSummary/Chip.tsx
bmartel, line 1: Do we not have a similar Badge component in the UI lib which could be used here?

web/libs/editor/src/components/TaskSummary/utils.ts
bmartel, line 1: Similarly it seems like we have many ways we handle color cycling for labels, it might be worthwhile consolidating the logic to a shared core util so everywhere we ever need this label cycling, it is the same.

hlomzik: Good idea in general, but we have different logic for labels. Plus it uses outdated libraby we don't control, so it would be a nice effort to make all this controllable and consolidated, but definitely not in this PR.

web/libs/editor/src/components/TaskSummary/Chip.tsx
hlomzik: This Chip has its own display logic. We might use `Badge` as display component inside `Chip`, but styles are different, we work on this later with @ricardoantoniocm

niklub (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19432922680)
> Successfully merged:  create mode 100644 web/libs/datamanager/src/components/Filters/types/TaskStateFilter.jsx

PR #8837, opened 2025-11-18
docs: Adding metadata to the tutorials

ReDeiPirati:
This PR adds metadata to all the new engineering tutorials, and it's parent page. It also updates Marcel's notebook by cleaning a code cell output that wasn't useful


luarmr (approved):

luarmr:
Nice, 
Let's agree on the metadata we will need and add it to the original repo in a manifest or similar.

PR #8839, opened 2025-11-18
chore: FIT-1008: Update glob package to address security advisory

bmartel:
Updated glob package to latest secure version per security advisory.

Changes:
- Added glob resolution (>=11.1.0) to package.json
- Updated yarn.lock to enforce secure version across dependencies
- Upgraded glob to v11.1.0

This update addresses a security advisory for the glob package. All transitive dependencies have been updated accordingly.


yyassi-heartex (approved):

bmartel:
This change would require far more updates to take place, and the vulnerable aspect is not the library, but the cli package (which we do not use CLI explicitly or implicitly).

PR #8841, opened 2025-11-18
feat: FIT-974: [FSM API] Manual Transition Endpoint

matt-bernstein:
This PR was automatically generated via Follow Merge. [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/19518304411)
- https://github.com/HumanSignal/label-studio-sdk/pull/676

Add api to transition entities between states


matt-bernstein:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19516764422)

label_studio/fsm/api.py
```diff
@@ -99,3 +106,117 @@ def list(self, request, *args, **kwargs):
...
+            raise ValidationError({'detail': str(e)})
+        except TransitionValidationError as e:
+            # Explicit validation failure
+            raise ValidationError({'detail': str(e), **(e.context or {})})
+        except ValueError as e:
+            # Unknown transition or similar issues
+            raise ValidationError({'detail': str(e)})
+        except Exception as e:
+            # Unexpected failure
+            raise APIException(detail='Transition execution failed') from e
```
mcanu: I wouldn't catch these, these look like 5XX errors that we'll return to the user and catch them in Sentry. The detail of the exception isn't showed to the user

label_studio/fsm/api.py
```diff
@@ -99,3 +106,117 @@ def list(self, request, *args, **kwargs):
...
+            self.check_object_permissions(self.request, entity)
+        except PermissionDenied as e:
+            # Return 404 instead of 403 to avoid leaking entity existence
+            raise NotFound() from e
+        return entity
+
+    def post(self, request, *args, **kwargs):
+        entity_name = kwargs['entity_name']
+        if entity_name not in state_model_registry.get_all_models():
+            raise NotFound()
```
mcanu: Can we extract this to a mixin that is shared by the history endpoint as well?

label_studio/fsm/api.py
```diff
@@ -99,3 +106,117 @@ def list(self, request, *args, **kwargs):
...
+                transition_data=transition_data,
+                user=request.user,
+                organization_id=getattr(request.user, 'active_organization_id', None),
+            )
+        except PydanticValidationError as e:
+            # Pydantic schema validation errors from transition instantiation
+            raise ValidationError({'detail': str(e)})
+        except TransitionValidationError as e:
+            # Explicit validation failure
+            raise ValidationError({'detail': str(e), **(e.context or {})})
```
mcanu: q: What do this look like? Just a warning that we don't show unwanted information to the user

label_studio/fsm/api.py
```diff
@@ -99,3 +106,117 @@ def list(self, request, *args, **kwargs):
...
+                'state_record': None,
+            }
+        else:
+            state_record_data = StateModelSerializer(state_record).data
+            response_payload = {
+                'success': True,
+                'new_state': state_record.state,
+                'state_record': state_record_data,
+            }
+        return Response(response_payload, status=status.HTTP_200_OK)
```
mcanu: Lets use the response serializer her

label_studio/fsm/api.py
matt-bernstein: Here's the only time TransitionValidationError is used with context outside of a test:
https://github.com/HumanSignal/label-studio/blob/9c9bf2024783bfe1560d27d3006f53a7ba8c9804/label_studio/fsm/transitions.py#L291-L297

It's definitely possible that this could leak unwanted info in the future, but as of now, it's very rare to use `context` in the first place, and when it's used it's for obvious information. Here's an example from a test:
```
raise TransitionValidationError('Invalid assignee ID', {'assignee_id': self.assignee_id})
```

I'm torn on this one, it could make debugging much easier to have, but there is no guardrail to stop us from writing something sensitive there even though it seems unlikely

matt-bernstein:
/git merge develop
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19538594387)
> Successfully merged:  4 files changed, 20 insertions(+)

label_studio/fsm/tests/test_transition_api.py
mcanu, line 1: I would move this to `fsm/tests/test_api.py`

mcanu (approved):

label_studio/fsm/api.py
mcanu: I would not include the context, for now at least. We can log it in a warning log there to help us for troubleshooting. But I agree we don't know what can be in the context, and there might be some internal stuff we don't want to show.

mcanu: I'd defer to any public info worth showing to the user for the Exception message

bmartel (approved):
Looks great, thanks @matt-bernstein 🙌

PR #8848, opened 2025-11-19
fix: BROS-592: Correctly handle `whenRole` for required controls

hlomzik:
Basically it's just a duplication of the similar check in `Visibility` mixin.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19509448404)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/state-chip/state-chip.tsx

web/libs/editor/src/mixins/Required.js
```diff
@@ -13,6 +13,8 @@ const RequiredMixin = types
 
     return {
       validate() {
+        const whenRoles = self.whenrole ? self.whenrole.split(",") : null;
```
nick-skriabin: might worth doing `toLowerCase` here and below when you check `includes` so that user won't mess it up

hlomzik: we have different visibility params like `whenChoiceValue` and `whenLabelValue` and they all are case-sensitive, so I'd leave it like this for consistency. better to keep these values strict to avoid disambiguity.

web/libs/editor/src/mixins/Required.js
```diff
@@ -32,6 +34,8 @@ const RequiredMixin = types
 
                 if (label && label !== self.whentagname) continue;
               }
+
+              if (whenRoles && !whenRoles.includes(reg.chatmessage?.role)) continue;
```
bmartel:
```
              if (whenRoles && !whenRoles.includes(reg.chatmessage?.role)) continue;
```

This feels oddly specific to include `chatmessage` in a base mixin

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19772352842)

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19809168509)
> Successfully merged:  4 files changed, 35 insertions(+), 10 deletions(-)

hlomzik:
@bmartel @nick-skriabin thanks for feedback! I changed it to less specific `role`, without `chatmessage`, I don't think we can do better given that param is called `whenRole`.
Also I improved code, fixed issues and added test for LSE.

nick-skriabin (approved):

bmartel (approved):

PR #8853, opened 2025-11-20
fix: BROS-482: Hiding and locking

nick-skriabin:
This PR fixes inability to lock or hide Vector regions.

- Original `disabled` prop was renamed to `selected` and inverted logic to clearly reflect what it does
- Introduced new `disabled` prop that will actually disable all shape interactions
- Fixed region hiding


web/libs/editor/src/components/KonvaVector/components/VectorPoints.tsx
```diff
@@ -82,7 +84,7 @@ export const VectorPoints: React.FC<VectorPointsProps> = ({
         return (
           <>
             {/* White outline ring for selected points - rendered outside the colored stroke */}
-            {!disabled && isSelected && (
+            {selected && isSelected && (
```
hlomzik, line 87: `selected` AND `isSelected`? feels a little awkward

hlomzik: \+ `isExplicitlySelected` + `isMultiSelection` + `selectedPoints`
can it all be simplified? feels that we went too deep

web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
```diff
@@ -2154,11 +2172,11 @@ export const KonvaVector = forwardRef<KonvaVectorRef, KonvaVectorProps>((props,
         pixelSnapping,
         isDraggingNewBezier,
         ghostPointDragInfo,
-        disabled,
+        selected,
         isShiftKeyHeld: refShiftState,
       } = currentValuesRef.current;
 
-      if (disabled || isDragging.current || isDraggingNewBezier || ghostPointDragInfo?.isDragging) {
+      if (disabled || !selected || isDragging.current || isDraggingNewBezier || ghostPointDragInfo?.isDragging) {
```
hlomzik, line 2182: you removed `disabled` from props but still use it in this condition. is it correct?

web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
```diff
@@ -2639,8 +2674,8 @@ export const KonvaVector = forwardRef<KonvaVectorRef, KonvaVectorProps>((props,
       }
 
       // Handle shape dragging first (if active, allow dragging to continue even outside bounds)
-      // Skip individual shape dragging if in multi-region mode (ImageTransformer handles it)
-      if (isDraggingShape && shapeDragStartPos.current && !isMultiRegionSelected) {
+      // Skip individual shape dragging if disabled, in multi-region mode (ImageTransformer handles it)
```
hlomzik: you missed "or"?

```suggestion
      // Skip individual shape dragging if disabled or in multi-region mode (ImageTransformer handles it)
```

web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
nick-skriabin: i didn't remove it. it was renamed and re-introduced

original `disabled` was renamed to `selected` to better reflect what it does. newly added `disabled` is specifically for disabling user interactions

hlomzik: you removed it couple of lines earlier in var destructuring and replaced with `selected`. so apparently it's used from some code above, which makes it not obvious plus I believe it's missing from `useEffect` dependencies.
I'd suggest to return it back to this function.

hlomzik (approved):
in general code looks like a good fit (after fixing issue with missing `disabled` var)

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19599002345)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/dropdown/index.ts

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19631353939)
> Successfully merged:  35 files changed, 45 insertions(+), 45 deletions(-)

PR #8854, opened 2025-11-20
chore: FIT-1025: Update frontend monorepo NX dependency to 21.4.0

bmartel:
Updating the monorepo dependencies to as far out as we can go in one update. This is so we can keep pace with security patches and ensure updates are even possible, as we don't want to stale out and have to make a massive migratory effort.

This removes Stylus entirely from both Monorepos, among other updates to test harness and correcting tests to follow best practices of React 18+ with latest Jest.


yyassi-heartex (approved):

web/yarn.lock
```diff
@@ -19024,8 +20466,9 @@ typescript-plugin-css-modules@^5.1.0:
     reserved-words "^0.1.2"
     sass "^1.70.0"
     source-map-js "^1.0.2"
-    stylus "^0.62.0"
     tsconfig-paths "^4.2.0"
+  optionalDependencies:
+    stylus "^0.62.0"
```
luarmr, line 20460: oh man they move this to optional, great

luarmr, line 20460: Maybe we should target it to avoid it?
{
  "resolutions": {

    "typescript-plugin-css-modules/stylus": false
  }
}

bmartel: Lets try it 👍

bmartel: Didn't change anything unfortunately, still has the optionalDependency.

luarmr: :( looks this work in yarn 2 :(

luarmr: @bmartel I found this trick:

  "resolutions": {
    "stylus": "npm:empty-npm-package@1.0.0"
  }

bmartel:
Single test that is failing is flakey before this branch, and these changes don't affect that test harness whatsoever as it has its own set of dependencies isolated from those updated in the monorepo. Going to bypass this as rerunning is a waste of CI time.

PR #8859, opened 2025-11-21
fix: BROS-636: Regression with storage syncs and RQ workers

makseq:
If you run a sync job and if you try to click Sync storage button again while the sync job is already running, clicking Sync storage caused a 500 with `ValueError: signal only works in main thread of the main interpreter`. This happened because is_job_on_worker() used StartedJobRegistry.get_job_ids(), which internally runs registry cleanup using RQ’s “death penalty” (SIGALRM). When called from a Django request thread (not the main interpreter thread), that signal setup fails and bubbles up as a runtime error.

The fix changes is_job_on_worker() to perform a read-only membership check directly in Redis using ZSCORE on the registry’s sorted set key, instead of calling get_job_ids(). Behavior is unchanged from the caller’s perspective (we still answer “is this job id currently started?”), but we no longer trigger any signal-based timeout logic in request threads, so storage sync works reliably without runtime errors.

image


triklozoid (approved):

makseq:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19570221108)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/dropdown/index.ts

PR #8865, opened 2025-11-21
chore: FIT-1040: remove boto2 dependency

matt-bernstein:
unused


luarmr (approved):
wow surprise is not used at all

PR #8873, opened 2025-11-24
chore: FIT-1040: remove boto2 dependency

matt-bernstein:
cherry-pick


luarmr (approved):

luarmr:
For this cherry pick, should we run poetry again or get the latest from develop? I'm wondering what the best practice is for this.

matt-bernstein:
I ran poetry again and verified that the diff against lse-release is the same as the diff against develop, except for the hash. So it doesn't matter in this case

PR #8889, opened 2025-11-25
fix: BROS-646: More undo issues

nick-skriabin:
This PR addresses the issues where:

- User is unable to draw a new region after completely undoing the previous one. Scenario is the following:
  - Add 3 points
  - Undo 3 times so that the first point and the region are removed
  - Select label
  - Try drawing new region
  - Observe that points are not added and the region is not created


hlomzik (approved):

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19711403530)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/dropdown/dropdown.stories.tsx

PR #8898, opened 2025-11-26
fix: BROS-641: Select annotation from username in Compare

hlomzik:
We have a regression: username/userpic still have pointer cursor in Compare table in View All, but click on them does nothing.

The reason is the change in `selectAnnotation()` method — now it doesn't close View All by default, we need to pass an option `exitViewAll: true` to it.

Also fixed "Create New Annotation" and "Duplicate Annotation" actions.


yyassi-heartex (approved):

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19750555591)

nick-skriabin (approved):

PR #8904, opened 2025-11-27
fix: FIT-1059: Not sure 'Created' is the right name for the annotation state, perhaps 'Initial' could make more sense, take a look at the screenshot

yyassi-heartex:
This pull request introduces improvements to how FSM (Finite State Machine) states are handled for tasks that are bulk-created, ensuring their state is properly initialized even when the usual model save hooks are bypassed. Additionally, it updates the label for the initial task state in the UI for clarity.

before:
image

after:
Screenshot 2025-11-27 at 2 16 18 PM

FSM state handling improvements:

* Added a `_backfill_fsm_states` method in `label_studio/tasks/serializers.py` to manually initialize FSM states for tasks created via `bulk_create()`, ensuring that FSM transitions registered with `triggers_on_create=True` are properly set even when the model's `save()` method isn't called. This includes robust error handling and logging to avoid interrupting imports if FSM backfill fails for any task or batch.

UI label update:

* Changed the label for the `CREATED` state from "Created" to "Initial" in `state-registry.ts` to better reflect its meaning in the UI.


yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19769203134)

bmartel (approved):

PR #8915, opened 2025-11-29
chore: BROS-648: Remove fflag_feat_front_bros58_timeseries_multichannel_short

makseq:
chore: BROS-648: Remove fflag_feat_front_bros58_timeseries_multichannel_short


web/libs/editor/tests/integration/e2e/timeseries/charts-displaying.cy.ts
```diff
@@ -139,10 +129,6 @@ describe("TimeSeries charts displaying - MultiChannel", () => {
 });
 describe("TimeSeries charts displaying - Single Channel", () => {
   it("should not displace charts while navigating", () => {
-    LabelStudio.setFeatureFlags({
-      fflag_feat_front_bros58_timeseries_multichannel_short: false,
-    });
-
```
yyassi-heartex, line 145: it looks like we're removing the feature flag set but we're not removing the related tests - I see the same thing where we set the flag to true. this means the tests will fail because the tests for when FF is on and FF is off are still there and will likely have opposite expectations

makseq: Thank you for the review! After this change the flag no longer exists and there is only one implementation path - the same renderer is used for both multi-channel and single-channel configs. The tests remain valid because they now simply exercise two different TimeSeries configurations:
  - The “MultiChannel” tests feed multiChannelConfig, which still renders multiple channels (the primary flow we just made permanent).
  - The “Single Channel” sections feed singleChannelConfig, which renders a single-channel timeline using the same renderer. These cases still provide value because they ensure resizing, navigation, etc. work for single-channel datasets, even though they’re no longer toggling a different feature flag.

So there isn’t an “opposite expectation” anymore - both test groups expect the same renderer but with different inputs. If we want to make that clearer, we can rename the describe blocks or add a short comment explaining that the “single channel” suite now validates behavior with single-channel configs rather than a legacy renderer.

yyassi-heartex (approved):

hlomzik (approved):

PR #8926, opened 2025-12-01
revert: fix: FIT-1058: I don't see task states in JSON Export

yyassi-heartex:
This reverts commit dcef58c5d9650bfaf724401606555e2bb82a8bcd.


yyassi-heartex:
we likely no longer need to worry about this - the selenium test failures required test updates to unrelated change

PR #8953, opened 2025-12-04
fix: BROS-633: Fix Labeling Stream with missing buttons

hlomzik:
If we go to View All in Quick View and then go to Label Stream, we'll end up with a broken state: part of LSF thinks it's View All, while the UI shows only one annotation.
Fix `viewingAll` getter to return false in such cases.


hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/19976686363)
> Successfully merged:  create mode 100644 web/libs/editor/src/components/TopBar/__tests__/Controls.test.tsx

PR #8959, opened 2025-12-05
docs: Add clarity about Manager access in Analytics

caitlinwheeless:
call out the fact that Managers can see full user list in the performance dashboard


shereezhang (approved):
looks good, thanks for adding!

PR #8981, opened 2025-12-10
fix: FIT-1120: label-stream cypress test failing consistently

yyassi-heartex:
This pull request introduces an important fix to the task navigation logic in the data manager queue of the labeling workflow. The change ensures that when a user clicks "next" in the label stream, they are not repeatedly assigned the same locked task, allowing for smooth progression through tasks.

Task navigation improvements:

* Updated the logic in `get_next_task` (in `label_studio/projects/functions/next_task.py`) to exclude the currently locked task when using the data manager queue, preventing users from getting stuck on the same task when navigating forward.

### DIFF
--- a/label_studio/projects/functions/next_task.py
+++ b/label_studio/projects/functions/next_task.py
@@ -405,7 +405,19 @@ def get_next_task(
             if dm_queue:
                 queue_info += (' & ' if queue_info else '') + 'Data manager queue'
                 logger.debug(f'User={user} tries sequence sampling from prepared tasks')
-                next_task = not_solved_tasks.first()
+                # IMPORTANT: When using data manager queue, we need to exclude the currently locked task
+                # to prevent returning the same task repeatedly when navigating forward in label stream.
+                # Without this, clicking "next" would return the same task because:
+                # 1. not_solved_tasks excludes completed tasks but not locked tasks
+                # 2. .first() would keep returning the locked task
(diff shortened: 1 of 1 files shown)


label_studio/projects/functions/next_task.py
```diff
@@ -405,7 +405,19 @@ def get_next_task(
...
                 queue_info += (' & ' if queue_info else '') + 'Data manager queue'
                 logger.debug(f'User={user} tries sequence sampling from prepared tasks')
-                next_task = not_solved_tasks.first()
+                # IMPORTANT: When using data manager queue, we need to exclude the currently locked task
+                # to prevent returning the same task repeatedly when navigating forward in label stream.
+                # Without this, clicking "next" would return the same task because:
+                # 1. not_solved_tasks excludes completed tasks but not locked tasks
+                # 2. .first() would keep returning the locked task
+                # 3. The user would be stuck on the same task unable to progress
+                locked_task = Task.get_locked_by(user, tasks=not_solved_tasks)
```
nass600, line 414: Seems like a fix for the tests but I wonder how this will affect production logic

yyassi-heartex: valid concern - i'd much rather a frontend solution - but this seems to be the only consistent solution

yyassi-heartex:
closing this PR - addressing issue by cleaning up the test

PR #9003, opened 2025-12-12
fix: FIT-1129: YouTube embedded player no longer works inside Label Studio (HyperText/iframe), videos play only when opened directly on YouTube

yyassi-heartex:
Iframes were previously disabled altogether - with this change we allow youtube iframes when inline="true". beyond that embeds as before should work

[screen-recorder-fri-dec-12-2025-15-56-50.webm](https://github.com/user-attachments/assets/8d344b5e-46db-4996-b3e3-1bbd377c98a6)

**Security and Sanitization Enhancements:**

* Added a whitelist of allowed iframe domains (YouTube and related domains) and a helper function to ensure only HTTPS iframes from these domains are allowed in the HTML sanitizer (`sanitizeHtml` in `html.js`).
* Updated the HTML sanitizer to discard iframes not from allowed domains, ensuring only safe embeds are rendered (`sanitizeHtml` in `html.js`). [[1]](diffhunk://#diff-de49f3b6544a835e65a60d923938854afd4222fe05d587cfbe7c0132c200b797L616-R668) [[2]](diffhunk://#diff-de49f3b6544a835e65a60d923938854afd4222fe05d587cfbe7c0132c200b797R683-R684)

**Rich Text Editor Improvements:**

* Expanded the `sandbox` attribute on rich text iframes to include `allow-presentation`, improving compatibility with embedded presentations or videos (`RichTextPieceView` in `view.jsx`).

**Styling Adjustments:**

* Changed iframe sizing from `width`/`height` to `min-width`/`min-height` to improve layout flexibility in the project config page (`Config.scss`).


nass600 (approved):

web/libs/editor/src/utils/html.js
```diff
@@ -520,6 +520,47 @@ function findNodeAt(context, at) {
...
   if (!html) return "";
 
+  // Whitelist of allowed iframe domains - easily extensible for future additions
+  const ALLOWED_IFRAME_DOMAINS = [
+    "www.youtube.com",
+    "youtube.com",
+    "www.youtube-nocookie.com",
+    "youtube-nocookie.com",
+    "youtu.be", // YouTube's shortened URL format
+  ];
```
hlomzik, line 530: so all non-youtube iframes will be blocked? is it correct?

hlomzik (changes requested):

web/libs/editor/src/utils/html.js
yyassi-heartex: thats right

yyassi-heartex: we can white list more in the future - but given that we had iframes disabled before it feels right to conditionally enable it

hlomzik: ah, so you are actually allowing iframes which were completely blocked by html sanitization before! nice! let's rewrite PR description there because right now it says
> This pull request enhances the security and usability of embedded iframes

and in fact we introduce embedded iframes!

web/libs/editor/src/tags/object/RichText/view.jsx
```diff
@@ -587,7 +587,7 @@ class RichTextPieceView extends Component {
           key="root"
           className={cn("richtext").elem("iframe").mix("htx-richtext").toClassName()}
           referrerPolicy="no-referrer"
-          sandbox="allow-same-origin allow-scripts"
+          sandbox="allow-same-origin allow-scripts allow-popups allow-popups-to-escape-sandbox allow-presentation"
```
hlomzik: You use `inline="true"` in the video and it actually makes sense to not have stacked iframes. Also this attribute applies to top-level iframe not the youtube one.
I'd recommend to check youtube iframes with both values for `inline` without changes to `sandbox` param and if it works then we can roll back changes here. Check fullscreen in youtube as well!
And if it requires this change to `sandbox` then we can leave it here.

hlomzik: In any case I don't think we need `allow-popups allow-popups-to-escape-sandbox` in `sandbox` though

web/libs/editor/src/utils/html.js
```diff
@@ -520,6 +520,47 @@ function findNodeAt(context, at) {
...
+    if (!src) {
+      console.log("[isAllowedIframeSrc] No src provided, blocking");
+      return false;
+    }
+
+    // Allow template variables (e.g., $partner_public_url) - they'll be validated after substitution
+    if (src.startsWith("$")) {
+      console.log("[isAllowedIframeSrc] Template variable detected, allowing:", src);
+      return true;
+    }
```
hlomzik: HTML sanitization receives final html with substituted values, so we should not get url starting with `$`, so better to remove this branch.

hlomzik: And all debugging console logs as well

web/libs/editor/src/utils/html.js
```diff
@@ -613,17 +654,51 @@ function sanitizeHtml(html = []) {
...
-  return sanitizeHTML(html, {
+  const result = sanitizeHTML(html, {
     allowedTags: false,
     allowedAttributes: false,
     disallowedTagsMode: "discard",
     allowVulnerableTags: true,
     exclusiveFilter(frame) {
-      //...except those in the blacklist
+      // Block scripts unconditionally
+      if (frame.tag === "script") return true;
```
hlomzik:
```suggestion
```
It will be blocked by `disallowedTags` check

web/libs/editor/src/utils/html.js
```diff
@@ -613,17 +654,51 @@ function sanitizeHtml(html = []) {
...
+          if (attribs[attr]) {
+            allowedIframeAttribs[attr] = attribs[attr];
+          }
+        });
+
+        return {
+          tagName,
+          attribs: allowedIframeAttribs,
+        };
+      },
```
hlomzik, line 668: only youtube iframes are allowed and they might require some additional attributes, so better to not limit them for now
```suggestion
```

web/libs/editor/src/utils/html.js
yyassi-heartex: updated!

web/libs/editor/src/utils/html.js
yyassi-heartex: cleaned up

hlomzik (approved):

PR #9010, opened 2025-12-15
docs: FIT-1124: Async import returns HTTP 201 but fails silently when task data or predictions don’t match label_config

yyassi-heartex:
This pull request updates the API documentation and response schema for task import and project import endpoints to clarify the differences between Community and non-Community (Enterprise) editions, especially regarding asynchronous import behavior. The changes improve both the OpenAPI documentation and the response format descriptions, making it clear how clients should handle imports in each edition.

**API Documentation and Response Schema Improvements**

* Clarified the `task_create_response_scheme` object in `label_studio/data_import/api.py` to specify that non-Community editions process imports asynchronously, returning an `import_id` for polling, while Community edition processes synchronously and returns task counts. Added the `import` field and updated descriptions for all properties to indicate which edition they apply to. [[1]](diffhunk://#diff-0d189d758de143ca7197ad41a825d2285df6df57e018672bfe1451dc2785dd3bL57-R90) [[2]](diffhunk://#diff-0d189d758de143ca7197ad41a825d2285df6df57e018672bfe1451dc2785dd3bL94-R104) [[3]](diffhunk://#diff-0d189d758de143ca7197ad41a825d2285df6df57e018672bfe1451dc2785dd3bL108-R113)
* Added a new section to the import endpoint docstring describing async import behavior for non-Community editions, including instructions for polling import status and where to find error information.

**Project Import Status Endpoint Documentation**

* Updated the `GET /api/projects/{project_id}/imports/{import_id}` endpoint documentation to clarify its use for polling asynchronous import status, including usage steps, and specifying that errors are only visible in the GET response.
* Updated the `GET /api/projects/{project_id}/reimport/{reimport_id}` endpoint documentation similarly, clarifying polling usage and where to find error details.

### DIFF
--- a/label_studio/data_import/api.py
+++ b/label_studio/data_import/api.py
@@ -54,35 +54,40 @@ ProjectImportPermission = load_func(settings.PROJECT_IMPORT_PERMISSION)
 
 task_create_response_scheme = {
     201: OpenApiResponse(
-        description='Tasks successfully imported',
+        description='Tasks successfully imported or import queued. **For non-Community editions**, the response will be `{"import": <import_id>}` which you can use to poll the import status. **For Community edition**, the response contains task counts and is processed synchronously.',
         response={
             'title': 'Task creation response',
-            'description': 'Task creation response',
+            'description': 'Response format varies by edition. Non-Community editions return `{"import": <import_id>}` for async processing. Community edition returns the detailed response below with task counts.',
             'type': 'object',
             'properties': {
+                'import': {
+                    'title': 'import',
+                    'description': 'Import ID for async operations (non-Community editions only). Use this ID to poll `/api/projects/{project_id}/imports/{import_id}` for status.',
+                    'type': 'integer',
+                },
                 'task_count': {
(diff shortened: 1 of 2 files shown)


label_studio/data_import/api.py
```diff
@@ -177,6 +182,19 @@
             must include a "text" field.
             <br>
 
+            ## Async Import Behavior
+            <hr style="opacity:0.3">
+
+            **For non-Community editions, this endpoint processes imports asynchronously.**
+            
+            - The POST request will **not** fail if something goes wrong during import.
```
mcanu: It can fail if wrong parameters, body, etc is passed. But for data validation errors that will be async. I would specify that. We don't "immediatly" return an OK response.

mcanu (approved):

bmartel (approved):

label_studio/data_import/api.py
yyassi-heartex: updated

nass600 (approved):

PR #9023, opened 2025-12-16
docs: Annotator eval enhancements and unskippable tasks

caitlinwheeless:
Update the project settings page to reflect changes to annotator eval settings

Add a new page on skipping tasks


docs/source/guide/skip.md
```diff
@@ -0,0 +1,78 @@
...
+
+* [**Annotation > Task Skipping**](project_settings_lse#skip)
+
+* [**Annotation > Skip Queue**](project_settings_lse#skip-queue)
+
+While you can disallow skipping entirely from the project settings, if you want to have specific tasks be unskippable, you will need to configure that by adding a special key as part of the JSON task definition that you import to your project.
+
+</div>
+
+## Unskippable tasks
```
nass600: I believe Individual unskippable tasks will clarify a bit more

docs/source/guide/skip.md
```diff
@@ -0,0 +1,78 @@
...
+
+To make a task unskippable, specify `"allow_skip": false` as part of the JSON task definition that you import to your project.
```
nass600: I believe we could explain it as a process so newcomers find it easy:

1. Once the project is created import tasks via file (screenshoot of the data manager import)
2. The json must contain... (example as the one is below)
3. Once imported, you can keep track of which tasks are unskippable via toggling the column visibility or use it as a filter (screenshots)

Important: This will affect only reviewers and annotators, if you are a manager+ you will be able to bypass this setting but you will find out an explanatory tooltip next to the skip button (info icon screenshoot)

docs/source/guide/skip.md
caitlinwheeless: yes! agreed - will update this

docs/source/guide/skip.md
caitlinwheeless: I made some of these changes by adding a note, but didn't include screenshots. When a page is a mix of LSE only and LSO audiences, lots of screenshots get tricky due to the differences in the UI

nass600 (approved):

PR #9066, opened 2025-12-23
feat: UTC-447: Contributed projects fields in membership detail endpoint

mcanu:
This PR was automatically generated via Follow Merge. [Workflow Run](https://github.com/HumanSignal/label-studio/actions/runs/20488830051)
- https://github.com/HumanSignal/label-studio-sdk/pull/714


label_studio/organizations/serializers.py
```diff
@@ -21,25 +26,37 @@ class Meta:
...
+
 class OrganizationMemberListParamsSerializer(serializers.Serializer):
     active = serializers.BooleanField(required=False, default=False)
     contributed_to_projects = serializers.BooleanField(required=False, default=False)
 
 
+@extend_schema_serializer(
+    deprecate_fields=[
+        'created_projects',
+        'contributed_to_projects',
```
matt-bernstein, line 46: why is this being deprecated if you're adding it?

mcanu: these two fields already appear nested in the `user` field when listing memberships. I moved those fields to the root of the response, so we don't "contaminate" the user entity with this fields only relevant to this endpoint. The detail endpoint didn't have these, so I added them to the root directly

mcanu: The deprecated fields are only the ones nested in user not the ones in the root.

matt-bernstein (approved):

mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/20488759076)

PR #9072, opened 2025-12-24
ci: test LLM migration check

nikitabelonogov:
DO NOT MERGE

### DIFF
--- a/label_studio/data_manager/migrations/0018_remove_allow_skip.py
+++ b/label_studio/data_manager/migrations/0018_remove_allow_skip.py
@@ -13,7 +13,6 @@ from django.conf import settings
 from django.db import migrations
 
 from core.redis import start_job_async_or_sync
-from core.utils.iterators import iterate_queryset
 
 logger = logging.getLogger(__name__)
 migration_name = __name__
@@ -36,7 +35,7 @@ def forward_migration_job(*, migration_name: str) -> None:
         migration.save()
 
     try:
-        views = iterate_queryset(View.objects.all())
+        views = View.objects.all()
         updated_count = 0
 
         for view in views:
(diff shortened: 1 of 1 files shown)


nikitabelonogov (commented):
## Migration Review

🚨 **Memory Risk**: Removing `iterate_queryset` and using `View.objects.all()` directly will load all View objects into memory at once, which can cause OOM errors in production with large datasets.

**Recommendation**: Use `View.objects.all().iterator()` to process views in chunks, or use `View.objects.all().values('id', 'data')` with `update()` if full model instances aren't needed.

label_studio/data_manager/migrations/0018_remove_allow_skip.py
```diff
@@ -36,7 +35,7 @@ def forward_migration_job(*, migration_name: str) -> None:
         migration.save()
 
     try:
-        views = iterate_queryset(View.objects.all())
+        views = View.objects.all()
```
nikitabelonogov, line 38: 🚨 Memory risk: Using `View.objects.all()` directly loads all View objects into memory. Use `View.objects.all().iterator()` to process in chunks.

label_studio/data_manager/migrations/0018_remove_allow_skip.py
```diff
@@ -75,7 +74,7 @@ def reverse_migration_job(*, migration_name: str) -> None:
 
     logger.info(f'Starting reverse migration {migration_name}')
 
-    views = iterate_queryset(View.objects.all())
+    views = View.objects.all()
```
nikitabelonogov, line 77: 🚨 Same memory risk in reverse migration. Use `View.objects.all().iterator()` instead.

nikitabelonogov (commented):
## Summary

🚨 **Critical Memory Issue**: Removing `iterate_queryset` reintroduces OOM risk by loading all View objects into memory. Both forward and reverse migrations should use `View.objects.all().iterator()` to process in chunks.

PR #9117, opened 2026-01-05
fix: BROS-689: Fix Download icon in Dark mode

hlomzik:
Cherry-pick from https://github.com/HumanSignal/label-studio/pull/9109


hlomzik:
Wrong initial base

PR #9141, opened 2026-01-07
fix: FIT-1153: StateChip does not contrast with various backgrounds

bmartel:
The utility for Badge component wasn't using the proper tailwind merge version of the classnames utility, and ended up incorrectly with duplicated tailwind classes in the same group, ie. `border-transparent` in this case would never be able to be supplied an override and include a border of any color as both classes would be present on the element, and border-transparent took precedence. This made Badges have very low or no contrast on same or similar backgrounds in Headers, Cards etc.

Screenshot 2026-01-07 at 9 18 41 AM
Screenshot 2026-01-07 at 9 20 09 AM
Screenshot 2026-01-07 at 9 20 01 AM
Screenshot 2026-01-07 at 9 19 36 AM
Screenshot 2026-01-07 at 9 19 26 AM
Screenshot 2026-01-07 at 9 19 03 AM
Screenshot 2026-01-07 at 9 18 49 AM

### DIFF
--- a/web/libs/ui/src/shad/components/ui/badge.tsx
+++ b/web/libs/ui/src/shad/components/ui/badge.tsx
@@ -1,7 +1,7 @@
 import type * as React from "react";
 import { cva, type VariantProps } from "class-variance-authority";
 
-import { cn } from "@humansignal/shad/utils";
+import { cnm } from "@humansignal/shad/utils";
 
 const badgeVariants = cva(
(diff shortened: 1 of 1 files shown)


yyassi-heartex (approved):

bmartel:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/20788045191)

PR #9159, opened 2026-01-12
fix: BROS-708: Zoom shape corruption

nick-skriabin:
This PR addresses the issue when dragging shape points in zoomed-in view would cause abrupt zoom reset and shape corruption.

The root cause was in the legacy transformation logic (pre-group-transform) that was running after every transform of a shape. This logic was unnecessary in the current setup and group transformations (multi-shape) are handled elsewhere.

Removing the logic fixes the bug, zoom reset doesn't happen anymore, shape stays stable after all transformations.


yyassi-heartex (approved):
was this not used else where? cleanup is great i just want to make sure we're not removing things un-expectedly

PR #9161, opened 2026-01-12
fix: BROS-712: Free form transformation on double click

nick-skriabin:
This PR addresses the issue when on double clicking an unselected region, the user will enter free-form transform mode for a brief second.

Before: on UNSELECTED region double-click it, the region will become selected, the transformer will appear for a brief second, then disappear.

After: on UNSELECTED region click will select the shape, double-click will enter transformation mode ONLY when the shape is selected.
 
Reasoning: free-form transformation is an editing operation. Most of the editing operations can only be performed on a currently selected region. Transform is no exception, so we're moving this behavior under the selected region guard.

### DIFF
--- a/web/libs/editor/src/regions/VectorRegion.jsx
+++ b/web/libs/editor/src/regions/VectorRegion.jsx
@@ -246,9 +246,16 @@ const Model = types
           const wasNotSelected = !self.selected;
 
           if (wasNotSelected) {
+            // Set the flag before selecting to prevent double-click issues
+            // This will be cleared by selectRegion() when called from RegionStore
+            self._justSelected = true;
             annotation.selectArea(self);
           } else {
-            annotation.unselectAll();
+            // If _justSelected is true, it means this click is part of a double-click
+            // Don't unselect - let the double-click handler manage selection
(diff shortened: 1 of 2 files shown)


web/libs/editor/src/regions/VectorRegion.jsx
```diff
@@ -430,14 +433,29 @@ const Model = types
...
-        self.setTransformMode(false);
+        // unless preserveTransformMode is true
+        if (!preserveTransformMode) {
+          self.setTransformMode(false);
+        }
+        // Mark that we just selected this region (to prevent double-click from enabling transform mode)
+        self._justSelected = true;
+        setTimeout(() => {
+          self.clearJustSelectedFlag();
+        }, 300); // Clear after double-click detection window
```
yyassi-heartex, line 465: might be good to clear time out on beforeDestroy or on cleanup

yyassi-heartex (approved):
minor recommendation - but looks good otherwise approving

yyassi-heartex:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/20930486614)

PR #9198, opened 2026-01-16
fix: BROS-722: Unselection and skeleton mode

nick-skriabin:
This PR addresses the following issues:

**Region re-selects when explicitly exiting drawing via Esc key.**
This was cause by "Select region after creating it" setting. Fix ensures that for Vector we DO NOT reselect unfinished shapes (e.g. shapes that haven't reached maxPoints or not closed when `closable=true`)

**Inability to continue drawing from either side of the shape when `skeleton=false`.**
The issue contained two smaller ones: no matter the mode, user could select ANY point on the shape BUT could only draw from the last even if it's not selected.
Fix ensures that in `skeleton=off` user can only select first or last point of the shape and can continue drawing from either of these, but can't select or draw from any points in between.


nass600 (approved):

nick-skriabin:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21147985557)

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21360837048)

PR #9211, opened 2026-01-19
fix: FIT-287: [FE] Large labeling interface causing to freeze LS

yyassi-heartex:
This pull request introduces a significant performance improvement to the project configuration page by adding a new "lightweight" static preview mode for large labeling configurations. The lightweight preview bypasses MobX-State-Tree (MST) and uses a fast, React-only renderer, providing instant feedback for users editing very large configs. The preview automatically switches between full and lightweight modes based on config size and content. Additionally, the codebase is refactored to make preview selection adaptive and robust, and new preview components are added to the shared core library.

**Adaptive Preview and Performance Improvements:**

* Introduced an `AdaptivePreview` component in `Config.jsx` that automatically selects between the full interactive preview and the new lightweight static preview based on config size and the presence of `<ReactCode>` tags. For large configs (over 50,000 characters) without custom React code, the lightweight preview is used and an informational banner is displayed. [[1]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R28-R59) [[2]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L609-R645) [[3]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L4-R4) [[4]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R14)

* Added a new `LightweightPreview` component to the core library, which renders configs using fast, static React components without MST. This provides instant updates and avoids UI blocking for large configs. [[1]](diffhunk://#diff-ec443aacc52867d5b85e5ab52940e677b43c1993879f71d9f3d045a20534085aR4) [[2]](diffhunk://#diff-bffff1792427df5d9dcfcdc2cef1018e01f506146fbaea7639a0f132e8a450d0R1-R105) [[3]](diffhunk://#diff-bfbedecab9123ab9e6244b9b0469929d90b59731b1e72ee8b3cf63f0a743d1a1R1-R7)

* Implemented `StaticConfigPreview`, a subcomponent that parses and renders Label Studio XML configs directly, showing placeholders for media and unknown tags. [[1]](diffhunk://#diff-81d8d6b9322af490986e032c41269367eb8bed9847d2356890f0a126c00e6b67R1-R326) [[2]](diffhunk://#diff-bfbedecab9123ab9e6244b9b0469929d90b59731b1e72ee8b3cf63f0a743d1a1R1-R7)

* Added preview-related constants (`LIGHTWEIGHT_PREVIEW_MESSAGE`, `LIGHTWEIGHT_PREVIEW_THRESHOLD`) to the core library for consistent messaging and threshold management. [[1]](diffhunk://#diff-f8b53444869392203a2dfe0494e51f0478381fba5f4330f900ce86b1ae76ec73R1-R10) [[2]](diffhunk://#diff-bfbedecab9123ab9e6244b9b0469929d90b59731b1e72ee8b3cf63f0a743d1a1R1-R7)

**UX and Code Quality Improvements:**

* Improved config change handling by debouncing validation API calls to 300ms, reducing unnecessary network requests during rapid typing.

* Refactored `setTemplate` and `setConfig` logic in `ConfigPage` for better error handling and state management when updating the template from user actions.

* Updated UI library exports to include new `callout` component, supporting future UI enhancements.


web/libs/core/src/lib/preview/StaticConfigPreview.tsx
```diff
@@ -0,0 +1,311 @@
...
+  const children = useMemo(() => {
+    const childElements = Array.from(element.children);
+    if (childElements.length === 0) return null;
+    return childElements.map((child, i) => (
+      <RenderElement key={`${child.tagName}-${i}`} element={child} context={{ ...context, depth: context.depth + 1 }} />
+    ));
+  }, [element, context]);
+
+  // Tag-specific rendering
+  switch (tagName) {
```
bmartel: I didn't realize that was how it was achieving the performance 🤔. Two concerns I have with this:

1. 50k char length is likely not the issue, its MST node count created with the tree of elements.
2. The lightweight preview might as well just be a hidden or grayed out preview in cases of heavy config sizes unless the output visually ends up the same with just reduced footprint, otherwise what the user is interacting with would have no correlation whatsoever with what an annotator would see.

Other than that, I really think this approach has promise and can help to at least make the upper bound use case management possible as the MST node thrash in the CodeEditor -> Preview is what causes the most prevalent freezes.

yyassi-heartex: > 50k char length is likely not the issue, its MST node count created with the tree of elements.

you're right but if we're spinning up MST and counting nodes then we've already lost the battle :P

bmartel: We can approximate based on number of tags, I think tags are a closer measure to how heavy a config is rather than length of the string.

yyassi-heartex: fair enough, i'll get that going in the morning

yyassi-heartex: 1) is resolved, we now check for 200 tags before we toggle to light weight preview.
2) lets chat about when you get a minute because im not sure if a greyed out preview will do it but also what I currently have is probably not enough either. wondering if we should wrap this in a FF

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21300556840)

bmartel (approved):

PR #9223, opened 2026-01-21
fix: PLT-994: OOM in data manager list api

makseq:
## Summary

- Fix OOM (Out of Memory) issues in `/api/tasks` endpoint by preventing duplicate loading of heavy task data fields during pagination
- Reduces memory usage by ~50% for projects with large `task.data` fields (e.g., TimeSeries data)

## How we found it

A user reported 502 Bad Gateway errors on production for the `/api/tasks` endpoint. Datadog logs showed:
- Request: `GET /api/tasks?page=1&page_size=30&view=460599&project=223282`
- Duration: 6.99s before failing with 502
- All `x_ls_*` headers were empty (indicating the request never completed)

We created a diagnostic script to simulate the API call step-by-step with memory tracking, which revealed the root cause.

See script here:
[script_debug_task_api.py](https://github.com/user-attachments/files/24771999/script_debug_task_api.py)

## The problem

The `TaskListAPI.get()` endpoint was loading full task objects **twice**:

1. **First load (pagination):** `page = self.paginate_queryset(queryset)` loads full Task objects including the heavy `data` field, just to extract IDs
2. **Second load (with annotations):** `Task.prepared.annotate_queryset(Task.objects.filter(id__in=ids))` loads the same tasks again with annotations

For a project with ~5.8 MB of TimeSeries data per task:
- Step 1: +739 MB for 30 tasks (full objects loaded for pagination)
- Step 2: +726 MB for 30 tasks (same objects loaded again with annotations)
- **Total: ~1.5 GB for just 30 tasks**

The pagination step only needs task IDs, but was loading entire task objects including multi-megabyte `data` fields.

## The fix

Use Django's `.only('id')` to defer loading heavy fields during pagination:

```python
def sync_paginate_queryset(self, queryset, request, view=None):
    self.total_predictions = Prediction.objects.filter(task_id__in=queryset).count()
    self.total_annotations = Annotation.objects.filter(task_id__in=queryset, was_cancelled=False).count()
    # Use .only('id') to avoid loading heavy task.data fields during pagination
    # Full task objects are loaded later with proper annotations
    id_only_queryset = queryset.only('id')
    return super().paginate_queryset(id_only_queryset, request, view)
```

This is safe because:
- The paginated `page` is only used to extract IDs: `ids = [task.id for task in page]`
- Full task objects are loaded separately with proper annotations and prefetching
- Ordering/filtering still works (annotations for ORDER BY are still computed)

## Expected impact

| Metric | Before | After |
|--------|--------|-------|
| Pagination memory | ~740 MB | ~1 MB |
| Total memory (30 tasks) | ~1.5 GB | ~750 MB |
| Memory reduction | - | **~50%** |

## Test plan

- [ ] Run existing Data Manager tests
- [ ] Test pagination with filters and ordering
- [ ] Test with projects containing large `task.data` fields
- [ ] Verify task list displays correctly in Data Manager UI

---


mcanu (approved):

farioas:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21220900382)

makseq:
/selenium test
> See '/help'

triklozoid:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21244698926)

PR #9260, opened 2026-01-26
fix: FIT-1294: Clear window.Htx in V17 destroy path

bmartel:
## Problem

`window.Htx` is never cleared in the V17 destroy path, retaining the entire MST store in memory after navigation. This causes memory leaks as the store and all its references remain accessible via the global window object.

## Solution

Always set `window.Htx = null` in the `destroy()` method regardless of feature flags. This allows garbage collection of the store after destruction.

## Files Changed

- `web/libs/editor/src/LabelStudio.tsx`


robot-ci-heartex:

robot-ci-heartex:

PR #9262, opened 2026-01-26
perf: FIT-1296: Consolidate extractNames into single tree traversal

bmartel:
## Problem

`extractNames()` in Tree.tsx traverses the config tree twice - once to build the names map and once to process toName bindings. For configs with 200+ tags, this doubles the initialization time.

## Solution

Consolidate both traversals into a single pass:
- Use `Set` for objectTypes lookup (O(1) instead of O(n) includes)
- Collect control tags during traversal for toName processing afterward
- Reduces traversal from 400+ node visits to 200+ for large configs

## Files Changed

- `web/libs/editor/src/core/Tree.tsx`


robot-ci-heartex:

PR #9264, opened 2026-01-26
perf: FIT-1298: Cache Area union types by tag type in dispatcher

bmartel:
## Problem

The Area dispatcher creates a new `types.union()` for every region deserialization, even when the same tag type would produce the same union. For annotations with many regions, this creates thousands of duplicate union types.

## Solution

Add a cache keyed by tag type so that the union is only created once per tag type and reused for subsequent areas of the same type.

## Files Changed

- `web/libs/editor/src/regions/Area.js`


robot-ci-heartex:

robot-ci-heartex:

PR #9270, opened 2026-01-26
feat: add strict task overlap filtering for main queue

bmartel:
## Summary
- Add filtering to `get_not_solved_tasks_qs()` to exclude tasks that have already reached their annotation overlap limit
- Add frontend handling in `lsf-sdk.js` for `overlap_reached` flag - disables submit/update/skip buttons and shows informational toast
- Postponed queue is intentionally NOT filtered - users see their postponed tasks in read-only mode with clear messaging

## Test plan
- [ ] Verify main queue filtering excludes tasks at overlap for new users
- [ ] Verify users who already have annotations can still access their tasks
- [ ] Verify frontend shows informational message when overlap is reached
- [ ] Verify navigation (prev/next) remains functional when overlap is reached

## Related
- Requires label-studio-enterprise PR for full functionality (project setting and API validation)

### DIFF
--- a/label_studio/projects/functions/next_task.py
+++ b/label_studio/projects/functions/next_task.py
@@ -231,18 +231,26 @@ def get_not_solved_tasks_qs(
     # so users can see their work and understand why they can't submit
     lse_project = getattr(project, 'lse_project', None)
     if lse_project and getattr(lse_project, 'strict_task_overlap', False):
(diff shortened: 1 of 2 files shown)


bmartel:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21375015092)

PR #9291, opened 2026-01-29
fix: BROS-711: Decrease points hit radius to make it easier to drag vectors themselves

nick-skriabin:
This PR addresses the issue when it's hard to position entire shape. Often instead of dragging the shape as a whole the user starts dragging individual nodes. This is caused by a large points' hit radius. This PR decreases the radius to 10px so the points are still easy to aim, but at the same time it's easier to drag entire shape when necessary.


nass600 (approved):

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21602864890)
> Successfully merged:  create mode 100644 web/libs/ui/src/utils/getTokenColor.ts

PR #9292, opened 2026-01-29
fix: FIT-1276: Graph lines at Analytics dashboards are not adapted to Dark Mode

ricardoantoniocm:
This pull request introduces a new theme-aware color token utility for the UI library, making it easier to resolve and use design token colors in React components. The main additions are a `getTokenColor` utility for resolving color tokens (with support for themes, alpha, and output formats) and a `useTokenColor` React hook that provides a convenient, theme-aware `getColor` function. These are now exported from the UI library for use throughout the codebase.

**Theme-aware color token utilities:**

* Added `getTokenColor` utility in `utils/getTokenColor.ts` to resolve color tokens from `design-tokens.json`, supporting theme-specific values, alpha transparency, and multiple output formats (hex, rgb, rgba).
* Added `useTokenColor` React hook in `hooks/useTokenColor.ts` to provide a theme-aware `getColor` function that automatically uses the current theme from `themeAtom` and re-renders when the theme changes.

**Exports and library integration:**

* Exported `getTokenColor` and `useTokenColor` from the main UI library index (`index.ts`) for easy import and usage in other parts of the application.

### DIFF
--- a/web/libs/ui/src/hooks/useTokenColor.ts
+++ b/web/libs/ui/src/hooks/useTokenColor.ts
@@ -1,46 +1,11 @@
-/**
- * React hook for theme-aware color token resolution
- *
- * This hook provides a `getColor` function that automatically uses the current theme
- * from the themeAtom. It re-renders components when the theme changes.
- *
- * @example
- * ```typescript
- * const MyComponent = () => {
- *   const getColor = useTokenColor();
- *
- *   return (
- *     <div style={{ backgroundColor: getColor("primary-surface") }}>
- *       <span style={{ color: getColor("neutral-content") }}>
- *         Hello
- *       </span>
- *     </div>
- *   );
- * };
- * ```
- */
-
 import { useCallback } from "react";
 import { useAtomValue } from "jotai";
 import { themeAtom } from "../lib/ThemeToggle/ThemeToggle";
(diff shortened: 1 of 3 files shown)


web/libs/ui/src/hooks/useTokenColor.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const getColor = useCallback(
+    (tokenName: string, options?: Omit<GetTokenColorOptions, "theme">): string => {
+      return getTokenColor(tokenName, { ...options, theme });
+    },
+    [theme],
+  );
+
+  return getColor;
+}
```
yyassi-heartex, line 24: we have similar checks in other spots - if you have this happening here in a shared spot lets make sure we update other usages to use the same function

ricardoantoniocm: I did some digging and it looks like there’s nothing exactly like this. Other implementations just apply hard coded colors after detecting the current theme in specific scenarios.

There is an opportunity to improve components that use the Canvas, like the Audio/Video players (maybe TimeSeries).

That said, I wouldn’t want to couple refactoring other components in this PR but it’s definitely something we can do going forward or in follow up PRs. What do you think?

yyassi-heartex (approved):

PR #9296, opened 2026-01-29
fix: FIT-1327: Correct MultiChannel margin logic to align x-axis

bmartel:
### Problem
MultiChannel visualization introduced an incorrect left margin even when showYAxis was false, causing misalignment with other channels.

### Solution
Updated margin calculation in MultiChannel.jsx to respect showYAxis setting.

### Verification
Added regression test in multichannel.cy.ts that verifies pixel-perfect alignment between MultiChannel and standard Channel.

**Misaligned (Before)**
multichannel_misaligned

**Aligned (After)**
multichannel_aligned

### DIFF
--- a/web/libs/editor/src/tags/object/TimeSeries/MultiChannel.jsx
+++ b/web/libs/editor/src/tags/object/TimeSeries/MultiChannel.jsx
@@ -45,6 +45,9 @@ const Model = types
     },
 
     get margin() {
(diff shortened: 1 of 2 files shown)


bmartel:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21498842165)

yyassi-heartex (approved):

hlomzik (approved):

PR #9304, opened 2026-01-30
fix: FIT-1316: don't leak CurrentContext across RQ jobs

matt-bernstein:
Was seeing in develop that FSM history was not showing after this backfill because it was filtered out by organization, because the backfill was completed with the wrong organization_id being set! This affects all FSM functions that run on RQ and have empty context, since it doesn't get reset properly when the context is empty.

### DIFF
--- a/label_studio/core/redis.py
+++ b/label_studio/core/redis.py
@@ -189,11 +189,10 @@ def start_job_async_or_sync(job, *args, in_seconds=0, **kwargs):
         try:
             context_data = _capture_context()
 
-            if context_data:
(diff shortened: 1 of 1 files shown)


bmartel (approved):

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21636895082)

matt-bernstein:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21651355128)

PR #9306, opened 2026-01-30
feat: FIT-720: Add global image cache for annotation switching

yyassi-heartex:
Add ImageCache.ts utility to prevent re-downloading images when switching between annotations on the same task. This improves performance by:
- Caching images as blob URLs with reference counting
- Deduplicating concurrent requests for the same image
- Automatic cache eviction with LRU strategy
- Error recovery with retry logic for failed loads

Changes:
- Add ImageCache.ts with ImageCacheManager class
- Update ImageEntity.js to use global cache when FF enabled
- Add onError handler to Image.jsx for blob URL failures
- Add FF_FIT_720_LAZY_LOAD_ANNOTATIONS feature flag

All changes are gated behind fflag_fix_all_fit_720_lazy_load_annotations

### DIFF
--- a/web/biome.json
+++ b/web/biome.json
@@ -40,8 +40,7 @@
       "a11y": {
         "noAutofocus": "off",
         "useKeyWithClickEvents": "off",
-        "noLabelWithoutControl": "off",
-        "noRedundantAlt": "off"
+        "noLabelWithoutControl": "off"
       },
       "suspicious": {
         "noExplicitAny": "off",
--- a/web/biome.json
+++ b/web/biome.json
@@ -40,7 +40,8 @@
       "a11y": {
         "noAutofocus": "off",
         "useKeyWithClickEvents": "off",
-        "noLabelWithoutControl": "off"
+        "noLabelWithoutControl": "off",
+        "noRedundantAlt": "off"
       },
       "suspicious": {
         "noExplicitAny": "off",
--- a/web/libs/core/src/lib/utils/ImageCache.ts
+++ b/web/libs/core/src/lib/utils/ImageCache.ts
@@ -285,6 +285,18 @@ class ImageCacheManager {
       this.cache.delete(url);
     }
   }
+
+  /**
+   * Force clear all cached images and release memory
+   * Called when LSF is destroyed to prevent memory leaks
+   */
+  forceClear(): void {
+    for (const cached of this.cache.values()) {
+      this.safeRevokeBlobUrl(cached);
+    }
+    this.cache.clear();
+    this.pendingLoads.clear();
+  }
 }
 
 // Singleton instance - persists across annotation switches
--- a/web/libs/editor/src/components/ImageView/Image.jsx
+++ b/web/libs/editor/src/components/ImageView/Image.jsx
@@ -40,6 +40,13 @@ export const Image = observer(
       [updateImageSize, imageEntity],
     );
 
+    const onError = useCallback(() => {
+      // Handle image load failure (e.g., revoked blob URL, empty blob, network error)
+      // This sets the error state so the UI can show an error message
+      imageEntity.setError(true);
+      imageEntity.setImageLoaded(false);
+    }, [imageEntity]);
+
     return (
       <div className={cn("image").toClassName()} style={imageSize}>
         {overlay}
@@ -56,6 +63,7 @@ export const Image = observer(
             ref={ref}
             src={imageEntity.currentSrc}
             onLoad={onLoad}
+            onError={onError}
             isLoaded={imageEntity.imageLoaded}
             imageTransform={imageTransform}
           />
@@ -85,7 +93,7 @@ const ImageProgress = observer(({ downloading, progress, error, src, usedValue }
 const imgDefaultProps = { crossOrigin: "anonymous" };
 
 const ImageRenderer = observer(
-  forwardRef(({ src, onLoad, imageTransform, isLoaded }, ref) => {
+  forwardRef(({ src, onLoad, onError, imageTransform, isLoaded }, ref) => {
     const imageStyles = useMemo(() => {
       // We can't just skip rendering the image because we need its dimensions to be set
       // so we just hide it with 0x0 dimensions.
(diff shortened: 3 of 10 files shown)


web/libs/editor/src/components/ImageView/Image.jsx
```diff
@@ -40,6 +40,13 @@ export const Image = observer(
       [updateImageSize, imageEntity],
     );
 
+    const onError = useCallback(() => {
+      // Handle image load failure (e.g., revoked blob URL, empty blob, network error)
+      // This sets the error state so the UI can show an error message
+      imageEntity.setError(true);
+      imageEntity.setImageLoaded(false);
```
luarmr: looks like setError already set setImageLoaded to false, no big to ensure

web/libs/editor/src/tags/object/Image/ImageEntity.js
```diff
@@ -79,48 +84,110 @@ export const ImageEntity = types
...
+
+      // Check if already cached in global cache
+      const cached = imageCache.get(self.src);
+      if (cached) {
+        // Add reference to prevent cache eviction while we're using this image
+        imageCache.addRef(self.src);
+        self.setCurrentSrc(cached.blobUrl);
+        self.setDownloaded(true);
+        self.setProgress(1);
+        self.setDownloading(false);
```
luarmr: This looks like repeated code that can be grouped. it's not a big issue if it's only repeated twice, but if this sequence needs to be enforced, it's better to group it.

luarmr: actually maybe is more than twice

web/libs/editor/src/tags/object/Image/ImageEntity.js
```diff
@@ -79,48 +84,110 @@ export const ImageEntity = types
...
+          self.setProgress(1);
           self.setDownloading(false);
-          self.setCurrentSrc(url);
+          // DO NOT set imageLoaded here - wait for <img> onLoad
         })
         .catch(() => {
-          self.setDownloading(false);
-          self.setError(true);
+          // Fallback to old behavior if global cache fails
+          if (isFF(FF_IMAGE_MEMORY_USAGE)) {
```
luarmr, line 128: should be this feature flag be checked early on?

web/libs/editor/src/utils/ImageCache.ts
```diff
@@ -0,0 +1,344 @@
...
+        this.safeRevokeBlobUrl(value);
+        this.cache.delete(key);
+      }
+    }
+  }
+
+  /**
+   * Force clear all cache entries (use with caution - may break active images)
+   */
+  forceClear(): void {
```
luarmr: is this use?

web/biome.json
```diff
@@ -40,7 +40,8 @@
       "a11y": {
         "noAutofocus": "off",
         "useKeyWithClickEvents": "off",
-        "noLabelWithoutControl": "off"
+        "noLabelWithoutControl": "off",
+        "noRedundantAlt": "off"
```
luarmr: jum jum why?

web/libs/editor/src/utils/ImageCache.ts
```diff
@@ -0,0 +1,344 @@
...
+
+          // Validate blob size - reject empty or too small blobs
+          if (!blob || blob.size < this.minBlobSize) {
+            reject(new Error(`Empty or invalid image data received: ${url} (size: ${blob?.size ?? 0} bytes)`));
+            return;
+          }
+
+          // Validate content type is an image
+          if (blob.type && !blob.type.startsWith("image/")) {
+            reject(new Error(`Invalid content type for image: ${blob.type} (url: ${url})`));
```
luarmr: let's be vigilant on this errors, I am not 100% sure we want to populate those to sentry, maybe we should add the noSentry flag

web/libs/core/src/lib/utils/ImageCache.ts
```diff
@@ -0,0 +1,344 @@
...
+class ImageCacheManager {
+  private cache: Map<string, CachedImage> = new Map();
+  private pendingLoads: Map<string, Promise<CachedImage>> = new Map();
+  /** Track revoked blob URLs to detect invalid references */
+  private revokedUrls: Set<string> = new Set();
+
+  // Cache for 30 minutes by default
+  private readonly maxAge = 30 * 60 * 1000;
+  // Maximum cache size (100 images)
+  private readonly maxSize = 100;
```
luarmr, line 39: I believe this is correct, but I'm wondering if there should be a KB size constraint.

web/biome.json
yyassi-heartex: this one is because we have an alt="image" and with that part getting wrapped in forwardRef it caused it to go over 2 lines with return (
...
);
previously we had an inline ignore on it which wont work with the update. I didnt want to remove the alt tag incase there are tests that refrence it. felt it was safe to disable this lint rule

https://github.com/HumanSignal/label-studio/pull/9306/changes#diff-1d586dd58ae31c29ee4eb0128746541cd1c0862e8bb4dafafbb3f5fbb4294875R117

luarmr (approved):
Found neats but not logic suggestions

web/libs/core/src/lib/utils/ImageCache.ts
yyassi-heartex: we can always add one in the future - lets get out there and see impact

luarmr (commented):
Wondering if the main functionality exists as a library. Probably is heavier and we are ok with this custom solution. But is worth the research

web/biome.json
luarmr: dont you prefer to add that to  the line or the file instead the complete codebase?

yyassi-heartex: good point i can use a file specific one instead of whole code base - it just seemed like a useless rule. i'll ignore it for the file instead

web/libs/editor/src/utils/ImageCache.ts
yyassi-heartex: hmm - thought i asked claude to do that when switch tasks to release memory - will take a closer at that. currently it doesn't appear to be used

web/libs/editor/src/tags/object/Image/ImageEntity.js
yyassi-heartex: not really - it only comes into play on catch when we fall back if cache failed - but with the refactor maybe we can shift it around a little

web/libs/editor/src/tags/object/Image/ImageEntity.js
yyassi-heartex: for sure - this could be refactored

web/libs/editor/src/utils/ImageCache.ts
yyassi-heartex: cleaned it up plus a bunch of other things in the file that were also redundant + improved clean up

PR #9318, opened 2026-02-04
fix: BROS-775: Show original urls in task code

makseq:
### Summary

This PR adds the ability to view original storage URLs (e.g., `s3://bucket/file.jpg`, `gs://...`) in the Task Source Viewer modal instead of resolved proxy URLs (`/tasks/.../resolve/?fileuri=...`). A new "Resolve URLs" toggle in the modal header allows users to switch between viewing modes. This is particularly useful for debugging storage configurations and understanding the actual data source paths.

### Changes

**Backend**: Extended the `/api/tasks/<id>/` endpoint to respect the `resolve_uri` query parameter (previously it was hardcoded to `true`). When `resolve_uri=false`, the API returns original storage URLs without converting them to proxy URLs. This aligns the detail endpoint behavior with the list endpoint (`/api/tasks/`) which already supported this parameter.

**Frontend**: Added a "Resolve URLs" toggle to the `ViewToggle` component alongside the existing Code/Interactive view switcher. The `TaskSourceViewer` component now manages this state, persists the user preference to localStorage, and reloads task data when the toggle changes. By default, the toggle is OFF to show original URLs, making it easier for users to inspect their storage configurations.

Bonus: Fixed very annoying behavior for modals in web/libs/ui/src/lib/modal/ModalPopup.tsx when a mouse click occurred inside a modal and the mouse was released outside the modal; it caused the modal to close, so you could not select text normally.

---

image

image

### DIFF
--- a/web/libs/datamanager/src/components/Common/TaskSourceViewer/TaskSourceViewer.tsx
+++ b/web/libs/datamanager/src/components/Common/TaskSourceViewer/TaskSourceViewer.tsx
@@ -1,4 +1,4 @@
-import { type ChangeEvent, type FC, useEffect, useState, useCallback, useRef } from "react";
+import { type ChangeEvent, type FC, useEffect, useState, useCallback } from "react";
 import { JsonViewer, type FilterConfig, Toggle } from "@humansignal/ui";
 import { FF_LOPS_E_3, FF_INTERACTIVE_JSON_VIEWER, isFF } from "../../../utils/feature-flags";
 import { CodeView } from "./CodeView";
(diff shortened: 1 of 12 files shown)


web/libs/datamanager/src/components/Common/TaskSourceViewer/TaskSourceViewer.tsx
```diff
@@ -64,12 +70,18 @@ export const TaskSourceViewer: FC<TaskSourceViewerProps> = ({
   const isInteractiveViewerEnabled = isFF(FF_INTERACTIVE_JSON_VIEWER);
 
   const [taskData, setTaskData] = useState(content);
+  const isInitialLoad = useRef(true);
```
hlomzik: This ref is not used

hlomzik (approved):

PR #9349, opened 2026-02-06
fix: FIT-1375: Prevent FSM leakage in update_task_states

bmartel:
### Problem
\`update_tasks_states\` was unconditionally calling FSM-related state updates (\`update_project_state_after_task_change\`), even when executed in a context where FSM is explicitly disabled (e.g., specific rqworker jobs). This caused \`FieldError\` ("Cannot resolve keyword 'current_state' into field") because the necessary fields/context were not available.

### Solution
Added a guard clause \`if CurrentContext.is_fsm_enabled():\` in \`label_studio/projects/models.py\`.

### Verification
- Verified that the call is now guarded.

### DIFF
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -543,8 +543,9 @@ class Project(ProjectMixin, FsmHistoryStateModel):
 
         if tasks_number_changed:
             # FSM: Recalculate project state after task deletion or import
-            user = CurrentContext.get_user()
-            update_project_state_after_task_change(self, user=user)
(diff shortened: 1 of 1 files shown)


bmartel:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21762918304)
> Successfully merged:  3 files changed, 41 insertions(+), 15 deletions(-)

matt-bernstein (approved):

PR #9357, opened 2026-02-09
fix: BROS-731: Support pointStyle

nick-skriabin:
This PR adds support for `pointStyle` attribute in Vector (rectangle, circle).

### DIFF
--- a/web/libs/editor/src/components/KonvaVector/components/VectorPoints.tsx
+++ b/web/libs/editor/src/components/KonvaVector/components/VectorPoints.tsx
@@ -1,5 +1,5 @@
 import type React from "react";
-import { Circle } from "react-konva";
+import { Circle, Rect } from "react-konva";
 import type Konva from "konva";
 import type { BezierPoint } from "../types";
 import { HIT_RADIUS } from "../constants";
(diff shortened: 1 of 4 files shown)


web/libs/editor/src/components/KonvaVector/components/VectorPoints.tsx
```diff
@@ -22,6 +22,7 @@ interface VectorPointsProps {
   pointStroke?: string;
   pointStrokeSelected?: string;
   pointStrokeWidth?: number;
+  pointStyle?: "circle" | "rectangle";
```
yyassi-heartex, line 25: minor nit-pick - lets switch ` "circle" | "rectangle";` to an enum or a type so we can reuse it that way instead of a string.

web/libs/editor/src/components/KonvaVector/components/VectorPoints.tsx
```diff
@@ -42,6 +43,7 @@ export const VectorPoints: React.FC<VectorPointsProps> = ({
   pointStroke = "#3b82f6",
   pointStrokeSelected = "#ffffff",
   pointStrokeWidth = 2,
+  pointStyle = "circle",
```
yyassi-heartex, line 46: we can then reuse it here

yyassi-heartex (approved):
we should use an enum or a type and reuse it rather than strings every time

nass600 (approved):

PR #9359, opened 2026-02-09
fix: FIT-720: Avoid loading all annotations in LabelStream when the user is admin and goes to previous tasks

yyassi-heartex:
This pull request introduces a feature for lazy loading annotation data in the Data Manager, significantly improving performance for tasks with many annotations. The main changes add support for lightweight annotation stubs, backend feature flag control, frontend hydration of full annotation data on selection, and robust cache invalidation for annotation updates. The implementation is gated behind a feature flag (`fflag_fix_all_fit_720_lazy_load_annotations`) and includes test fixtures to enable or disable the flag for testing.

**Backend changes for lazy loading annotations:**

* Added `AnnotationStubSerializer` in `label_studio/tasks/serializers.py` to provide lightweight annotation metadata when the feature flag is enabled, minimizing data sent for annotation lists.
* Updated `DataManagerTaskSerializer` and related serializers to use stub mode when the feature flag and request parameter are set, including logic to select the appropriate serializer and control expansion behavior. [[1]](diffhunk://#diff-7ce60240b37e69943356ab25fe22bef989b3eed3066f58372bdfd8edb029b6dbL442-R443) [[2]](diffhunk://#diff-7ce60240b37e69943356ab25fe22bef989b3eed3066f58372bdfd8edb029b6dbR519-R547) [[3]](diffhunk://#diff-2052cfb5bd8bfc2fd3ee34d9b4dc7b100e4c2e16b2a48356384380a21fe9b8aaR317-R327) [[4]](diffhunk://#diff-2052cfb5bd8bfc2fd3ee34d9b4dc7b100e4c2e16b2a48356384380a21fe9b8aaL337-R347) [[5]](diffhunk://#diff-3829314442f71e22038252a7dd1b60e33066aa55ad7462fed59a398ad7ecb820L874-R934)
* Added pytest fixtures to easily enable or disable the feature flag for testing.

**Frontend integration for annotation hydration and cache management:**

* Modified `LSFWrapper` in `web/libs/datamanager/src/sdk/lsf-sdk.js` to hydrate stub annotations by fetching full data from the backend when selected, ensuring seamless user experience. [[1]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595L379-R394) [[2]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595L448-R463) [[3]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R511-R573) [[4]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R630-R631) [[5]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595L695-R773) [[6]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595L865-R959) [[7]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R1158-R1191)
* Implemented cache invalidation for annotation and distribution data after annotation submission or update, ensuring the frontend always reflects the latest state. [[1]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R856-R865) [[2]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R899-R904)
* Debounced annotation selection events to prevent redundant hydration calls during batch operations, improving performance and stability.

**API and feature flag plumbing:**

* Added new API endpoint for fetching a single annotation by ID to support hydration.
* Updated feature flag integration and imports to support the new lazy loading flag. [[1]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R7-R12) [[2]](diffhunk://#diff-679489a404ba3dd06905287db35ee48bccafc5cc6a61567ef7531cb785012595R21-R34)

These changes collectively enable efficient handling of large annotation lists by sending minimal data initially and loading full annotation details only when needed, controlled via a feature flag for gradual rollout.


web/libs/datamanager/src/utils/feature-flags.js
```diff
@@ -50,6 +50,12 @@ export const FF_BETTER_USER_FILTER = "fflag_root_45_better_user_filter";
 export const FF_DISABLE_GLOBAL_USER_FETCHING =
   "fflag_all_feat_utc_204_users_performance_improvements_in_dm_for_large_orgs";
 
+/**
+ * Lazy load annotations in LabelStream to improve performance for tasks with many annotations
+ * @link https://labelstudio.atlassian.net/browse/FIT-720
+ */
+export const FF_FIT_720_LAZY_LOAD_ANNOTATIONS = "fflag_fix_all_fit_720_lazy_load_annotations";
```
yyassi-heartex: remove the FF from here - have any usages that use it from core

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21950709245)
> Successfully merged:  create mode 100644 web/libs/core/src/lib/utils/ImageCache.test.ts

web/libs/core/src/hooks/useResolveUser.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // Deduplicate in-flight requests
+  if (pendingFetches.has(userId)) {
+    return pendingFetches.get(userId)!;
+  }
+
+  const fetchPromise = (async () => {
+    try {
+      const response = await fetch(`/api/users/${userId}`);
```
yyassi-heartex: use tanstack query for this - fetch once and use cache after if refetching

web/libs/datamanager/src/sdk/lsf-sdk.js
```diff
@@ -17,6 +17,20 @@ import { annotationToServer, taskToLSFormat } from "./lsf-utils";
...
 import { imageCache } from "@humansignal/core";
 
+// FIT-720: Import cache invalidation functions
+let invalidateAnnotationCache = null;
+let invalidateDistributionCache = null;
+
+// Dynamically import from editor to avoid circular dependency
+// This is safe because the functions will be available when the app is fully loaded
+try {
+  const editorHooks = require("@humansignal/editor/src/hooks/useAnnotationQuery");
```
yyassi-heartex: lets see about require at the top - why do we do this?

web/libs/editor/src/components/InteractiveOverlays/CommentsOverlay.tsx
```diff
@@ -166,23 +165,15 @@ const CommentsOverlayInner = observer(({ annotation, commentStore }: CommentsOve
...
-  useEffect(() => {
-    loadComments();
-    // id is internal id,
-    // always different for different annotations, even empty ones;
-    // remain the same when user submit draft, so no unneeded calls.
-  }, [commentStore.annotation?.id]);
+  // FIT-720: CommentsOverlay no longer fetches comments eagerly.
+  // The Comments panel will load comments when the user clicks on the comments tab.
+  // CommentsOverlay just renders markers for whatever comments are already loaded.
+  // This prevents duplicate/early requests.
```
hlomzik, line 171: Are we sure about this? This way comments can easily get ignored because they are not visible. And we don't show a number of comments anywhere in annotation, making it even more risky.

bmartel: Yea this one I think requires some more investigation, because this overlay is the one that puts the comment markers over various regions. If it doesn't load the comments data, the comment markers wont display until the tab is focused, based on this PR.

yyassi-heartex: valid concerns, we have comment_count coming back for the stubs, if the count is > 0 we show the icon in the tab - unless there is another spot we need to show something that im missing?

[screen-recorder-fri-feb-13-2026-08-48-51.webm](https://github.com/user-attachments/assets/ceeb2686-864d-4d96-b34b-14c9cc8beb57)

hlomzik: But is this decision supported by product team?

hlomzik:
In general this PR looks enormous and it changes a lot of fundamental things. Too big for the issue it's solving. For example `lsf-sdk` is one of the most fragile and most cumbersome places we have and would be great to rewrite and improve it and not add more complexity there...

web/libs/editor/src/components/InteractiveOverlays/CommentsOverlay.tsx
bmartel: The thing I was referring to is the overlay comment markers. If you click on the comment region button to link to a bounding box, submit a comment that is targeting that bounding box, then reload the page where the comment tab is not open. What happens?

yyassi-heartex: @hlomzik 
> But is this decision supported by product team?

> In general this PR looks enormous and it changes a lot of fundamental things. Too big for the issue it's solving. For example lsf-sdk is one of the most fragile and most cumbersome places we have and would be great to rewrite and improve it and not add more complexity there...

I hear you, there is alot of changes here and it would seem daunting - which is why its important to be thorough in the review and test it (which is what we're doing)

we have a ticket FIT-720 which is in the support queue which triggered all these changes + we have had many reports of performance degradation from many customers over time and from looking into this issue its partially because we're loading everything when the task is loaded - it causes a bigger payload to be retrieved, every annotation history to be fetched and every element to be rendered before the user has even interacted with the task (beyond selecting it) this puts an un-necessary load on the server (fetching things the user may never access),  on memory (keeping un-necessary things that user may never touch) and browser time to render as it would be rendering way too many things as a result.

while its valid that lsf-sdk is fragile and there are a lot of changes in this PR - this isnt an indicator that we shouldn't touch it, its an indicator that we should fix it and make it better - a refactor is surely coming in the future but we have to take steps to make things better. I see your concerns and I want to address them - if you like, lets huddle and go over each of these concerns and test it together to make sure we're all set.

FTR, this is the 4th related PR in this ticket, I've been trying to deliver it in smaller pieces but this last chunk is very inter-dependant on the other changes and breaking it down further would take away from the user experience that it would deliver.

to roughly outline the changes:
- annotation stub + stub hydration
- reducing the things we request on task load
- lazy loaded comments and task history so we dont flood the users browser and the server
- adding skeleton loaders as things get fetched (especially in compare mode)
- improved task source modal support for large tasks with many annotations/predictions
- selecting an annotation from compare mode will scroll the annotation button into visibility

yyassi-heartex: @bmartel @hlomzik I just pushed an update that resolves the issue - great catch!

[screen-recorder-fri-feb-13-2026-10-30-23.webm](https://github.com/user-attachments/assets/ff385cd2-2fee-4547-ab60-a75ac7325cbe)

web/libs/core/src/hooks/useResolveUser.ts
yyassi-heartex: now using tanstack query 

[screen-recorder-fri-feb-13-2026-10-43-23.webm](https://github.com/user-attachments/assets/5ce15625-ebb6-4891-8ae8-7b3d583d8f5b)

web/libs/datamanager/src/sdk/lsf-sdk.js
bmartel: This hook needs to be moved to core, we are importing from LSF -> DM, which is possible to be circular, but more importantly its confusing.

yyassi-heartex: 100%, in process of doing that now

yyassi-heartex: done, please take another peak

label_studio/data_import/api.py
```diff
@@ -600,6 +600,13 @@ def _create_legacy(self, project):
                         f'from project {project} tasks'
                     )
 
+            # Strip FSM state field from predictions before validation
+            # Exported data may include 'state' which is not a valid prediction field
+            if flag_set('fflag_feat_fit_568_finite_state_management', user=self.request.user) and flag_set(
```
yyassi-heartex: @bmartel pytests were failing and claude is recommending we pop state here

label_studio/tasks/serializers.py
```diff
@@ -568,6 +568,14 @@ def add_predictions(self, task_predictions):
                     validation_errors.append(f'Task {i}, prediction {j}: Prediction must be a dictionary')
                     continue
 
+                # Strip FSM state field from predictions before validation
+                # Exported data may include 'state' which is not a valid prediction field
+                ff_user = self.project.organization.created_by
```
yyassi-heartex, line 582: @bmartel and here

label_studio/data_import/api.py
bmartel: @matt-bernstein just merged changes surrounding this.

bmartel (approved):
I don't know about the need to change the data_import checks in this PR, they shouldn't be necessary as I believe this change, but the actual core changes here are really impressive results. 🔥

label_studio/data_import/api.py
```diff
@@ -603,6 +603,13 @@ def _create_legacy(self, project):
...
+                'fflag_feat_fit_710_fsm_state_fields', user=self.request.user
+            ):
+                item.pop('state', None)
+
```
matt-bernstein:
```suggestion
```

fixed here https://github.com/HumanSignal/label-studio-enterprise/pull/10144

PR #9371, opened 2026-02-10
feat: BROS-782: Add `darkMode` getter to handle theme changes

hlomzik:
Now it's accessible via `Htx.darkMode` or through reaching `AppStore`.
It will change dynamically after theme change.


hlomzik:
We don't need this right now

PR #9379, opened 2026-02-11
fix: UTC-567:  can be configured with a value exceeding

nass600:
Fixed validation for playback speed related video attirbutes:
1. Added 10 as max value for both `minPlaybackSpeed` and `defaultPlaybackSpeed`
2. `defaultPlaybackSpeed` cannot be lower than `minPlaybackSpeed`

Screenshot 2026-02-11 at 12 00 45


nass600:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/21908681483)

PR #9396, opened 2026-02-13
fix: FIT-1392: exported annotations cannot be reimported as predictions

matt-bernstein:
strip state field in all prediction import paths


bmartel (approved):
Nice catch!

PR #9441, opened 2026-02-20
fix: BROS-803: Disallow deleting last/all points of the shape

nick-skriabin:
This pull request introduces safeguards to prevent users from deleting all points in a vector region, ensuring that at least one point always remains. This helps maintain data integrity and avoids empty or invalid vector regions.

Validation improvements for point deletion:

* Added a check in `KonvaVector.tsx` to prevent deletion of all points at once, ensuring at least one point remains in the region.
* Updated `deletePoint` in `pointManagement.ts` to block deletion when only one point is left, preventing an empty region.

### DIFF
--- a/web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
+++ b/web/libs/editor/src/components/KonvaVector/KonvaVector.tsx
@@ -2013,6 +2013,9 @@ export const KonvaVector = forwardRef<KonvaVectorRef, KonvaVectorProps>((props, 
 
       if (indicesToDelete.length === 0) return;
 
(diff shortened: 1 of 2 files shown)


yyassi-heartex (approved):
this will work, but im wondering if there should be some sort of visual feedback to indicate to the user why the action they're making didn't work as expected (maybe a toast message or something subtle)

nick-skriabin:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22309865279)
> Successfully merged:  11 files changed, 342 insertions(+), 74 deletions(-)

PR #9460, opened 2026-02-23
fix: UTC-574: Checkbox selection persists in 'Select Members' dropdown after clearing search

nass600:
Added select all capabilities to the select component


web/libs/ui/src/lib/select/select.module.scss
```diff
@@ -99,36 +99,50 @@
...
 
-  &:focus {
-    outline: 2px solid var(--color-primary-border);
-    outline-offset: -2px;
+    &:hover {
+      background: var(--color-primary-background-hover);
+      border-radius: var(--border-radius-medium);
+    }
+
+    &:focus {
```
ricardoantoniocm:
```suggestion
    &:focus-visible {
```

ricardoantoniocm (approved):
Approved with suggestion so that the focus outline is only shown when navigating to element using keyboard and not the mouse.

yyassi-heartex (approved):

nass600:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22321876428)

nass600:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22342739244)

PR #9461, opened 2026-02-23
fix: FIT-1428: Extra line is displayed in annotation tab

ricardoantoniocm:
This pull request makes a minor visual adjustment to the `AnnotationButton` component. The height of the top border area has been slightly increased to avoid a visual artifact that occurs when the top border radii are the same or higher than the height of the element.

### Before
image

### After
image

### DIFF
--- a/web/libs/editor/src/components/AnnotationsCarousel/AnnotationButton.scss
+++ b/web/libs/editor/src/components/AnnotationsCarousel/AnnotationButton.scss
@@ -111,7 +111,7 @@
     top: -1px;
     left: 0;
     width: 100%;
-    height: 8px;
+    height: 10px;
     border-top-style: solid;
     border-top-width: 2px;
     border-top-left-radius: var(--corner-radius-small);


hlomzik (changes requested):
This bottom border is pretty weird, it should not be there in a first place, better to try to eliminate it completely.
Also would not this change of height change the curvature of the border? It's visible on a screenshot

hlomzik:
I found a small hack for `::before` element:
```css
border-bottom-right-radius: 0.1px;
border-bottom-left-radius: 0.1px;
```

ricardoantoniocm:
> This bottom border is pretty weird, it should not be there in a first place, better to try to eliminate it completely. Also would not this change of height change the curvature of the border? It's visible on a screenshot

@hlomzik 

That is the whole purpose of this change, to remove this weird border.

The top corners radii are 0.5rem / 8px. The bottom border is not intentional, it is a visual artifact that appears on some zoom levels when the border radius equals the height of the container. There is no actual bottom border to remove as there is no bottom border set.

Changing the pseudo element container's height to 10px cannot affect the top corners radii since it is only 8px.

ricardoantoniocm:
> I found a small hack for `::before` element:
> 
> ```css
> border-bottom-right-radius: 0.1px;
> border-bottom-left-radius: 0.1px;
> ```

@hlomzik are you suggesting that we apply this hack instead of simply adjusting the height of the pseudo element slightly to avoid this visual artifact?

The other ::after pseudo element (applied for draft states) already has a height of 10px so technically this would introduce consistency to both pseudo elements, at least code-wise.

nass600 (approved):

hlomzik:
@ricardoantoniocm but they are shifted. and they are not aligned anyway, which is weird, because they should. nevermind, it's very subtle, approving!

hlomzik (approved):

ricardoantoniocm:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22316062993)

ricardoantoniocm:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22320257818)

ricardoantoniocm:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22320850034)

PR #9471, opened 2026-02-25
fix: UTC-590: Fix side issues after being paused

hlomzik:
Paused users receive 403 responses for all annotation-related API requests. The LSE `ApiProvider` detects these via its `onError` handler — it checks for `status === 403` with `display_context.reason === "PAUSED"` and shows a blocking "You are paused in this project" modal. However, the DataManager code that calls these APIs didn't account for error responses, assuming it always receives valid data back. This caused JavaScript errors to appear behind the pause modal.

**Fixed in this PR:**

- `loadNextTask`: When `next_task` returns 403, the error response was processed as task data by `applyTaskSnapshot`, causing `TypeError: Cannot read properties of undefined (reading 'queue')`. Added an early return for error/non-200 responses.
- `loadTaskHistory`: When `taskHistory` returns 403, `.map()` was called on the error object instead of an array, causing `TypeError: l.map is not a function`. Added an `Array.isArray` guard.
- `invokeAction`: The `next_task` call had no `errorHandler`, so the DM's global error handling would also set `serverError` (showing an "Errors occurred" indicator in the toolbar). Added `errorHandler` passthrough from `invokeAction` to `apiCall`, and used it in `loadNextTask` to suppress `PAUSED`/`OVERLAP_REACHED` errors that are already handled at the app level.

**Screenshots**

`TypeError: Cannot read properties of undefined (reading 'queue')` error in background
WarningAnnPaused

`TypeError: l.map is not a function` in console
Screenshot 2026-02-24 at 17 13 13


yyassi-heartex (approved):

yyassi-heartex:
/fmt
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22417243206)

hlomzik:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22419804918)
> Successfully merged:  create mode 100644 docs/themes/v2/source/images/analytics/overview.png

PR #9477, opened 2026-02-26
docs: Vector tags

caitlinwheeless:
Updating the vector tag docs for GA


docs/source/includes/tags/vector.md
```diff
@@ -14,7 +14,6 @@
 | [smartOnly] | <code>boolean</code> |  | Only show smart tool for interactive pre-annotations |
 | [snap] | <code>pixel</code> \| <code>none</code> | <code>none</code> | Snap vector to image pixels |
 | [closable] | <code>boolean</code> | <code>false</code> | Allow closed shapes |
-| [curves] | <code>boolean</code> | <code>false</code> | Allow Bezier curves |
```
caitlinwheeless, line 17: Removing this here is not how this should be done - I'm just doing this quickly so that the docs are updated for the 2.33 release notes. However, this parameter really should be removed from the tag source page in the code.

nick-skriabin (approved):

PR #9483, opened 2026-02-26
fix: FIT-1462: Annotation data fetch when we have less annotations than available space in compare mode may not be firing as expected

yyassi-heartex:
This pull request addresses issues with annotation hydration and panel loading in the grid component, particularly for cases with a small number of annotations (such as compare-all scenarios). The changes ensure that stub hydration is triggered reliably, fixing scenarios where the second panel would not load. Additionally, the logic for choosing between virtualized and non-virtualized grid rendering has been updated, and corresponding tests have been adjusted.

before:

[screen-recorder-thu-feb-26-2026-15-09-39.webm](https://github.com/user-attachments/assets/c0b4f104-02f8-41b7-81fc-d069e6fe7ee3)

after:

[screen-recorder-thu-feb-26-2026-15-08-58.webm](https://github.com/user-attachments/assets/a6ebb463-80ba-4f0c-9510-4bd1927abf6b)

Grid rendering and hydration improvements:

* Updated the `VirtualizedGrid` component to hydrate all annotation stubs when the annotation count is small, ensuring that all panels load reliably even when only two annotations are present.
* Changed the logic in the `Grid` wrapper to always use `VirtualizedGrid` when the feature flag is enabled, regardless of annotation count, so that stub hydration runs in all cases (fixes compare-all scenarios).

Testing updates:

* Modified the test case to verify that `VirtualizedGrid` is rendered when the feature flag is enabled, including cases with few annotations, reflecting the new logic.


web/libs/editor/src/components/App/Grid.jsx
```diff
@@ -640,11 +645,10 @@ class GridClassComponent extends Component {
...
 export default function Grid(props) {
-  const { annotations } = props;
-  const visibleCount = annotations.filter((c) => !c.hidden).length;
-
-  // FIT-720: Use virtualization when FF is enabled AND there are many annotations
-  const shouldVirtualize = isFF(FF_FIT_720_LAZY_LOAD_ANNOTATIONS) && visibleCount > VIRTUALIZATION_THRESHOLD;
+  // FIT-720: Use VirtualizedGrid when FF is on so stub hydration runs (fixes compare-all with
+  // 2 annotations where the second panel would never load in GridClassComponent).
+  // Virtualization (react-window) is still only a win when we have many annotations.
+  const shouldVirtualize = isFF(FF_FIT_720_LAZY_LOAD_ANNOTATIONS);
```
bmartel, line 644: What happens if we remove the `visibleCount > VIRTUALIZATION_THRESHOLD` check here as you have it, but leave the other checks completely alone. Does this solve the problem as well? I am thinking that `VIRTUALIZATION_THRESHOLD` might just be completely not needed. I would pull it all out and check.

yyassi-heartex: you're right - doesn't look like its needed - committing the update

bmartel (approved):

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22464590195)
> Successfully merged:  create mode 100644 web/libs/ui/src/lib/badge/badge.module.scss

PR #9488, opened 2026-02-27
feat: BROS-825: Improve UX of DM filters

makseq:
### Summary

This PR addresses several long-standing UX pain points in the Data Manager filter panel: remembering recently used fields, reducing repetitive input when switching columns, and enabling filter configuration sharing via clipboard.

image

---

### 1. Recently Used Fields

When opening the column selector dropdown, the last 3 fields used in the current project now appear at the top under a **"RECENT"** header, separated from the full **"ALL FIELDS"** list below.

Selecting a recent field restores its full state — column, operator, and value — exactly as it was when last used. The recents list is stored in `localStorage` per project, so it persists across page reloads. When you select a field from the "Recent" section, the list order stays stable; it only reshuffles when you pick a new (non-recent) column. As you work with filters, their latest operator/value state is continuously saved in the background so recents always reflect your most recent configuration.

### 2. Smart State Preservation on Column Switch

Previously, switching a filter's column would reset the operator and value to defaults, losing user input. Now the system attempts to carry over the current operator and value to the new column. If the operator is compatible with the new column type (e.g. both are strings), it is preserved along with the value. If the types are incompatible, only then does it fall back to defaults. This minimizes repetitive re-entry when exploring different columns.

### 3. Auto-Copy Settings on New Filter

When adding a new filter via "Add Another Filter", the column and operator are automatically copied from the last existing filter. This eliminates the need to re-select the same column and operator when building multiple filters on the same field, which is a common workflow.

### 4. Copy / Paste / Undo Filters

Three new buttons appear in the filter toolbar (right-aligned):

- **Copy** — serializes all current filters (including partially filled ones) to JSON and copies to clipboard. The format is compatible with the Label Studio SDK, so users can paste it directly into API scripts.
- **Paste** — reads JSON from clipboard, validates the format and column IDs against the current project, and imports the filters. Shows specific toast error messages for invalid JSON, wrong format, or unrecognized columns.
- **Undo** — appears only after a successful paste. Restores the filter state from before the paste. Disappears after clicking or after the next paste.

### 5. Dropdown UI Polish

The column selector dropdown received several visual improvements:
- **Auto-sizing width** — the dropdown expands to fit long column names (capped at 500px) instead of truncating them.
- **Taller list** — shows 8 items instead of 5 before scrolling, via a new `virtualListMaxVisible` prop on the shared `Select` component.
- **Compact decorative rows** — the "RECENT" header, separator line, and "ALL FIELDS" header use custom per-item heights (8–36px instead of the default 40px) via a new `height` option field, so they don't waste vertical space.

---

### Changed Files

| Area | Files | What changed |
|------|-------|-------------|
| Filters UI | `Filters.jsx`, `Filters.scss` | Recents state, dropdown option assembly, copy/paste/undo handlers, action bar layout, cursor fix |
| Filter row | `FilterLine.jsx` | Column switch logic (recent vs non-recent), custom option renderer, continuous recents tracking via `useEffect` |
| Filter helpers | `filter-helpers.js` *(new)* | Extracted `filterFieldSearchHandler` and `findSelectedOption` for testability |
| Recents persistence | `filter-recents.js` *(new)* | `localStorage` read/write with full state `{ id, operator, value }`, backward-compatible with old string format |
| Filter model | `tab_filter.js` | `setFilter` preserves compatible operator/value; new `setFilterFromRecent` restores full state |
| Tab model | `tab.js` | `createFilter` copies column+operator from last filter; `allFiltersSnapshot` getter; `importFilters` action |
| Snapshot validation | `filter_snapshot_utils.js` *(new)* | Extracted `validateFilterSnapshot` for testability |
| Dropdown component | `FilterDropdown.jsx` | Passes `virtualListMaxVisible={8}` to Select |
| Select component | `select.tsx`, `types.ts` | Per-item `height` in virtual list, `virtualListMaxVisible` prop, `max-h-none` fix for double scrollbar |

### Tests

44 new tests across 3 new test files:

- `filter-recents.test.js` (18 tests) — localStorage persistence, ordering, capping, backward compat
- `filter-helpers.test.js` (16 tests) — search handler visibility rules, option finder with flat/grouped items
- `tab.test.js` (10 tests) — snapshot validation: format checks, column matching, edge cases


nass600 (approved):

web/libs/datamanager/src/components/Filters/FilterLine/FilterLine.jsx
```diff
@@ -143,122 +298,5 @@ export const FilterLine = observer(({ filter, availableFilters, index, view, sid
         )}
       </div>
     );
-  }
-
-  // Main layout uses parent grid structure - render children as direct grid items
```
hlomzik, line 148: we need to instruct agents to NOT remove explanatory comments

PR #9495, opened 2026-02-27
fix: BROS-827: Performance optimization for task switching

makseq:


web/libs/datamanager/src/stores/AppStore.js
```diff
@@ -211,17 +212,22 @@ export const AppStore = types
 
       self.setLoadingData(true);
 
+      // Yield to browser so loading indicator paints before heavy store operations
+      yield new Promise((resolve) => requestAnimationFrame(() => requestAnimationFrame(resolve)));
```
bmartel, line 216: Time slicing!!!!! This is key to breaking down the long pauses in the Mobx work that would trigger re-renders and lock the browser for a long time 🔥

bmartel (approved):
This looks great 😎

web/libs/datamanager/src/stores/AppStore.js
makseq: wow!

yyassi-heartex (approved):

PR #9511, opened 2026-03-03
fix: FIT-1387: DM Checkboxes display the checkbox when unchecked

ricardoantoniocm:
This pull request makes a minor change to the checkbox styles by removing the rule that displayed the check icon when the input was checked. This simplifies the styling and may affect how the checkbox appears when selected.

### Before
image
image

### After
image

image


bmartel (approved):
Seems reasonable to me, I wonder why we had this duplicate handling path, seems possible we changed the structure and state management of Checkboxes which yielded this not required, or it was an accidental inclusion in the first place.

nass600 (approved):

PR #9521, opened 2026-03-04
fix: FIT-1482: Resolve BrushRegion memory leak and highlight lag

bmartel:
**Problem**: The BrushRegion tool suffers from a significant memory leak and UI degradation when many mask regions are created (e.g., ~100 mask regions). The issue degrades browser memory (up to 4+ GB) and causes severe UI lag (panning, zooming, drawing).

**Solution**:
1. **Remove Nested Konva `<Layer>` Components**: Change the wrapping `<Layer>` elements in `HtxBrushView` to `<Group>` to prevent creating new canvases.
2. **Remove `toDataURL()` and Highlight Caching**.
3. **Refactor BBox Calculation** to use `self.touches` instead of `getImageData`.
4. **Optimize Serialization** to use cached RLE directly when `self.touches.length === 0`.
5. **Fix Highlight Rendering** to use opacity instead of Konva shadow.

**Files changed:**
- `web/libs/editor/src/regions/BrushRegion.jsx`
- `web/libs/editor/src/components/ImageView/ImageView.jsx`
- `web/libs/editor/src/regions/__tests__/BrushRegion.test.jsx`
- `web/libs/editor/src/regions/__tests__/BrushRegion.test.js`

**Manual verification:**
- Created large number of mask regions and confirmed memory stays flat (~450MB) and UI doesn't lag.
- Verified updating regions is performant and doesn't block UI.
- Verified highlights apply successfully via opacity.

### DIFF
--- a/web/libs/editor/src/LabelStudio.tsx
+++ b/web/libs/editor/src/LabelStudio.tsx
@@ -177,20 +177,13 @@ export class LabelStudio {
          */
         this.store.selfDestroy();
       }
-      destroy(this.store);
-      Hotkey.unbindAll();
 
       window.Htx = null;
-
-      if (isFF(FF_LSDV_4620_3_ML)) {
-        /*
-            ...
-            as well as nulling all these this.store
-         */
-        this.store = null;
-        this.destroy = null;
-        LabelStudio.instances.delete(this);
-      }
+      destroy(this.store);
+      Hotkey.unbindAll();
+      this.store = null;
+      this.destroy = null;
+      LabelStudio.instances.delete(this);
     };
   }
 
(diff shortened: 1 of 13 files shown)


web/libs/editor/src/LabelStudio.tsx
```diff
@@ -177,17 +177,13 @@ export class LabelStudio {
          */
         this.store.selfDestroy();
       }
+
+      window.Htx = null;
       destroy(this.store);
       Hotkey.unbindAll();
-      if (isFF(FF_LSDV_4620_3_ML)) {
```
yyassi-heartex, line 182: oooh we're removing a FF check here - was that intended?

yyassi-heartex (approved):

web/libs/editor/src/LabelStudio.tsx
bmartel: yea this check was gating something very important which we needed, but only in part. There are a lot of things that FF is gating which in another cooldown type ticket we should outright remove the code and FF that remains, not retain any of whatever else is there. The other code is dangerous and actually can severely harm performance and could break the application.

PR #9527, opened 2026-03-05
fix: UTC-643: NX-related security vulnerabilities

nass600:


web/package.json
```diff
@@ -248,6 +248,15 @@
     "on-headers": "1.1.0",
     "form-data": "4.0.4",
     "qs": ">=6.14.1",
-    "stylus": "npm:empty-npm-package@1.0.0"
+    "stylus": "npm:empty-npm-package@1.0.0",
+    "rollup": ">=4.59.0",
+    "@isaacs/brace-expansion": ">=5.0.1",
+    "immutable": ">=5.1.5",
+    "axios": ">=1.13.5",
+    "lodash": ">=4.17.23",
```
yyassi-heartex, line 256: oh no - lodash again? all usages should be gone already

yyassi-heartex: ahh likely used by dependencies

bmartel:
All the build + test fails due to conflicting versions of these dependencies that were pinned. It will require some threading of the needle to get this across the line.

bmartel (approved):
One linting error, looks good to go if all builds/tests pass 🎉

bmartel (approved):

PR #9538, opened 2026-03-06
fix: FIT-1499: Resolve endless loading state for TimeSeries in Review Stream

bmartel:
**Problem**: A customer project is able to label/review using DataManager but the review stream is unable to access the Timeseries tag correctly, resulting in an endless loading state.

**Solution**: Fixed a bug in the `updateValue` for TimeSeries which was using async/await in a MST action which is not guaranteed to work correctly. It requires flow wrapped generator functions with yield on any async functions.


yyassi-heartex (approved):

bmartel:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22776905450)
> Successfully merged:  2 files changed, 22 insertions(+), 8 deletions(-)

yyassi-heartex (approved):

PR #9555, opened 2026-03-09
fix: BROS-841: fix lock for unfinished polygons

borisheartex:
Problem: When a user starts drawing a polygon (PolygonLabels) but does not finish it (leaving item.closed = false), clicking the "Lock region" button in the Regions panel causes the unfinished polygon to disappear instead of becoming locked and visible.

Root Cause: In web/libs/editor/src/regions/PolygonRegion.jsx, the Edges and renderCircles components (the only visible representation of an unfinished polygon) were conditionally rendered based on !item.isReadOnly(). Locking sets isReadOnly() to true, causing both elements to stop rendering. The filled Poly component only renders for closed polygons, so the polygon vanishes entirely.

Current quick solution to prevent the issue:
1. Updated RegionItem.tsx (Details Panel) - Added conditional rendering to hide both the lock/unlock and hide/unhide buttons when region.incomplete is true. This ensures these controls only appear for finished polygons.
2. Updated OutlinerTree.tsx (Outliner Panel) - Added conditional rendering to hide both the lock/unlock and hide/unhide buttons when item.incomplete is true, matching the behavior in the Details panel.
3. Updated Annotation.js hotkey handlers - Modified lockSelectedRegions() and hideSelectedRegions() methods to skip regions where incomplete === true. This prevents the "Lock Region" (Ctrl+L) and "Toggle Region Visibility" (Ctrl+H) hotkeys from affecting unfinished polygons.


borisheartex:
@cursoragent this fix doesn't work as expected. 
Change of approach: when polygon is in the 'unfinished' state, the lock/ unlock and hide/ unhide options should not be available. They just shouldn't be displayed to the user. But as soon as the polygon is finished, they need to appear and work as they did before this change. Make the requested change, making absolutely sure to only update the display of the two options, and that  this does NOT affect anything other than the unfinished polygon.

borisheartex:
@cursoragent the solution looks good.
Now that unfinished polygon cannot be hidden/ locked using the UI buttons, make sure that user can't do the same using hotkeys. The 'Toggle Region Visibility' and 'Lock Region' hotkeys should not toggle unfinished polygons. Again, make sure this change affects the unfinished polygons exclusively, and nothing else.

hlomzik (approved):

borisheartex:
@cursoragent remove the new tests added in this PR. We don't need them.

borisheartex:
/fm merge
> Command | Description
> --- | ---
> /fm sync <Optional branch override> | Sync upstream prs and merge with pull request base

PR #9560, opened 2026-03-09
fix: FIT-1500: fonts are getting loaded in from google

yyassi-heartex:
This pull request removes the local serving and usage of the Roboto font CSS from the project, switching away from both the self-hosted and Google-hosted Roboto font styles. The changes simplify static file handling and remove unnecessary font assets from the codebase.

Font asset and static file cleanup:

* Removed the entry for serving `roboto.css` from the static URL mapping in `label_studio/core/all_urls.json` and deleted the corresponding route from `label_studio/core/urls.py`. [[1]](diffhunk://#diff-a5903a47cd7871164704d1e04dab7c177ba6218ab0e25e5e97b48efa7228f3cdL44-L49) [[2]](diffhunk://#diff-6f771ffaa5d142672ce28a883fbf37dd7b7837a94f7bfbf209ba95b4450fbd5cL58-L61)
* Deleted the local `roboto.css` file and all its font-face definitions from `label_studio/core/static/fonts/roboto/roboto.css`.
* Removed the Google Fonts Roboto stylesheet link from the HTML head in `web/libs/datamanager/public/index.html`.


label_studio/core/static/fonts/space-grotesk/space-grotesk.css
bmartel, line 1: Why are we adding this for? Wasn't the problem with Roboto getting loaded from CDN?

yyassi-heartex: we have a bug report for a customer with infrastructure that does not allow outbound egress to Google APIs. figured this would avoid the problem altogether

bmartel: But the font here is space-grotesk? Not Roboto?

yyassi-heartex: as discussed we'll be removing these font usages in favor of figtree to match fonts with the rest of the application

bmartel (approved):

PR #9562, opened 2026-03-09
fix: FIT-1502: Large labeling interface causing to freeze LS on Create Project modal

yyassi-heartex:
This pull request improves the user experience and reliability of the project configuration preview in the Label Studio app, especially for large configurations. The main focus is on making the preview update mechanism more robust, removing a now-unnecessary feature flag, and fixing issues with stale or empty previews after configuration changes. The changes also add a placeholder while the preview updates, and ensure the preview component is properly remounted and initialized.

**Preview update experience and reliability:**

* The manual preview update mode for large configs (≥200 tags) is now always enabled, regardless of feature flags. The preview only updates when the user clicks "Update Preview," preventing slow re-renders on every keystroke. [[1]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L32-R48) [[2]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L470-R508)
* The preview component (`Preview`) is now forcibly remounted with a new `key` whenever the displayed config changes, ensuring the Label Studio frontend always initializes with the new config and avoids showing stale or empty previews. [[1]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R541) [[2]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R726-R748)
* A placeholder is shown while the preview is updating, preventing flickers and race conditions where the old preview or empty state would show during loading. [[1]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R619-R621) [[2]](diffhunk://#diff-ff8a3990dde3c08403f325ed9c16889ed489c1bd97830d40ece9078165444a01R673-R681) [[3]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R726-R748)

**Code and dependency cleanup:**

* The now-unused feature flag logic (`ff` and `FF_PREVIEW_PERFORMANCE`) has been removed from the config page, simplifying the code. [[1]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L14-R14) [[2]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L32-R48) [[3]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89R449-R462) [[4]](diffhunk://#diff-b8cade5f083b7ee986ecfffd94abcf8500ea563ec194ccf3ca3d94f2edda8a89L470-R508)

**Preview component robustness:**

* The `Preview` component now ensures the correct DOM node is used for initialization after remounts, and tracks mounting state to avoid updating unmounted components, preventing errors and memory leaks. [[1]](diffhunk://#diff-c884ae2a78e98837ff6a1b4d873182c36b1eb476b7a9135c95349797d6b26e73L71-R81) [[2]](diffhunk://#diff-c884ae2a78e98837ff6a1b4d873182c36b1eb476b7a9135c95349797d6b26e73L107-R117) [[3]](diffhunk://#diff-c884ae2a78e98837ff6a1b4d873182c36b1eb476b7a9135c95349797d6b26e73L127-L135)

These changes collectively make the configuration preview more predictable and user-friendly, especially for users working with large or frequently changing configs.

### DIFF
--- a/web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
+++ b/web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
@@ -25,6 +25,7 @@ export const Preview = ({ config, data, error, loading, project }) => {
   const [storeReady, setStoreReady] = useState(false);
   const lsf = useRef(null);
   const rootRef = useRef();
+  const mountedRef = useRef(true);
   const api = useAPI();
   const projectRef = useRef(project);
   projectRef.current = project;
@@ -68,14 +69,16 @@ export const Preview = ({ config, data, error, loading, project }) => {
   }, [config]);
 
   const initLabelStudio = useCallback(async (config, task) => {
-    // wait for dependencies to load, the promise is resolved only once
-    // and is started when the component is mounted for the first time
     await loadDependencies();
 
     if (lsf.current || !task.data) return;
 
+    // Resolve root by id so we don't pass null after remounts; getElementById finds the current DOM node
+    const rootEl = document.getElementById("label-studio") ?? rootRef.current;
+    if (!rootEl) return;
+
     try {
-      lsf.current = new window.LabelStudio(rootRef.current, {
+      lsf.current = new window.LabelStudio(rootEl, {
         config,
         task,
         interfaces: ["side-column"],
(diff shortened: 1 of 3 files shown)


web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
```diff
@@ -68,14 +69,16 @@ export const Preview = ({ config, data, error, loading, project }) => {
   }, [config]);
 
   const initLabelStudio = useCallback(async (config, task) => {
-    // wait for dependencies to load, the promise is resolved only once
-    // and is started when the component is mounted for the first time
     await loadDependencies();
 
     if (lsf.current || !task.data) return;
 
+    // Resolve root by id so we don't pass null after remounts; getElementById finds the current DOM node
```
bmartel, line 76: I am not sure I quite follow this comment and below checks that use DOM directly, and fall back to the rootRef. Is this looking at a different element depending on the state of the tear down/ reinit?

bmartel (approved):

web/apps/labelstudio/src/pages/CreateProject/Config/Preview.jsx
yyassi-heartex: essentially yes. here is cursors response:
> Short answer: Yes. The code can end up using a different DOM node depending on teardown/remount. That’s intentional so we don’t pass null into Label Studio after a remount.
What’s going on
The same node is both the ref and the one with the id: at line 164 the div has id="label-studio" and ref={rootRef}. So under normal, synchronous rendering they refer to the same element.
The important case is when the component (or its parent) unmounts and then remounts, e.g.:
React 18 Strict Mode (mount → unmount → mount again), or
Parent re-renders and remounts Preview (e.g. key change).
Why use getElementById and then fall back to rootRef?
initLabelStudio is async (it await loadDependencies()). So there’s a gap between when the effect runs and when we actually need the root element.
After that gap:
Ref: If this instance was unmounted, React clears rootRef.current to null. So we’d pass null into new window.LabelStudio(rootEl, ...).
getElementById: The current DOM might already have a new <div id="label-studio"> from a new mount of Preview. So document.getElementById("label-studio") returns that current node, not the old one.
So we’re not always looking at the same element: we prefer the current node in the document (by id), and only use the ref when the component is still the one that owns the current DOM (ref is set and matches the same element). The fallback ?? rootRef.current is for environments where the id might not be present or we want to rely on the ref when it’s known to be correct.
Summary you can paste into the review
“We resolve the root by id first so that after a remount (e.g. Strict Mode or parent re-mounting Preview), we don’t pass null into Label Studio. The ref can be cleared when this instance unmounts, but getElementById('label-studio') returns the current div in the document, which may be the new instance’s node. So yes — we can be looking at a different element depending on teardown/reinit; we intentionally prefer the current DOM node so init always has a valid root when the element exists.”

PR #9565, opened 2026-03-09
fix: UTC-660: Matrices not deleted when predictions deleted

mcanu:


mcanu:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22910228770)

hakan458 (approved):

PR #9569, opened 2026-03-10
fix: BROS-843: Fix hotkey styles in tooltips

AndrejOros:
The Hotkey.Tooltip component was generating <kbd> elements with the class 'hotkey__key' (via cn('hotkey').elem('key')), but no CSS rule existed for that class. The actual key badge styles are defined under '.key-group__key' in App.scss.

Changes:
- Fix BEM class name in Hotkey.Tooltip from cn('hotkey').elem('key') to cn('key-group').elem('key') so <kbd> elements match existing CSS styles
- Add explicit '+' separator elements between keys in multi-key combos (e.g. Ctrl+Alt+Space) for clarity even without full CSS rendering
- Add .key-group__separator CSS for the plus sign between keys
- Add inline-flex and align-items to .key-group for proper alignment

Affected tooltips: Play/Pause (video/audio), Create Relation, Edit Region Meta, and all other buttons using Hotkey.Tooltip.

Before:
Playpause
Altm

After:

VideoPlayDark
VideoPlayLight
AltMFixDark

### DIFF
--- a/web/libs/editor/src/components/App/App.scss
+++ b/web/libs/editor/src/components/App/App.scss
@@ -91,6 +91,9 @@ body {
 }
 
 .key-group {
+  display: inline-flex;
+  align-items: center;
+
   &__key {
     padding: 0 4px;
     height: 16px;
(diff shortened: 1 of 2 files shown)


AndrejOros:
@cursoragent fix looks good, but the Plus (+) symbols are barely visible. Adjust them to be more pronounced in both Dark and Light mode tooltips.

AndrejOros:
@cursoragent remove Plus (+) sign from the hotkeys, but keep the style changes intact

hlomzik (approved):

PR #9581, opened 2026-03-10
fix: FIT-1496: Previously imported file reappears in Data Import modal upon reopening

yyassi-heartex:
This pull request improves the file upload import experience in Label Studio by ensuring that previously imported files are not shown when reopening the import modal, and by preventing caching of sensitive file upload lists. It also updates the frontend logic to correctly manage and reset the file upload list state, and adds backend tests for these behaviors.

**Backend logic and API behavior:**

* Updated `FileUploadListAPI` in `data_import/api.py` to return an explicit empty queryset when the `ids` parameter is empty, so that previously imported files are never shown on modal reopen. Also, added a `Cache-Control: no-store` header to prevent caching of file upload lists.

* Added tests in `test_api.py` to verify that an empty `ids` parameter returns an empty list and that the `Cache-Control` header is set correctly in responses.

**Frontend state management and API usage:**

* Improved the reducer in `Import.jsx` to handle a `reset` action (clearing all uploads and IDs) and a `setList` action (setting the uploaded files and their IDs in one step).

* Moved the call to `onFileListUpdate` into a `useEffect` that tracks `files.ids`, ensuring the parent is always notified of file ID changes.

* Updated the `loadFilesList` function to always send the `ids` parameter as `"[]"` if no IDs are present, matching backend expectations, and to use the new `setList` reducer action. [[1]](diffhunk://#diff-69c27f83b5c2e14a1e4fb411596068407ea0fbb9cbb20adfefb11deb122789e4R204-R217) [[2]](diffhunk://#diff-69c27f83b5c2e14a1e4fb411596068407ea0fbb9cbb20adfefb11deb122789e4R164-R174)

* Ensured that the file upload state is reset when the project changes, so old uploads do not persist between projects.


label_studio/data_import/api.py
```diff
@@ -818,10 +818,15 @@ def get_queryset(self):
...
         logger.debug(f'File Upload IDs found: {ids}')
+        if not ids:
+            # Explicit empty queryset so reopened import modal never shows previously imported files
+            return FileUpload.objects.none()
         return FileUpload.objects.filter(project_id=project.id, id__in=ids, user=self.request.user)
 
     def get(self, request, *args, **kwargs):
-        return self.list(request, *args, **kwargs)
+        response = self.list(request, *args, **kwargs)
+        response['Cache-Control'] = 'no-store'
```
bmartel, line 828: I don't know about setting this here like that, that shouldn't be required. I think the issue was the other change above.

yyassi-heartex: I couldn't replicate the problem so i decided to just harden the front and backend. cursor's reasoning for applying this specifically was:
> In environments that do cache this URL (browser, reverse proxy, CDN), a “reopen modal” request might be served from cache. The user could see an old list again (the original bug), even though the server now correctly returns [] for ids=[].

yyassi-heartex:
/git merge
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/22927086995)
> Successfully merged:  delete mode 100644 web/libs/editor/src/assets/styles/_variables.scss

PR #9588, opened 2026-03-11
fix: BROS-847: Disable Submit/Update/Fix & Accept button when polygon region is unfinished

AndrejOros:
**Reason for change**
Users could previously submit, update, or accept an annotation even when it contained an unfinished (unclosed) polygon region. This change disables the Submit, Update, and Fix & Accept buttons — as well as the Ctrl+Enter hotkey — when an annotation has any incomplete polygon.

**Scope**
The fix is intentionally minimal and localized:

- Only polygon regions are checked (area.type === "polygonregion"). Vector regions and all other region types are not affected.
- Only the active BottomBar UI is changed. Legacy/deprecated UI components are not touched.
- Only submission-related buttons (Submit, Update, Fix & Accept) and the submit hotkey are gated. Skip, Reject, and all other buttons remain unchanged.

**Changes (4 files)**
File - Change

- Annotation.js - Added hasIncompletePolygons computed view — returns true only if any area with type === "polygonregion" has incomplete === true
- BottomBar/Controls.tsx - Added hasIncompletePolygons to isDisabled for Submit/Update buttons; updated tooltips
- BottomBar/buttons.tsx - AcceptButton checks hasIncompletePolygons for disabled state and tooltip
- AppStore.js - Added early-return guard entity.hasIncompletePolygons in annotation:submit hotkey handler

**Before**

https://github.com/user-attachments/assets/77d2244e-68ce-453b-8263-be30490a0ab4

**After**

**Testing**

1. Start drawing a polygon but do not close it.
2. Verify Submit / Update / Fix & Accept buttons are disabled with tooltip "Complete all regions before submitting/accepting".
3. Verify Ctrl+Enter hotkey does nothing.
4. Close the polygon — buttons and hotkey should work again.
5. Verify other region types (bounding box, rectangle, etc.) do not trigger the disabled state.

**Risks**
Low. Changes are scoped to 4 files with no modifications to region models, legacy UI, or non-polygon logic.****

### DIFF
--- a/web/libs/editor/src/components/BottomBar/Controls.tsx
+++ b/web/libs/editor/src/components/BottomBar/Controls.tsx
@@ -250,7 +250,7 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
               : "Save results: [ Ctrl+Enter ]";
 
         buttons.push(
-          <ButtonTooltip key="submit" title={title}>
+          <ButtonTooltip key="submit" title={title} className="whitespace-nowrap">
             <div className={cn("controls").elem("tooltip-wrapper").toClassName()}>
               <ButtonGroup>
                 <Button
@@ -305,14 +305,13 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
               ? "No changes were made"
               : "Update this task: [ Ctrl+Enter ]";
         const button = (
-          <ButtonTooltip key="update" title={updateTitle}>
+          <ButtonTooltip key="update" title={updateTitle} className="whitespace-nowrap">
             <ButtonGroup>
               <Button
                 aria-label="submit"
                 name="submit"
                 className="w-[150px]"
                 disabled={isUpdateDisabled}
-                tooltip={updateTitle}
                 onClick={async (event) => {
                   if ((event.target as HTMLButtonElement).classList.contains(dropdownTrigger)) return;
                   const selected = store.annotationStore?.selected;
--- a/web/libs/editor/src/components/BottomBar/Controls.tsx
+++ b/web/libs/editor/src/components/BottomBar/Controls.tsx
@@ -250,7 +250,7 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
               : "Save results: [ Ctrl+Enter ]";
 
         buttons.push(
-          <ButtonTooltip key="submit" title={title} className="whitespace-nowrap">
+          <ButtonTooltip key="submit" title={title} className="whitespace-nowrap max-w-none">
             <div className={cn("controls").elem("tooltip-wrapper").toClassName()}>
               <ButtonGroup>
                 <Button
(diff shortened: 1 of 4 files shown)


AndrejOros:
@cursoragent Tooltip for Update button is missing- fix that. Also, update tooltips to be properly formatted, 'submitting' and 'accepting' part of the tooltip is displayed in a new row.

AndrejOros:
@cursoragent the following part is incorrect:
 Formatted tooltips with newlines - Created three separate tooltip constants with \n to display the action word on a new row:

INCOMPLETE_SUBMIT_TOOLTIP: "Complete all regions before\nsubmitting"
INCOMPLETE_UPDATE_TOOLTIP: "Complete all regions before\nupdating"
INCOMPLETE_ACCEPT_TOOLTIP: "Complete all regions before\naccepting"

Fix it so everything in the tooltip is displayed in one row.

AndrejOros:
@cursoragent The tooltip still displays the “submitting”, “updating”, and “accepting” labels on a separate row. Fix that - they should be aligned on the same row as “Complete all regions before”.

web/libs/editor/src/components/BottomBar/buttons.tsx
```diff
@@ -54,22 +56,26 @@ export const AcceptButton = memo(
...
-        onClick={async () => {
-          annotation.submissionInProgress();
-          await store.commentStore.commentFormSubmit();
-          store.acceptAnnotation();
-        }}
-        data-testid="bottombar-accept-button"
-      >
-        {hasChanges ? "Fix + Accept" : "Accept"}
-      </Button>
+      <Tooltip title={tooltip} disabled={!store.settings.enableTooltips} className="whitespace-nowrap">
```
hlomzik: @yyassi-heartex do we also need to adjust styles here?

yyassi-heartex: yes, we will - i've added max-w-none already but i haven't pushed it yet. because im looking to add something similar for update

web/libs/editor/src/components/BottomBar/Controls.tsx
```diff
@@ -291,13 +297,15 @@ export const Controls = controlsInjector<{ annotation: MSTAnnotation }>(
...
+        const updateTitle = hasIncompleteRegions
+          ? INCOMPLETE_UPDATE_TOOLTIP
+          : overlapDisabled
+            ? store.overlapReachedMessage
+            : noChanges
+              ? "No changes were made"
+              : "Update this task: [ Ctrl+Enter ]";
         const button = (
-          <ButtonTooltip key="update" title={updateTitle}>
+          <ButtonTooltip key="update" title={updateTitle} className="whitespace-nowrap">
```
hlomzik: And here?

web/libs/editor/src/components/BottomBar/buttons.tsx
yyassi-heartex: there it goes

hlomzik (approved):

hlomzik:
@yyassi-heartex Nevermind my comments, all good and approved!

PR #9589, opened 2026-03-11
fix: UTC-612: Improvement: 'Show all' button on user details is displayed/enabled even if there is only one tag

nass600:
## Summary
- Add `onClose` and `maxWidth` props to `Badge` component with ellipsis truncation and conditional tooltip (tooltip only appears when text is actually clipped)
- Add `badgeMaxWidth` prop to `BadgeGroup` to forward truncation to individual badges
- Add truncation + tooltip to the tag label column in `TagMembers` settings page
- Fix `TagAutocomplete` double-deletion bug when backspacing from 2 chars to 1 (closing the dropdown was clearing the query)
- Add Storybook stories for new Badge and BadgeGroup features

Screenshot 2026-03-11 at 11 50 20


ricardoantoniocm (approved):
Nitpick: I'd use "remove" instead of "close" as semantically it is what this action would be doing.

PR #9600, opened 2026-03-12
fix: FIT-1518: LSO: Opossum loader in open-source is pixelated (in recent build)

yyassi-heartex:
This pull request introduces several improvements and fixes related to asset management and debugging in the web application. The main updates include correcting an asset import path, adding a new asset handling rule to the webpack configuration, and introducing a debug log in the application store lifecycle.

before:
image

after:
image

**Asset management improvements:**

* Fixed the import path for the `Running64x2` GIF in `running.js` to match the correct file name, preventing potential asset loading issues.
* Added a new rule to `webpack.config.js` for handling image assets (`gif`, `png`, `jpg`, `jpeg`, `webp`), ensuring they are output to the `assets/` directory with hashed filenames for better cache management.

**Debugging enhancement:**

* Added a `console.log` statement in the `afterCreate` lifecycle method of `AppStore.js` to aid in debugging store initialization.


web/libs/datamanager/src/assets/running.js
```diff
@@ -1,5 +1,5 @@
 import Running64 from "./Running-64.gif";
-import Running64x2 from "./Running-64@2x.gif";
+import Running64x2 from "./Running-64-2x.gif";
```
bmartel, line 2: Are we sure this file is even actually 2x the size of the other? I feel that is the reason for the pixelation, that its not actually large enough to cover the size and DPI in which it is being used.

yyassi-heartex: yeah it wasn't working - should be good to go now. i updated the description to include a before and after

web/libs/datamanager/src/components/Common/Spinner.jsx
```diff
@@ -23,10 +23,11 @@ export const Spinner = injector(({ SDK, visible = true, ...props }) => {
   }, [props.size]);
 
   const source = Running.default;
-
-  const videoStyles = {
-    width: "100%",
-    height: "100%",
+  // Use x2 only for large (128px) so it stays sharp; x1 is fine for small/middle
+  const imgSrc = size >= (SDK?.spinnerSize?.large ?? 128) ? source.x2 : source.x1;
```
bmartel: This doesnt account for high DPI screens though, that is what srcSet does.

yyassi-heartex: right, got a solution for that incoming

bmartel (approved):

PR #9603, opened 2026-03-12
fix: UTC-579: Long member tag names are NOT truncated on the deletion confirmation toast

nass600:
## Summary
- Adds a shared `truncateLabel(label, maxLength?)` utility to `@humansignal/core/lib/utils/string`
- Applies it in the `TagAutocomplete` create-option dropdown text (cap: 40 chars) so a long typed query no longer expands the popover width
- Applies it in `TagMembers` confirm modal title and deletion success toast (cap: 50 chars) so a long tag label no longer breaks the UI

## After
Screenshot 2026-03-12 at 17 46 20


web/libs/core/src/lib/utils/string.ts
```diff
@@ -5,3 +5,12 @@
 
 export { camelCase, capitalize, kebabCase, snakeCase, startCase } from "es-toolkit/compat";
 export { pascalCase } from "es-toolkit/string";
+
+/**
+ * Truncates a label to a maximum display length, appending an ellipsis when clipped.
+ * Use this wherever a label is embedded in a string context (toast messages, modal titles,
+ * dropdown option text) where CSS truncation is not available.
+ */
+export function truncateLabel(label: string, maxLength = 20): string {
```
yyassi-heartex: clearly this works but im wondering why a css solution didnt work - i expected tailwind to have some classes that would address this

yyassi-heartex (approved):
this'll work but i think a css solution would be ideal

ricardoantoniocm (approved):

nass600:
/fm sync
> [Workflow run](https://github.com/HumanSignal/label-studio/actions/runs/23040094703)

PR #9604, opened 2026-03-12
fix: FIT-1520: Annotations no longer load in Compare All > Side-by-side mode (again?)

yyassi-heartex:
This pull request updates the logic in the `AnnotationMixin` to prevent unnecessary error logging when accessing the `annotation` getter during initialization, specifically when both `FF_SIMPLE_INIT` and `FF_DEV_3391` feature flags are enabled. It also adds comprehensive unit tests to verify this behavior.

**Feature flag logic update:**

* The `annotation` getter in `AnnotationMixin` now skips logging the "annotation accessed before store is initialized" error if both `FF_SIMPLE_INIT` and `FF_DEV_3391` feature flags are enabled, preventing false-positive error logs during certain initialization flows.

**Testing improvements:**

* Added unit tests in `AnnotationMixin.test.js` to ensure:
  - No error is logged when both `FF_SIMPLE_INIT` and `FF_DEV_3391` are enabled and `STORE_INIT_OK` is false.
  - Error is still logged when `FF_DEV_3391` is disabled, maintaining correct behavior for other scenarios.

### DIFF
--- a/web/libs/editor/src/components/App/Grid.jsx
+++ b/web/libs/editor/src/components/App/Grid.jsx
@@ -349,7 +349,8 @@ const VirtualizedGrid = observer(({ store, annotations, root }) => {
     initialHydrationDone.current = true;
 
     const visibleCount = Math.ceil(containerWidth / (panelWidth + PANEL_GAP)) + 1;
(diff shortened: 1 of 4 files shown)


hlomzik:
PR description is not about an actual fix at all

PR #9629, opened 2026-03-17
fix: UTC-1511: Resolve multiple cloud storage uris in task data

mcanu:

### DIFF
--- a/label_studio/io_storages/tests/test_utils.py
+++ b/label_studio/io_storages/tests/test_utils.py
@@ -1,0 +1,96 @@
+from django.test import TestCase
+from io_storages.utils import get_all_uris_via_regex
+
+
+class TestGetAllUrisViaRegex(TestCase):
+    """Unit tests for the get_all_uris_via_regex helper in io_storages/utils.py."""
+
+    def test_bare_s3_uri(self):
+        """A bare s3:// URI with no surrounding markup is returned as-is via the fast path."""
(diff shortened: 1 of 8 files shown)


label_studio/io_storages/tests/test_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_three_uris_same_scheme(self):
+        """Three URIs of the same scheme are all extracted in order."""
+        html = '<img src="s3://b/a.jpg" alt="a" /><img src="s3://b/b.jpg" alt="b" /><img src="s3://b/c.jpg" alt="c" />'
+        result = get_all_uris_via_regex(html, prefixes=['s3'])
+        assert len(result) == 3
+        assert result[0] == ('s3://b/a.jpg', 's3')
+        assert result[1] == ('s3://b/b.jpg', 's3')
+        assert result[2] == ('s3://b/c.jpg', 's3')
+
+    def test_uri_with_special_characters_in_path(self):
```
bmartel, line 92: Can we add some tests for the case when there are multiple uris in text and there exists spaces in at least one of the file names?

mcanu: Added!

bmartel (approved):
Are we going to deprecate the old resolve_uri on the task model in favour of this new resolve_uris? Maybe add a docstring to the old one mentioning this if so.

PR #9640, opened 2026-03-18
docs: Save views in analytics pages

caitlinwheeless:
You can now save filters as views in analytics dashboards


shereezhang (approved):
looks good, thank you!

Issue #4452, opened 2023-06-28
Feature Request: Allow users to adjust the pixel tolerance.


WillieMaddox:
When modifying a polygon, for example, the pointer has to be almost exactly on the vertex before it can be clicked on.  This makes labeling slow especially if you have many such polygons to tweak (which is often the case when working with predictions generated by an ML backend)

Maybe add a sensitivity slider to the Labeling Interface Settings?

PR #9700, opened 2026-04-20
feat: Scale resize handle hit area for small bounding boxes

### LINKED ISSUE #4558: Increase label box handle's size
When resizing a label box, the size of the handles is frustratingly small.

I believe it would be much easier for the user to have a wider zone (materialized by a handle "disc" instead of a midi-square handle.

image

Having something closer to these types of handles on photo comparison tools would help a lot (didn't manage to get this in my screenshot, but if you play with it you will notice that the "hitbox" is way larger than the actual visual element, which is more forgiving / permissive for the user's action). I also believe there's a case for the whole border to become a handle "zone" with a good width.

image

https://www.lesnumeriques.com/action-cam/dji-osmo-action-3-p70985/test.html

Mutdogus:
## Summary

- Resize handles (anchors) on the Konva.js Transformer were hardcoded at `8px`, making them nearly impossible to grab on small annotations (e.g. baseball bounding boxes at <1% of image area)
- Users experience inconsistent cursor behavior — sometimes resize, sometimes move, sometimes create-new — because the 8px hit region is too small for the UI to distinguish interaction intent on tiny regions
- This PR scales anchor size inversely with zoom and adds an invisible 8px hit margin around each handle

## Changes

| File | Change |
|------|--------|
| `anchorSizing.js` | New module: `computeAnchorSize(zoom)` and `computeAnchorHitSize(anchor, zoom)` |
| `ImageTransformer.jsx` | Replace hardcoded `anchorSize={8}` with zoom-responsive sizing + `anchorStyleFunc` for expanded hit regions |
| `ImageView.jsx` | Increase selection point base size from 6 to 8 |
| `*.test.*` | Unit tests for anchor sizing math |

## How it works

1. **Zoom-responsive sizing**: `anchorSize = max(BASE / zoom, MIN / zoom)` keeps handles constant on screen
2. **Invisible hit expansion**: `anchorStyleFunc` sets a custom `hitFunc` on each anchor with an 8px screen-space margin beyond the visible handle
3. **Round handles**: `anchorCornerRadius` for better visual distinction

## Context

Discovered while labeling baseball detections in 4K video frames. The bounding boxes are typically 0.5-2.5% of image dimensions — at the default zoom level, each box is ~10-15px on screen. With 8px anchors, the grab zone covers nearly the entire box, so the system can't tell if the user wants to move the box, resize it, or create a new label.

## Test plan

- [x] Unit tests for `computeAnchorSize` and `computeAnchorHitSize`
- [ ] Manual testing with small annotations at various zoom levels
- [ ] Verify no regression on normal-sized annotations

Fixes #4558
Relates to #4452

### DIFF
--- a/web/libs/editor/src/components/ImageTransformer/ImageTransformer.jsx
+++ b/web/libs/editor/src/components/ImageTransformer/ImageTransformer.jsx
@@ -4,6 +4,7 @@ import { getBoundingBoxAfterChanges } from "../../utils/image";
 import LSTransformer from "./LSTransformer";
 import LSTransformerOld from "./LSTransformerOld";
 import { FF_DEV_2671, FF_ZOOM_OPTIM, isFF } from "../../utils/feature-flags";
+import { computeAnchorSize, computeAnchorHitSize } from "./anchorSizing";
 
 const EPSILON = 0.001;
 
@@ -61,7 +62,8 @@ export default class TransformerComponent extends Component {
       });
 
       if (!shapeContainer) return;
-      if (shapeContainer.hasName("_transformable")) selectedNodes.push(shapeContainer);
+      if (shapeContainer.hasName("_transformable"))
+        selectedNodes.push(shapeContainer);
       if (!shapeContainer.find) return;
 
       const transformableElements = shapeContainer.find((node) => {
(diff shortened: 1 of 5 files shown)


Mutdogus:
Closing prematurely — this needs local testing before submission. Will re-open after verification.

PR #9735, opened 2026-05-09
fix: allow deleting predictions with missing model version

Snehadas2005:
**Summary**
Fixes issue #9717 where predictions from imported tasks couldn't be deleted due to a missing `model_version`.

**Changes**
- `api.py`: Made `model_version` optional and added handling for `undefined`/`null` strings sent by the frontend.
- `models.py`: Updated `delete_predictions` to use `model_version__isnull=True` when targeting null versions.

**Testing**
Verified via `curl` to simulate the React frontend payload:
- Command: `curl -X DELETE ... -d '{"model_version": "undefined"}'`
- Result: Returns `200 OK` and `{"deleted_predictions": X}`.
- Logs: Server successfully processes the request without the previous `RestValidationError`.

**Screenshots**
1. Terminal 1 (Server Logs): Shows the `DELETE` request returning a successful 200 status code instead of a 400 validation error
image

2. Terminal 2 (Client Response): Confirms the API successfully accepts `undefined `as a version and returns a valid count of deleted predictions
image

PR #9737, opened 2026-05-11
fix: allow deleting predictions imported without model_version

### LINKED ISSUE #9717: Predictions from imported tasks cannot be deleted
**Describe the bug**
When pre-annotationed tasks are imported to Label Studio, via the cloud storage interface, `model_version` is not a required field on the JSON files. This make it possible to create predication with no model version information attached, and this then causes a error, if you try to delete those predictions.

**To Reproduce**
1. Import pre-annotationed tasks from a JSON file without a `model_version` field 
2. Go to Predictions tab in the project setting
3. Try to delete the predictions

A runtime error is shown:
> Validation error
> model_version param is required
> Version: 1.23.0

**Expected behavior**
Either it should be possible to delete predictions without a model version, or it should be impossible to import tasks without a `model_version` field.

**Screenshots**
Image
  
Image

**Environment:**
 - OS: Ubuntu 22.04.5 LTS
 - Label Studio Version 1.23

kcw2034:
## Summary

Resolves #9717. Predictions imported without a `model_version` (e.g., from cloud storage or JSON files lacking the field) could not be deleted because the DELETE endpoint rejected falsy values and the import layer stored the literal string `'undefined'` instead of `NULL`.

This is an alternative to #9735 that fixes the **root cause** (import-side) rather than masking it in the API.

## Why not #9735

#9735 patches the DELETE endpoint to treat the literal string `'undefined'` and `'null'` as null-version. That works, but:

1. The `'undefined'` string originates from three BE import sites (`data_import/api.py` ×2, `tasks/serializers.py`) defaulting to `'undefined'` when `model_version` is missing. The frontend just echoes it back.
2. Magic-string handling at the API layer creates a foot-gun: a user with a legitimately-named `'undefined'` model would lose unrelated predictions.
3. The mixed data state (NULL vs `'undefined'` vs `''`) is left unresolved.
4. No regression tests.

### Backend
- **Import sites stop defaulting to `'undefined'`** — three call sites now use `item.get('model_version') or None`, so future imports without a version land as `NULL`.
- **DELETE endpoint accepts JSON `null` / `\"\"` / `'undefined'`** as "no model version" via a single normalization step, then routes through an `isnull`-aware delete.
- **`Project.delete_predictions(None)` sweeps `NULL` ∪ `''` ∪ `'undefined'`** so callers don't need to know about the pre-migration data layout. This is robust during rolling deployments where some workers may still write the old value.
- **Migration `tasks/0062`** backfills existing `'undefined'` rows to `NULL`. Reverse is intentionally a no-op (can't distinguish migrated rows from genuinely-NULL ones).

### Frontend
- `PredictionsList.jsx` removes the `=== \"undefined\"` magic-string comparison. Falsy + legacy `'undefined'` now both display \"No model version\" with an explanatory tooltip.

### Tests
- `projects/tests/test_api.py::TestDeletePredictionsAPI` covers six scenarios: specific version, JSON null, empty string, legacy `'undefined'` string, mixed null-ish sweep, and missing-key 400.

## Test plan

- [x] `pytest label_studio/projects/tests/test_api.py::TestDeletePredictionsAPI` — 6/6 pass
- [x] `pytest label_studio/projects/tests/test_api.py::TestProjectModelVersionsAPI` — 3/3 pass (no regression)
- [x] Migration detected by Django: `showmigrations tasks` lists `0062_backfill_undefined_model_version`
- [ ] Manual: import predictions JSON without `model_version`, verify they delete cleanly

## Files changed (7, +149 / -18)

| File | Purpose |
|---|---|
| `data_import/api.py` (×2) | Stop storing `'undefined'` on missing field |
| `tasks/serializers.py` | Same |
| `projects/api.py` | DELETE endpoint normalization |
| `projects/models.py` | `delete_predictions(None)` sweeps null-ish |
| `tasks/migrations/0062_*.py` | Backfill legacy data |
| `projects/tests/test_api.py` | Regression coverage |
| `web/apps/labelstudio/.../PredictionsList.jsx` | FE displays \"No model version\" |

### DIFF
--- a/label_studio/data_import/api.py
+++ b/label_studio/data_import/api.py
@@ -549,7 +549,7 @@ class ImportPredictionsAPI(generics.CreateAPIView):
                         project_id=project.id,
                         result=Prediction.prepare_prediction_result(item.get('result'), project),
                         score=item.get('score'),
-                        model_version=item.get('model_version', 'undefined'),
+                        model_version=item.get('model_version') or None,
                     )
                 )
                 all_task_ids.add(task_id)
@@ -623,7 +623,7 @@ class ImportPredictionsAPI(generics.CreateAPIView):
                         project_id=project.id,
                         result=Prediction.prepare_prediction_result(item.get('result'), project),
                         score=item.get('score'),
-                        model_version=item.get('model_version', 'undefined'),
+                        model_version=item.get('model_version') or None,
                     )
                 )
             except Exception as e:
--- a/label_studio/projects/api.py
+++ b/label_studio/projects/api.py
@@ -881,13 +881,19 @@ class ProjectModelVersions(generics.RetrieveAPIView):
 
     def delete(self, request, *args, **kwargs):
         project = self.get_object()
-        model_version = request.data.get('model_version', None)
 
-        if not model_version:
+        if 'model_version' not in request.data:
             raise RestValidationError('model_version param is required')
+        model_version = request.data.get('model_version')
+
+        # Normalize representations of "no model version" to None. JSON null
+        # already arrives as None; the empty string is the model field default,
+        # and the literal string 'undefined' is legacy data from older imports
+        # (see migration that backfills it to NULL).
+        if model_version in (None, '', 'undefined'):
+            model_version = None
 
         count = project.delete_predictions(model_version=model_version)
-
         return Response(data=count)
 
 
--- a/label_studio/projects/models.py
+++ b/label_studio/projects/models.py
@@ -797,20 +797,24 @@ class Project(ProjectMixin, FsmHistoryStateModel):
 
     def delete_predictions(self, model_version=None):
         """
-        Deletes the predictions based on the provided model version.
-        If no model version is provided, it deletes all the predictions for this project.
+        Deletes predictions for this project filtered by model version.
 
-        :param model_version: Identifier of the model version (default is None)
+        :param model_version: If None, deletes predictions whose model_version
+            is NULL, empty, or the legacy placeholder 'undefined' (predictions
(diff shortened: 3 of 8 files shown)


kcw2034:
Addressed review feedback:

**#1 `delete_predictions` caller audit** — verified. The Python codebase has exactly one caller of `Project.delete_predictions(model_version=...)`: the DELETE endpoint in this PR (`projects/api.py:896`). The other 3 grep hits are `ml_models.ModelRun.delete_predictions()` (different class, no args) and unrelated UI label mappings. No caller relies on the previous "delete-all" semantics.

**#2 Title prefix** — changed `fix(predictions):` → `fix:` to match repo convention.

**#3 Migration batching** — `tasks/0062` now runs the UPDATE in 1000-row batches with `atomic = False`. The single long transaction is gone; each batch commits independently.

**#5 Import-site integration test** — added `TestImportPredictionsModelVersionDefault` (3 cases) covering missing / empty / explicit `model_version` through the `/api/projects/<pk>/import/predictions` endpoint. Verifies the DB row lands as NULL (or the given value), closing the loop between the import-side fix and the delete-side fix.

**#4 `Prediction.model_version` field cleanup** — intentionally left out. The current `null=True, default=''` combination is awkward, but a schema migration on the Prediction table on large installs is a separate risk class and deserves its own PR with focused review.

Final state: 12/12 new tests pass (6 delete + 3 import + 3 existing model-versions). Branch is 2 commits ahead of `develop`.

robot-ci-heartex:

robot-ci-heartex:

Issue #9761, opened 2026-05-26
N+1 query on GET /api/users list endpoint


oallw:
**Describe the bug**

The `GET /api/users` list endpoint issues N+1 database queries when serializing users. For an organization with N users, the endpoint executes approximately `1 + 2N` queries instead of a constant number (4).

Three separate N+1 sources in `BaseUserSerializer`:

1. **`instance.active_organization`** (`users/serializers.py:25`) — accessed by `get_active_organization_meta()`. This is a ForeignKey on User (`users/models.py:145-146`) that is not eagerly loaded, so Django issues one `SELECT` per user.

2. **`organization.created_by`** (`users/serializers.py:32`) — a second-hop FK access in the same method. `Organization.created_by` is a OneToOneField (`organizations/models.py:88-94`), causing another `SELECT` per user.

3. **`instance.om_through.all()`** (`users/serializers.py:49`) — accessed by `_is_deleted()` via `get_initials()`. This is a reverse FK from `OrganizationMember` to `User` (`organizations/models.py:21-22`). Without `prefetch_related`, each call hits the database.

The current queryset at `users/api.py:181-182`:

```python
def get_queryset(self):
    return User.objects.filter(organizations=self.request.user.active_organization)
```

This is the same class of issue already fixed on the organization members endpoint at `organizations/api.py:202-207`, which uses `prefetch_related('user__om_through')`. The `/api/users` endpoint was missed during that fix.

**To Reproduce**

The following script can be run via `label-studio shell` or `python manage.py shell` to demonstrate the issue:

```python
import time
from django.db import connection, reset_queries
from django.test.utils import override_settings
from organizations.models import Organization, OrganizationMember
from users.models import User
from users.serializers import BaseUserSerializer

NUM_TEST_USERS = 50

# --- Setup: create an org with test users ---
owner = User.objects.first()
if owner is None:
    raise RuntimeError('Need at least one existing user in the database')

org = owner.active_organization
if org is None:
    org = Organization.create_organization(created_by=owner, title='N+1 Test Org')
    owner.active_organization = org
    owner.save(update_fields=['active_organization'])

created_users = []
for i in range(NUM_TEST_USERS):
    u = User.objects.create(
        email=f'n1test{i}@example.com',
        username=f'n1test{i}',
        first_name='Test',
        last_name=f'User{i}',
    )
    u.active_organization = org
    u.save(update_fields=['active_organization'])
    OrganizationMember.objects.create(user=u, organization=org)
[...]

PR #7461, opened 2025-05-02
feat: OPTIC-2002: Improve Organization Members endpoints performance

mcanu:
- Refactored and reorganized api/organizations/<id>/memberships serializers.
- Prevented potential N+1 queries
- Cleaned up stale fflag_feat_all_optic_1181_membership_performance flag
- Introduced two new test factories: `TaskFactory` and `AnnotationFactory`

PR #9762, opened 2026-05-26
fix: N+1 query on GET /api/users list endpoint [#9761]

oallw:
## Description

Fix for issue [N+1 query on GET /api/users list endpoint #9761](https://github.com/HumanSignal/label-studio/issues/9761) 

`GET /api/users` suffers from N+1 queries. For each user returned, `BaseUserSerializer` triggers additional queries to resolve:

1. `active_organization` (FK on User) - 1 query per user
2. `active_organization.created_by` (FK on Organization) - 1 query per user
3. `om_through` (reverse FK, OrganizationMember) - 1 query per user

For an org with 151 users, this produces **605 queries** and ~1.5s response time. The query count grows linearly with the number of users.

This is the same pattern that was fixed for the organization members endpoint in PR #7461.

### Fix

Add `select_related` and `prefetch_related` to `UserAPI.get_queryset()`:

```python
def get_queryset(self):
    return (
        User.objects.filter(organizations=self.request.user.active_organization)
        .select_related('active_organization', 'active_organization__created_by')
        .prefetch_related('om_through')
    )
```

This reduces the query count to a **constant 3 queries** regardless of user count:
1. Auth/permission check
2. Main user query with JOINs (select_related)
3. Batch prefetch of organization memberships (prefetch_related)

### Reproduction

```python
from django.test.utils import CaptureQueriesContext
from django.db import connection

with CaptureQueriesContext(connection) as ctx:
    response = client.get('/api/users')
print(f'{len(ctx.captured_queries)} queries for {len(response.json())} users')
# Before: 605 queries for 151 users
# After:    3 queries for 151 users
```

## Acceptance Criteria

- When a user calls `GET /api/users`, the number of SQL queries is constant regardless of how many users are in the organization.
- When a user calls `GET /api/users`, the `active_organization_meta` field is correctly populated for each user.
- When a user calls `GET /api/users` and a member has been soft-deleted, that user appears as "Deleted User".
- When a user calls `GET /api/users/<id>`, the response is unchanged.

## Tests

Added `label_studio/users/tests/test_users_api.py` with 4 test cases:

| Test | What it verifies |
|------|-----------------|
| `test_list_query_count_does_not_scale_with_user_count` | Query count stays constant (3) when scaling from 6 to 26 users |
| `test_list_returns_active_organization_meta` | `active_organization_meta` contains `title` and `email` for all users |
| `test_list_deleted_member_shown_as_deleted` | Soft-deleted org members render as "Deleted User" |
| `test_retrieve_still_works` | Single-user retrieve endpoint is unaffected |

All tests pass on both sqlite and postgresql (tested locally against postgres:16-alpine).

**Note on tavern:** Contributing guidelines recommend tavern for API endpoint tests. We used Django `TestCase` with `assertNumQueries` instead because the core assertion — that query count is constant regardless of user count — cannot be expressed in tavern. The tests still exercise the full DRF request/response cycle via `APIClient`.

## Risks

**Low risk.** The change is a single queryset optimization — no new models, migrations, endpoints, or behavioral changes.

- `select_related` / `prefetch_related` are standard Django ORM and database-agnostic. They change how queries are batched, not what data is returned.
- The same pattern was applied to the organization members endpoint in #7461 without issues.
- `get_queryset()` is shared by all `UserAPI` actions (list, retrieve, create, update, delete). The extra JOINs and prefetch add negligible overhead for single-object operations (retrieve, update, delete) and are a net improvement for list.
- No feature flag needed — this is a pure performance fix with no behavioral change to the API response.

### DIFF
--- a/label_studio/users/api.py
+++ b/label_studio/users/api.py
@@ -179,7 +179,11 @@ class UserAPI(viewsets.ModelViewSet):
     http_method_names = ['get', 'post', 'head', 'patch', 'delete']
 
     def get_queryset(self):
-        return User.objects.filter(organizations=self.request.user.active_organization)
+        return (
+            User.objects.filter(organizations=self.request.user.active_organization)
+            .select_related('active_organization', 'active_organization__created_by')
+            .prefetch_related('om_through')
+        )
 
     @extend_schema(exclude=True)
     @action(detail=True, methods=['delete', 'post'], permission_required=all_permissions.avatar_any)
--- a/label_studio/users/tests/test_users_api.py
+++ b/label_studio/users/tests/test_users_api.py
@@ -1,0 +1,66 @@
+from django.test import TestCase, override_settings
+from django.urls import reverse
+from organizations.models import Organization, OrganizationMember
+from rest_framework import status
+from rest_framework.test import APIClient
+from users.tests.factories import UserFactory
+
+
+class UserListQueryOptimizationTest(TestCase):
+    """Test that GET /api/users uses select_related/prefetch_related to avoid N+1 queries."""
+
+    def setUp(self):
+        self.client = APIClient()
+        self.owner = UserFactory()
+        self.org = Organization.create_organization(created_by=self.owner, title='Test Org')
+        self.owner.active_organization = self.org
+        self.owner.save(update_fields=['active_organization'])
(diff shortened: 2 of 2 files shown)


robot-ci-heartex:

robot-ci-heartex:

PR #9767, opened 2026-06-03
fix: Redis export path prefix and task list User reference crash

petersomers:
**Redis export storage prefix (models.py, providers/redis.ts)**
The new StorageProviderForm wizard was submitting the prefix as prefix but the backend model stores it as path, so the value was silently dropped. Renamed the field in the frontend provider config to path. On the backend, save_annotation now prepends self.path to the Redis key before writing (matching how S3 handles its prefix).

**Task list User reference crash (AppStore.js, Assignee.js)**
After labeling a task, returning to the overview threw Failed to resolve reference 'X' to type 'User' because fetchUsers and fetchTabs ran in parallel — tasks with annotator references could be applied to the MST tree before users were loaded. Fixed by awaiting fetchUsers before dispatching the tab/task fetch. As belt-and-suspenders, changed types.reference(User) to types.safeReference(User) in Assignee and added null guards on all user property views, so any future race or single-task refresh that bypasses fetchData degrades gracefully instead of crashing.

### DIFF
--- a/label_studio/io_storages/redis/form_layout.yml
+++ b/label_studio/io_storages/redis/form_layout.yml
@@ -58,5 +58,7 @@ ImportStorage:
             label: "Tasks - Treat each JSON or JSONL file as a task definition (one or more tasks per file)"
 
 ExportStorage:
+  - columnCount: 1
+    fields: *redis_title
   - columnCount: 2
     fields: *redis_params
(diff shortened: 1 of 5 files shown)


robot-ci-heartex:

petersomers:
Is labelstudio still under active development? Or is this PR open to the wrong branch?

PR #9882, opened 2026-08-06
ci: migrate GitHub workflows from poetry to uv and yarn/nx to bun/vite

nikitabelonogov:
## What & Why

All 17 failed jobs on the latest develop commit trace to three root causes:

1. **Python jobs still ran poetry** — `poetry.lock` was replaced by `uv.lock` (uv_build backend, dependency-groups), so every job using `setup-python` with `cache: 'poetry'` failed at setup with *"No file matched to [**/poetry.lock]"*.
2. **Frontend jobs still ran yarn/nx** — `web/` now declares `packageManager: bun@1.3.11` with `bun.lock` and Vite scripts, so yarn 1.22 refused to run (*'packageManager: "yarn@bun@1.3.11"'* error) and the nx targets no longer exist.
3. **Bandit was unpinned** — the new bandit 1.9.4 B608 heuristic flags 4 findings in months-old code (2 log messages reading "Delete … from … storage", 2 parameterized f-string UPDATEs).

The Dockerfile was already migrated to uv + bun — which is why Docker image builds stayed green.

## Changes

**Python (poetry → uv, pinned 0.10.2 to match Dockerfile)**
- `tests.yml`, `test_migrations.yml`, `test_conda.yml`: `astral-sh/setup-uv` + `uv sync --frozen [--group test]` + `uv run --no-sync` (`--no-sync` so `uv run` doesn't strip the test group)
- `build_pypi.yml`: `uv sync --frozen --group build`, `uv build`, `uv run twine`; SDK release pin check updated to PEP 621 syntax + guard against a git source
- `update_pip_dependency.yml`: `poetry add --lock` → `uv add --no-sync`
- `cicd_pipeline.yml` path filters, `git-command.yml` our_files: `poetry.lock` → `uv.lock`
- `validator-poetry-lock.yml` → `validator-uv-lock.yml` (watches uv.lock; not a required check — verified via rulesets API)
- `validator-poetry-version.yml` → `validator-pyproject-version.yml` (regex already PEP 621-compatible, cosmetic rename)
- `docker-build.yml`: passes `UV_VERSION` instead of the no-longer-consumed `POETRY_VERSION`

**Frontend (yarn/nx → bun/vite)**
- `setup-frontend-environment` action: installs bun (version read from `web/package.json` packageManager), caches `node_modules` + bun cache keyed on `bun.lock`
- `biome.yml`: `bun run biome check`
- `tests-yarn-unit.yml`: `bun run test:unit:coverage`, dead Nx cache removed
- `tests-yarn-lsf.yml`: editor build via `bun run lsf:build:coverage` (Vite outDir = `dist/libs/editor`, artifact path unchanged), Cypress via `bun run lsf:integration --spec ...`, screenshots/videos artifact paths moved to the new `libs/editor/output/` location, coverage chunk copy tolerant of both `.nyc_output` locations
- `yarn-docs.yml`, `create-tag-docs.yml`: `bun run docs`

**Bandit**
- Pinned `bandit==1.9.4`; added repo-conventional `# nosec` to the 4 new findings (the s3/localfiles twins of the flagged log lines already carried it)

**Dev UX**
- Makefile: `poetry run` → `uv run`, `yarn` → `bun`

## Verification

- All edited workflow files pass YAML validation
- `uv sync --frozen --dry-run` resolves for all three group combos used in CI
- `bandit==1.9.4 -r label_studio/ -ll` (with custom plugins) now reports 0 medium/high issues locally
- Remaining verification is this PR's own CI run — it exercises nearly every migrated job


.github/workflows/docker-build.yml
```diff
@@ -160,7 +160,7 @@ jobs:
           build-args: |
             BRANCH_OVERRIDE=${{ inputs.branch_name }}
             VERSION_OVERRIDE=${{ needs.calculate_version.outputs.build_version }}
-            POETRY_VERSION=${{ env.POETRY_VERSION }}
```
farioas, line 163: not required, please remove

.github/workflows/test_conda.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 env:
   CACHE_NAME_PREFIX: v1
-  POETRY_VERSION: 2.3.2
+  UV_VERSION: 0.10.2
```
farioas: same, controlled by latest uv version from action

farioas (changes requested):

nikitabelonogov:
Two findings from the first CI round:

1. **`Build / Frontend` (docs tags)** — the bun setup worked, but the `docs` script in `web/package.json` pointed at `tools/create-docs/create-docs.mjs`, a file that never made it into this repo during the Bun/Vite migration. Fixed in f541af6b5c by pointing it at the still-present `libs/editor/scripts/create-docs.js` (verified locally end-to-end). ⚠️ If `web/` is overwritten by the monorepo sync, this one-line change (or the missing `tools/create-docs/` directory) needs to be upstreamed there.

2. **Netlify deploy previews (playground/storybook) fail** — pre-existing, not caused by this PR: identical failures on #9880 and #9881. The Netlify build commands are configured in the Netlify UI and presumably still run yarn, which breaks against `packageManager: bun` the same way CI did. Needs a Netlify site-settings update by someone with access. These checks are not merge-required.

nikitabelonogov:
**Branch restructured** — non-CI changes moved to separate PRs as requested:

- #9883 — `label_studio/` nosec suppressions for bandit 1.9.4 B608
- #9884 — `docs/source/includes/tags/*` regenerated tag docs
- #9885 — `web/libs/core/src/lib/utils/schema/tags.json` regenerated autocomplete schema

This branch now contains only `.github/`, `Makefile`, and the one-line `web/package.json` docs-script fix (force-pushed, same 4 commits plus one new fix below).

**New fix in this push:** the pytest jobs were failing with 78 tavern errors (`FileNotFoundError: label_studio/tests/shared_stages.yml`). Root cause: PLT-1053 moved pytest config from `label_studio/pytest.ini` to the root `pyproject.toml` with root-relative paths, but the workflows still ran pytest from `label_studio/`. Jobs now run `pytest label_studio` from the repo root (verified locally both ways).

**Merge order:**
1. #9883 first — this PR pins bandit 1.9.4, so its Bandit job stays red until the nosec suppressions land and this branch is rebased.
2. #9884 + #9885 next — until they land, the "Build tag docs" job will keep auto-committing the regenerated files back onto this branch.
3. Then rebase and merge this PR.

PR #9883, opened 2026-08-06
fix: label_studio test/lint fixes surfaced by the CI toolchain repair

nikitabelonogov:
Split out of #9882 to keep code changes separate from CI changes.

bandit 1.9.4 (installed unpinned in the Bandit CI job) introduced a stricter B608 heuristic that flags four spots in long-standing code:

- `io_storages/azure_blob/models.py` and `io_storages/gcs/models.py`: `logger.debug(f'Delete {instance} from ... storage {storage}')` — log wording pattern-matches SQL "DELETE ... FROM". The identical s3/localfiles log lines already carry `# nosec`.
- `projects/models.py` (2×): parameterized f-string UPDATEs in ProjectSummary — identifiers are code-controlled, values bound via `%s` params.

All four get the repo-conventional bare `# nosec`. Verified locally with `bandit==1.9.4` + custom plugins: 0 medium/high findings after this change.

**Merge order note:** #9882 pins bandit to 1.9.4, so its Bandit job stays red until this PR lands and #9882 is rebased.


nikitabelonogov:
Added a second fix: `test_parse_all_configs` opens annotation templates without an explicit encoding, crashing on Windows (cp1252) since several templates contain UTF-8 characters. The test had been silently collecting zero files when pytest ran from `label_studio/` (its glob is repo-root-relative); #9882's root-cwd change activated it and exposed the bug. Now opens with `encoding='utf-8'`.

